PR #529, opened 2019-12-10
[NOD-420] Process delayed blocks

dereeno:


blockdag/dag.go
```diff
@@ -1799,6 +1812,46 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
 	return dag.subnetworkID
 }
 
+// AddDelayedBlock adds a block to the delayed blocks list.
+func (dag *BlockDAG) AddDelayedBlock(block *util.Block, delay time.Duration) {
+	// Trigger processing of current delayed blocks when adding a new delayed block.
+	dag.ProcessDelayedBlocks()
```
someone235: Process after adding the current block, not before

blockdag/dag.go
```diff
@@ -1799,6 +1812,46 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
...
+
+	dag.delayedBlocksLock.Lock()
+	defer dag.delayedBlocksLock.Unlock()
+	dag.delayedBlocks[*block.Hash()] = &delayedBlock{
```
someone235: Add to heap by timestamp

blockdag/process.go
```diff
@@ -168,6 +168,22 @@ func (dag *BlockDAG) ProcessBlock(block *util.Block, flags BehaviorFlags) (isOrp
...
 	}
 
+	// Handle the case of a block with a valid timestamp(non-delayed) which points to a delayed block.
+	dag.delayedBlocksLock.RLock()
+	delay = 0
+	for _, parentHash := range block.MsgBlock().Header.ParentHashes {
+		if delayedParent, exists := dag.delayedBlocks[*parentHash]; exists {
+			parentDelay := delayedParent.processTime.Sub(dag.timeSource.AdjustedTime())
+			if parentDelay > delay {
+				delay = parentDelay
```
someone235: Fix to `parentDleay + 1nanosecond` (pseudo code)

netsync/manager.go
```diff
@@ -474,9 +474,8 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
 	}
 
 	if delay != 0 {
-		spawn(func() {
-			sm.QueueBlock(bmsg.block, bmsg.peer, true, make(chan struct{}))
-		}, sm.handlePanic)
+		sm.dag.AddDelayedBlock(bmsg.block, delay)
```
someone235: Consider moving this logic inside ProcessBlock

blockdag/dag.go
```diff
@@ -1799,6 +1812,46 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
...
+		block:       block,
+		processTime: dag.timeSource.AdjustedTime().Add(delay),
+	}
+}
+
+// RemoveDelayedBlock removes a block from the delayed blocks list.
+func (dag *BlockDAG) RemoveDelayedBlock(block *util.Block) {
```
someone235: Please unexport this

blockdag/dag.go
```diff
@@ -1799,6 +1812,46 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
...
+// 1. When adding a new delayed block
+// 2. After processing a block (ProcessBlock method).
+func (dag *BlockDAG) ProcessDelayedBlocks() error {
+	dag.delayedBlocksLock.RLock()
+	defer dag.delayedBlocksLock.RUnlock()
+	for _, delayedBlock := range dag.delayedBlocks {
+		if dag.timeSource.AdjustedTime().After(delayedBlock.processTime) {
+			_, _, err := dag.ProcessBlock(delayedBlock.block, BFAfterDelay)
+			if err != nil {
+				log.Debugf("Error while processing delayed block (block %s)", delayedBlock.block.Hash().String())
```
someone235: Fix to Errorf

blockdag/process.go
dereeno: Added below (line 185)

blockdag/process.go
```diff
@@ -168,6 +168,24 @@ func (dag *BlockDAG) ProcessBlock(block *util.Block, flags BehaviorFlags) (isOrp
 		}
 	}
 
+	// Handle the case of a block with a valid timestamp(non-delayed) which points to a delayed block.
+	dag.delayedBlocksLock.RLock()
+	delay = 0
+	for _, parentHash := range block.MsgBlock().Header.ParentHashes {
```
svarogg: This check should only happen after base check that all parents exist, and loop only over parents that don't exist

blockdag/process.go
```diff
@@ -164,19 +164,26 @@ func (dag *BlockDAG) ProcessBlock(block *util.Block, flags BehaviorFlags) (isOrp
...
 	for _, parentHash := range block.MsgBlock().Header.ParentHashes {
 		if !dag.BlockExists(parentHash) {
-			allParentsExist = false
+			missingParents = append(missingParents, parentHash)
 		}
 	}
 
-	if !allParentsExist {
+	// Handle the case of a block with a valid timestamp(non-delayed) which points to a delayed block.
+	delay = dag.maxDelayOfParents(missingParents)
```
svarogg: What if block has both delayed parents, and genuinely missing parents orphan-style?

At the very least - the missing parents should be requested.

Also, I don't think we are re-validating all parents are present once delay is over.

blockdag/dag.go
```diff
@@ -1839,7 +1839,9 @@ func (dag *BlockDAG) ProcessDelayedBlocks() error {
 		earliestDelayedBlockProcessTime := dag.delayedBlocksQueue.peek().processTime
 		if dag.timeSource.AdjustedTime().After(earliestDelayedBlockProcessTime) {
 			delayedBlock := dag.popDelayedBlock()
-			err := dag.maybeAcceptBlock(delayedBlock.block, BFAfterDelay)
+			dag.dagLock.Unlock()
```
svarogg: I'd rather if you created function processBlockNoLock that has all the logic from process block, and than call it from here and the public ProcessBlock.

Unlocking here means someone might grab the process and do stuff that you don't want at this stage.

blockdag/dag.go
```diff
@@ -1799,6 +1813,79 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
 	return dag.subnetworkID
 }
 
+// AddDelayedBlock adds a block to the delayed blocks list.
+func (dag *BlockDAG) AddDelayedBlock(block *util.Block, delay time.Duration) {
```
someone235: Please unexport

blockdag/dag.go
```diff
@@ -1799,6 +1813,79 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
...
+	dag.delayedBlocksLock.Lock()
+	defer dag.delayedBlocksLock.Unlock()
+
+	processTime := dag.timeSource.AdjustedTime().Add(delay)
+	log.Debugf("Adding block to delayed blocks queue (block hash: %s, process time: %s)", block.Hash().String(), processTime)
+	db := &delayedBlock{
```
someone235: Can you change `db` to something else in all places? It reminds database

blockdag/dag.go
```diff
@@ -1799,6 +1813,79 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
...
+		processTime: processTime,
+	}
+
+	dag.delayedBlocks[*block.Hash()] = db
+	dag.delayedBlocksQueue.Push(db)
+}
+
+// ProcessDelayedBlocks loops over all delayed blocks and processes blocks which are due.
+// This method is invoked after processing a block (ProcessBlock method).
+func (dag *BlockDAG) ProcessDelayedBlocks() error {
```
someone235: Please unexport

blockdag/dag.go
```diff
@@ -1799,6 +1813,79 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
...
+			delayedBlock := dag.popDelayedBlock()
+			dag.dagLock.Unlock()
+			_, _, err := dag.ProcessBlock(delayedBlock.block, BFAfterDelay)
+			dag.dagLock.Lock()
+			if err != nil {
+				log.Errorf("Error while processing delayed block (block %s)", delayedBlock.block.Hash().String())
+				return err
+			}
+			log.Debugf("Processed delayed block (block %s)", delayedBlock.block.Hash().String())
+		} else {
```
someone235: Please use the opposite condition to break, and then you can drop the `else` condition

blockdag/dag.go
```diff
@@ -1799,6 +1813,79 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
...
+// this function must be called with delayedBlocksLock held (for writes)
+func (dag *BlockDAG) popDelayedBlock() *delayedBlock {
+	db := dag.delayedBlocksQueue.pop()
+	delete(dag.delayedBlocks, *db.block.Hash())
+	return db
+}
+
+// maxDelayOfParents returns the maximum delay of the given block hashes.
+// Returns 0 if none of the parents are delayed.
+func (dag *BlockDAG) maxDelayOfParents(parentHashes []*daghash.Hash) time.Duration {
```
someone235: return additional boolean instead of 0

blockdag/dag.go
```diff
@@ -1799,6 +1813,79 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
...
+			parentDelay := delayedParent.processTime.Sub(dag.timeSource.AdjustedTime())
+			if parentDelay > delay {
+				delay = parentDelay
+			}
+		}
+	}
+
+	if delay != 0 {
+		// Add Nanosecond to ensure that parent process time will be after its child.
+		delay += time.Nanosecond
```
someone235: I think this should be added in the call site, or at least change the function name to represent it

blockdag/delayedblockheap.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (h *baseDelayedBlocksHeap) Pop() interface{} {
+	oldHeap := *h
+	oldLength := len(oldHeap)
+	popped := oldHeap[oldLength-1]
+	*h = oldHeap[0 : oldLength-1]
+	return popped
+}
+
+func (h baseDelayedBlocksHeap) Peek() interface{} {
```
someone235: Please unexport

blockdag/dag.go
someone235: Didn't we say that we wanna process delayed blocks before adding a new one?

blockdag/process.go
```diff
@@ -164,19 +164,26 @@ func (dag *BlockDAG) ProcessBlock(block *util.Block, flags BehaviorFlags) (isOrp
 		}
 
 		if delay != 0 {
+			dag.AddDelayedBlock(block, delay)
 			return false, delay, err
```
someone235: Do we still want to return the delay? maybe just a boolean isDelayed?

blockdag/dag.go
someone235: I'm not sure when exactly do you use the return value of maxDelayOfParents except comparing it to 0. If it's the case, return only boolean

blockdag/process.go
someone235: It seems like you forgot to add the block with delayed parent to the delayed pool

blockdag/process.go
```diff
@@ -147,36 +147,51 @@ func (dag *BlockDAG) ProcessBlock(block *util.Block, flags BehaviorFlags) (isOrp
...
 	// The block must not already exist as an orphan.
 	if _, exists := dag.orphans[*blockHash]; exists {
 		str := fmt.Sprintf("already have block (orphan) %s", blockHash)
-		return false, 0, ruleError(ErrDuplicateBlock, str)
+		return false, false, ruleError(ErrDuplicateBlock, str)
 	}
 
-	if !isDelayedBlock {
+	if _, exists := dag.delayedBlocks[*blockHash]; exists {
+		dag.processDelayedBlocks()
```
someone235: This should never happen, so I think it's better to return an error

blockdag/dag.go
```diff
@@ -1817,6 +1829,63 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
...
+	dag.delayedBlocksQueue.Push(delayedBlock)
+	dag.processDelayedBlocks()
+}
+
+// processDelayedBlocks loops over all delayed blocks and processes blocks which are due.
+// This method is invoked after processing a block (ProcessBlock method).
+func (dag *BlockDAG) processDelayedBlocks() error {
+	// Check if the delayed block with the earliest process time should be processed
+	for dag.delayedBlocksQueue.Len() > 0 {
+		earliestDelayedBlockProcessTime := dag.delayedBlocksQueue.peek().processTime
```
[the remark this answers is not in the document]
dereeno: The break condition for the loop is `dag.delayedBlocksQueue.Len() > 0`

blockdag/dag.go
```diff
@@ -1817,6 +1829,51 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
...
+	processTime := dag.timeSource.AdjustedTime().Add(delay)
+	log.Debugf("Adding block to delayed blocks queue (block hash: %s, process time: %s)", block.Hash().String(), processTime)
+	delayedBlock := &delayedBlock{
+		block:       block,
+		processTime: processTime,
+	}
+
+	dag.delayedBlocks[*block.Hash()] = delayedBlock
+	dag.delayedBlocksQueue.Push(delayedBlock)
+	dag.processDelayedBlocks()
```
stasatdaglabs: Please handle processDelayedBlocks' error.

blockdag/validate.go
```diff
@@ -432,6 +432,8 @@ func (dag *BlockDAG) checkBlockHeaderSanity(header *wire.BlockHeader, flags Beha
 
 	// Ensure the block time is not too far in the future. If it's too far, return
 	// the duration of time that should be waited before the block becomes valid.
+	// This check needs to be last as it does not return an error but rather mark the
```
stasatdaglabs: mark -> marks

blockdag/process.go
```diff
@@ -189,8 +189,8 @@ func (dag *BlockDAG) processBlockNoLock(block *util.Block, flags BehaviorFlags)
 	if isParentDelayed {
 		// Add Nanosecond to ensure that parent process time will be after its child.
 		delay += time.Nanosecond
-		dag.addDelayedBlock(block, delay)
-		return false, true, nil
+		err := dag.addDelayedBlock(block, delay)
+		return false, true, err
```
someone235: Please return `false, false, err` if `err != nil`

PR #559, opened 2019-12-30
[NOD-517] Update doc.go files

stasatdaglabs:


doc.go
```diff
@@ -1,121 +1,20 @@
-// Copyright (c) 2013-2017 The btcsuite developers
```
svarogg: The main doc.go actually does need the copyright notice.

doc.go
```diff
@@ -1,121 +1,26 @@
...
-                            default settings for the active network.
-      --rejectnonstd        Reject non-standard transactions regardless of the
-                            default settings for the active network.
-
-Help Options:
-  -h, --help           Show this help message
+For an up-to-date help message:
+  kaspad --help
 
+An interesting point to note is that the long form of all option flags
```
someone235: Delete "An interesting point to note is that "

logs/LICENSE
```diff
@@ -1,15 +0,0 @@
-ISC License
-
-Copyright (c) 2013-2014 Conformal Systems LLC.
```
someone235: Please add it to the main copyright section

rpcclient/doc.go
```diff
@@ -123,7 +119,7 @@ The third category of errors, that is errors returned by the server, can be
 detected by type asserting the error in a *rpcmodel.RPCError. For example, to
 detect if a command is unimplemented by the remote RPC server:
 
-  amount, err := client.GetBalance("")
+  amount, err := client.GetNetTotals()
```
someone235: GetNetTotals doesn't return amount

PR #560, opened 2019-12-30
[NOD-541] Implement GHOSTDAG

someone235:


blockdag/ghostdag.go
```diff
@@ -0,0 +1,132 @@
...
+		var candidateAnticoneSize uint32
+		possiblyBlue := true
+
+		for chainBlock := newNode; possiblyBlue; chainBlock = chainBlock.selectedParent {
+			// If blueCandidate is in the future of chainBlock, it means
+			// that all remaining blues are in past(chainBlock) and thus
+			// in past(blueCandidate). In this case we know for sure that
+			// anticone(blueCandidate) will not exceed K, and we can mark
+			// it as blue.
+			//
```
svarogg: These are two separate comments. To signal this, it can be nice to remove the // on this line (and leave it as an empty line)

blockdag/ghostdag.go
```diff
@@ -0,0 +1,132 @@
...
+func (dag *BlockDAG) blueAnticoneSize(block, context *blockNode) (uint32, error) {
+	for node := context; node != nil; node = node.selectedParent {
+		if blueAnticoneSize, ok := node.bluesAnticoneSizes[*block.hash]; ok {
+			return blueAnticoneSize, nil
+		}
+	}
+	return 0, errors.Errorf("block %s is not in blue-set of %s", block.hash, context.hash)
+}
+
+func (dag *BlockDAG) ghostdag(newNode *blockNode) (*blockHeap, error) {
```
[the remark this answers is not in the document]
svarogg: Also: is there any significance to anyone outside ghostdag for selectedParentAnticone being a blockHeap (as opposed to []*blockNode)?

If not - please convert to []*blockNode before returning.

blockdag/ghostdag.go
```diff
@@ -0,0 +1,132 @@
...
+	newNode.selectedParent = newNode.parents.bluest()
+	newNode.bluesAnticoneSizes[*newNode.hash] = 0
+	newNode.blues = append(newNode.blues, newNode.selectedParent)
+	selectedParentAnticone, err := dag.selectedParentAnticone(newNode)
+	if err != nil {
+		return nil, err
+	}
+
+	for selectedParentAnticone.Len() > 0 {
+		blueCandidate := selectedParentAnticone.pop()
```
svarogg: If you are poping stuff out of selectedParentAnticone here, wouldn't it be empty by the time you want to return it?

blockdag/ghostdag.go
```diff
@@ -0,0 +1,132 @@
...
+		candidateBluesAnticoneSizes := make(map[*blockNode]uint32)
+		var candidateAnticoneSize uint32
+		possiblyBlue := true
+
+		for chainBlock := newNode; possiblyBlue; chainBlock = chainBlock.selectedParent {
+			// If blueCandidate is in the future of chainBlock, it means
```
svarogg: I was not able to understand what this comment says. Main reason for the confusion is because I don't really understand yet what you are trying to do.
Possibly because I'm stupid, but nevertheless, consider to revise, or to add context in comments beforehand.

It's possible that once you add comment that explains how ghostdag works, this will already be clear.

blockdag/ghostdag.go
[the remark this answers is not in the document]
svarogg: Also: Add comment that explains in general terms how ghostdag works.

blockdag/ghostdag.go
```diff
@@ -0,0 +1,132 @@
...
+				}
+
+				candidateBluesAnticoneSizes[block], err = dag.blueAnticoneSize(block, newNode)
+				if err != nil {
+					return nil, err
+				}
+				candidateAnticoneSize++
+				if candidateAnticoneSize > dag.dagParams.K || candidateBluesAnticoneSizes[block] == dag.dagParams.K {
+					// Two possible k-cluster violations here:
+					// 	(i) The candidate blue anticone now became larger than k
```
svarogg: I'd split the if into two lines, 1 per condition, and near every condition place the comment that says what kind of violation it is.

blockdag/blocknode.go
```diff
@@ -75,6 +76,11 @@ type blockNode struct {
 	// blueScore is the count of all the blue blocks in this block's past
 	blueScore uint64
 
+	// bluesAnticoneSizes is a map holding the set of blues affected by this block and their modified blue anticone size.
+	//
+	// TODO: Use *blockNode as the map key.
```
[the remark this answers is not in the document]
someone235: Because it affects deserialization

blockdag/ghostdag.go
```diff
@@ -0,0 +1,132 @@
...
+			queue = append(queue, parent)
+		}
+	}
+	return &anticoneHeap, nil
+}
+
+// blueAnticoneSize returns the blue anticone size of 'block' from the worldview of 'context'.
+// Expects 'block' to be ∈ blue-set(context)
+func (dag *BlockDAG) blueAnticoneSize(block, context *blockNode) (uint32, error) {
+	for node := context; node != nil; node = node.selectedParent {
```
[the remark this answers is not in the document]
someone235: The stop condition is only for sanity. In reality it should take K steps

blockdag/ghostdag.go
```diff
@@ -0,0 +1,132 @@
...
+			anticoneHeap.Push(parent)
+			queue = append(queue, parent)
+		}
+	}
+	return &anticoneHeap, nil
+}
+
+// blueAnticoneSize returns the blue anticone size of 'block' from the worldview of 'context'.
+// Expects 'block' to be ∈ blue-set(context)
+func (dag *BlockDAG) blueAnticoneSize(block, context *blockNode) (uint32, error) {
```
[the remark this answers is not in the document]
someone235: But it's dependent on the new block POV

blockdag/ghostdag.go
someone235: I think this paragraph is related to the above paragraph, so it's better to leave it like this

blockdag/blocknode.go
```diff
@@ -102,14 +108,21 @@ type blockNode struct {
...
+	return node.selectedParent.chainHeight + 1
+}
+
+// newBlockNode returns a new block node for the given block header and parent
+//nodes. This function is NOT safe for concurrent access.
+func (dag *BlockDAG) newBlockNode(blockHeader *wire.BlockHeader, parents blockSet) (*blockNode, *blockHeap) {
+	node := &blockNode{
+		parents:            parents,
+		children:           make(blockSet),
+		timestamp:          time.Now().Unix(),
```
dereeno: do we want to use time.Now() or dag.timeSource.AdjustedTime() ?

blockdag/ghostdag.go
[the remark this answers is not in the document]
someone235: > Also, here's a thought, let me know if you agree or not:
> How about ghostdag doesn't modify anything in newNode, instead it returns of ghostdag's outcomes and newBlockNode will use these values to modify newNode.
> 
> IMHO it's a bit of a cleaner design.

When we go over `chainBlock.blues` `chainBlock` can be `newNode`, and if `ghostdag` will not update `newNode.blues` on the fly it's gonna complicate things.

blockdag/accept.go
```diff
@@ -47,7 +47,7 @@ func (dag *BlockDAG) maybeAcceptBlock(block *util.Block, flags BehaviorFlags) er
 	}
 
 	// Create a new block node for the block and add it to the node index.
-	newNode := newBlockNode(&block.MsgBlock().Header, parents, dag.dagParams.K)
+	newNode, selectedParentAnticone := dag.newBlockNode(&block.MsgBlock().Header, parents)
```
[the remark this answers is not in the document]
someone235: I don't think it's a good idea to save the anticone in memory just for using it two times. Do you have any other idea?

blockdag/virtualblock_test.go
```diff
@@ -5,14 +5,25 @@
...
 )
 
 // TestVirtualBlock ensures that VirtualBlock works as expected.
 func TestVirtualBlock(t *testing.T) {
-	phantomK := uint32(1)
-	buildNode := buildNodeGenerator(phantomK, false)
+	// Create a new database and DAG instance to run tests against.
+	params := dagconfig.SimNetParams
+	params.K = 1
+	dag, teardownFunc, err := DAGSetup("TestChainUpdates", Config{
```
stasatdaglabs: This is not TestChainUpdates.

blockdag/virtualblock_test.go
```diff
@@ -5,14 +5,25 @@
...
+		DAGParams: &params,
+	})
+	if err != nil {
+		t.Fatalf("TestChainUpdates: Failed to setup DAG instance: %s", err)
```
stasatdaglabs: This is not TestChainUpdates.

blockdag/virtualblock_test.go
```diff
@@ -100,11 +111,21 @@ func TestVirtualBlock(t *testing.T) {
 }
 
 func TestSelectedPath(t *testing.T) {
-	phantomK := uint32(1)
-	buildNode := buildNodeGenerator(phantomK, false)
+	// Create a new database and DAG instance to run tests against.
+	params := dagconfig.SimNetParams
+	params.K = 1
+	dag, teardownFunc, err := DAGSetup("TestChainUpdates", Config{
```
stasatdaglabs: This is not TestChainUpdates.

blockdag/virtualblock_test.go
```diff
@@ -100,11 +111,21 @@ func TestVirtualBlock(t *testing.T) {
...
+		DAGParams: &params,
+	})
+	if err != nil {
+		t.Fatalf("TestChainUpdates: Failed to setup DAG instance: %s", err)
```
stasatdaglabs: This is not TestChainUpdates.

blockdag/ghostdag.go
```diff
@@ -0,0 +1,155 @@
...
+					break
+				}
+			}
+
+			for _, block := range chainBlock.blues {
+				// Skip blocks that exists in the past of blueCandidate.
+				// We already checked it for chainBlock above, so if the
+				// block is chainBlock, there's no need to recheck.
+				if block != chainBlock {
+					if isAncestorOf, err := dag.isAncestorOf(block, blueCandidate); err != nil {
```
dereeno: same here: I would change the name of the variable `isAncestorOf ` to something more meaningful like `isAncestorOfBlueCandidate`

blockdag/ghostdag.go
```diff
@@ -0,0 +1,155 @@
...
+			queue = append(queue, parent)
+		}
+	}
+	return &anticoneHeap, nil
+}
+
+// blueAnticoneSize returns the blue anticone size of 'block' from the worldview of 'context'.
+// Expects 'block' to be ∈ blue-set(context)
+func (dag *BlockDAG) blueAnticoneSize(block, context *blockNode) (uint32, error) {
+	for current := context; current != nil; current = current.selectedParent {
```
dereeno: why `current != nil` is the condition to break the loop, shouldn't it be `current == genesis`?

blockdag/ghostdag.go
```diff
@@ -0,0 +1,155 @@
...
+
+// ghostdag updates newNode.blues, newNode.selectedParent
+// and newNode.bluesAnticoneSizes according to the GHOSTDAG
+// protocol.
+// It updates newNode.blues by going over the anticone of
+// newNode.selectedParent (which is the parent with the
+// highest blue score) and adds it to newNode.blues if it
+// passes two conditions:
+// 1) |anticone(block) ∩ blueset(newNode)| <= K
+// 2) For every blue in blueset(newNode) |anticone(blue) ∩ blueset(newNode) ∩ {block}| <= K.
```
dereeno: I suppose the last `∩` should be `∪`

blockdag/ghostdag.go
someone235: It doesn't really matter. It's done only for sanity to prevent nil pointer exception

blockdag/ghostdag.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return 0, errors.Errorf("block %s is not in blue-set of %s", block.hash, context.hash)
+}
+
+// ghostdag updates newNode.blues, newNode.selectedParent
+// and newNode.bluesAnticoneSizes according to the GHOSTDAG
+// protocol.
+// It updates newNode.blues by going over the anticone of
+// newNode.selectedParent (which is the parent with the
+// highest blue score) and adds it to newNode.blues if it
```
svarogg: Not 100% clear what 'it' means here.
Also "going over" is an over-generalized choice of word, I'd rather use "iterating over

blockdag/ghostdag.go
```diff
@@ -0,0 +1,156 @@
...
+// passes two conditions:
+// 1) |anticone(block) ∩ blueset(newNode)| <= K
```
svarogg: `blueset` is not defined anywhere in our code.
Either give definition ahead of this line, or revise.

Also: `blueset(newNode)` is not the sort of notation we usually use (We always prefer something like `newNode.blueSet`).

blockdag/ghostdag.go
```diff
@@ -0,0 +1,156 @@
...
+	return 0, errors.Errorf("block %s is not in blue-set of %s", block.hash, context.hash)
+}
+
+// ghostdag updates newNode.blues, newNode.selectedParent
+// and newNode.bluesAnticoneSizes according to the GHOSTDAG
+// protocol.
+// It updates newNode.blues by going over the anticone of
+// newNode.selectedParent (which is the parent with the
+// highest blue score) and adds it to newNode.blues if it
+// passes two conditions:
```
svarogg: I'd be more concise about "if it passes two conditions":
It's not that the block should pass both conditions, it's that the DAG should not violate these two conditions if the block in question is added to newNode.blues.

blockdag/ghostdag.go
```diff
@@ -0,0 +1,156 @@
...
+// 1) |anticone(block) ∩ blueset(newNode)| <= K
+// 2) For every blue in blueset(newNode) |anticone(blue) ∩ blueset(newNode) ∪ {block}| <= K.
+//    We do this by maintaining for each block a map bluesAnticoneSizes which holds
```
svarogg: I can go-on nitpicking this text (sorry for being an asshole, but it's really critical :disappointed: ), but I think the best approach would be to catch Eyal and ask him to help us with good wording.

blockdag/ghostdag.go
```diff
@@ -0,0 +1,156 @@
...
+	selectedParentAnticone = make([]*blockNode, selectedParentAnticoneHeap.Len())
+	for i := 0; selectedParentAnticoneHeap.Len() > 0; i++ {
+		blueCandidate := selectedParentAnticoneHeap.pop()
+		selectedParentAnticone[i] = blueCandidate
+		candidateBluesAnticoneSizes := make(map[*blockNode]uint32)
+		var candidateAnticoneSize uint32
+		possiblyBlue := true
+
+		for chainBlock := newNode; possiblyBlue; chainBlock = chainBlock.selectedParent {
+			// If blueCandidate is in the future of chainBlock, it means
```
svarogg: I still don't have enough context to understand this comment. By the time that I reach this place in the code I'm completely lost.

I think what I'm missing in the top comment is an overlay of the procedure happening here (all there is right now is the objective). 

I'm not sure the top comment would be the right place for the procedure though, might be better to just add some more comments inside the code, explaining what the code is doing, and why is it doing it.

When writing the comment try to imagine you are someone who has never heard Michael's lecture about ghostdag.

blockdag/mining.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/kaspanet/kaspad/util/subnetworkid"
+	"github.com/kaspanet/kaspad/wire"
+	"sort"
+	"time"
+)
+
+// BlockForMining returns a block with the given transactions
+// that points to the current DAG tips, that is valid from
+// all aspects except proof of work.
+func (dag *BlockDAG) BlockForMining(transactions []*util.Tx, useMinimalTime bool) (*wire.MsgBlock, error) {
```
svarogg: Is useMinimalTime only there for tests?

If so, I'd rather it not be parameter of dag.BlockForMining. Instead, get the block with timestamp according to dag.timeSource, and than modify it for what the tests needs.

blockdag/mining.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, err
+	}
+	return coinbaseTx, nil
+}
+
+// MinimumMedianTime returns the minimum allowed timestamp for a block building
+// on the end of the DAG. In particular, it is one second after
+// the median timestamp of the last several blocks per the DAG consensus
+// rules.
+func MinimumMedianTime(dagMedianTime time.Time) time.Time {
```
svarogg: 1. MedianTime is implementation details. What this function does תאכלס is calculate the minimum timestamp for block, so it should be named accordingly.
1. I'd rather that this function takes dag and calls calculateMedianTime by itself.

blockdag/mining.go
```diff
@@ -0,0 +1,133 @@
...
+	return dagMedianTime.Add(time.Second)
+}
+
+// MedianAdjustedTime returns the current time adjusted to ensure it is at least
+// one second after the median timestamp of the last several blocks per the
+// DAG consensus rules.
+func MedianAdjustedTime(dagMedianTime time.Time, timeSource MedianTimeSource) time.Time {
```
svarogg: Same here.
This function should take a dag (or better - be a `dag.` method) and be called something like "NextBlockTimestamp".

Right now it leaks a lot of implementation details to the caller.

blockdag/blocknode.go
```diff
@@ -154,6 +155,14 @@ func (node *blockNode) updateParentsChildren() {
 	}
 }
 
+func (node *blockNode) lessThan(other *blockNode) bool {
```
svarogg: Maybe follow `sort` package convention and simply call it `less`?

blockdag/ghostdag.go
```diff
@@ -0,0 +1,167 @@
...
+				} else if isAncestorOfBlueCandidate {
+					break
+				}
+			}
+
+			for _, block := range chainBlock.blues {
+				// Skip blocks that exists in the past of blueCandidate.
+				// We already checked it for chainBlock above, so if the
+				// block is chainBlock, there's no need to recheck.
+				if block != chainBlock {
```
svarogg: You are ranging over chainBlock.blues... I don't think chainBlock can potentially be in it's own blues, so what's the point checking for it?

blockdag/ghostdag.go
```diff
@@ -0,0 +1,167 @@
...
+					break
+				}
+
+				if candidateBluesAnticoneSizes[block] == dag.dagParams.K {
+					// k-cluster violation: A block in candidate's blue anticone already
+					// has k blue blocks in its own anticone
+					possiblyBlue = false
+					break
+				}
+				if candidateBluesAnticoneSizes[block] > dag.dagParams.K {
```
svarogg: It looks like this is just a sanity check. Correct me if I'm wrong.
If so, do we really want it as part of the ghostdag code?

I'm not against sanity checks in general, but in code that is already complicated without it - this could be a bit of an overkill.

blockdag/ghostdag.go
```diff
@@ -0,0 +1,166 @@
...
+
+				if candidateBluesAnticoneSizes[block] == dag.dagParams.K {
+					// k-cluster violation: A block in candidate's blue anticone already
+					// has k blue blocks in its own anticone
+					possiblyBlue = false
+					break
+				}
+
+				// This is a sanity check that validates that a blue
+				// block blue anticone is not already larger than K.
```
svarogg: I think you wanted to say "a blue block**'s** blue anticone..."

blockdag/ghostdag.go
```diff
@@ -0,0 +1,166 @@
...
+				}
+
+				candidateBluesAnticoneSizes[block], err = dag.blueAnticoneSize(block, newNode)
+				if err != nil {
+					return nil, err
+				}
+				candidateAnticoneSize++
+
+				if candidateAnticoneSize > dag.dagParams.K {
+					// k-cluster violation: The candidate blue anticone now became larger than k
```
dereeno: typo: The candidate**'s** blue anticone... 
And, I would change it to "now exceeded k" (but maybe it sounds better for me, so feel free to ignore it)

dereeno: .

PR #563, opened 2019-12-31
[NOD-540] Fix object references in reachabilityData deserialization

stasatdaglabs:


blockdag/reachabilitystore.go
```diff
@@ -125,36 +125,55 @@ func (store *reachabilityStore) init(dbTx database.Tx) error {
...
+	store.loaded[hash] = &reachabilityData{
+		treeNode:          &reachabilityTreeNode{},
+		futureCoveringSet: make(futureCoveringBlockSet, 0),
+	}
+	return nil
+}
+
+func (store *reachabilityStore) loadReachabilityDataFromCursor(cursor database.Cursor) error {
+	var hash daghash.Hash
+	err := hash.SetBytes(cursor.Key())
```
svarogg: Why not use daghash.NewHash instead of SetBytes?

Both here and initReachabilityData

blockdag/reachabilitystore.go
```diff
@@ -125,36 +125,55 @@ func (store *reachabilityStore) init(dbTx database.Tx) error {
...
+	if err != nil {
+		return err
+	}
+
+	data, ok := store.reachabilityDataByHash(&hash)
```
svarogg: The name `data` is a bit problematic in this context, and even more so in `deserializeReachabilityData`'s context, since it does not hold any data at certain point.

I can't think of a better name RN, but try find something yourself, and if not, lets brainstorm together.

blockdag/reachabilitystore.go
```diff
@@ -364,34 +371,33 @@ func (store *reachabilityStore) deserializeReachabilityInterval(r io.Reader) (*r
...
+func (store *reachabilityStore) deserializeFutureCoveringSet(r io.Reader, destination *reachabilityData) error {
 	// Deserialize the set size
 	setSize, err := wire.ReadVarInt(r)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	// Deserialize each block in the set
-	futureCoveringSet := make(futureCoveringBlockSet, setSize)
```
someone235: Why you are not pre-allocating anymore?

PR #565, opened 2020-01-02
[NOD-613] Fix concurrent access to ecmh cache

stasatdaglabs:


mining/mining.go
```diff
@@ -346,6 +346,9 @@ func (g *BlkTmplGenerator) UpdateBlockTime(msgBlock *wire.MsgBlock) error {
 // height. It also recalculates and updates the new merkle root that results
 // from changing the coinbase script.
 func (g *BlkTmplGenerator) UpdateExtraNonce(msgBlock *wire.MsgBlock, extraNonce uint64) error {
+	g.dag.Lock()
```
svarogg: Move lock to be more localized

PR #568, opened 2020-01-05
[NOD-549] Update version to 0.1.0 and allow injection of appBuild

svarogg:


server/p2p/p2p.go
```diff
@@ -65,7 +65,7 @@ var (
 
 	// userAgentVersion is the user agent version and is used to help
 	// identify ourselves to other kaspa peers.
-	userAgentVersion = fmt.Sprintf("%d.%d.%d", version.AppMajor, version.AppMinor, version.AppPatch)
+	userAgentVersion = version.Version()
```
someone235: Please use this var in DefaultUserAgent

version/version.go
```diff
@@ -13,45 +9,31 @@ import (
 // semanticAlphabet
 const semanticAlphabet = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-"
```
someone235: Maybe rename because it doesn't refer to semver anymore?

version/version.go
```diff
@@ -13,45 +9,31 @@ import (
...
+// process with '-ldflags "-X github.com/kaspanet/kaspad/version.appBuild=foo"' if needed.
+// It MUST only contain characters from semanticAlphabet.
 var appBuild string
 
-// Version returns the application version as a properly formed string per the
-// semantic versioning 2.0.0 spec (http://semver.org/).
-func Version() string {
-	// Start with the major, minor, and patch versions.
-	version := fmt.Sprintf("%d.%d.%d", AppMajor, AppMinor, AppPatch)
+var version = ""
```
someone235: Can you add a comment that explains that it's a cache

server/p2p/p2p.go
svarogg: No can do :/
DefaultUserAgent is in wire package.

p2p package extends the user agent with UserAgentComment, so needs to re-do some work.

I guess btcd had DefaultUserAgent in wire because wire was considered a "library package" that should be used outside of btcd.

In our case it's less relevant, and it's only used for tests. I'm not sure it's a problem, but if we decide it is, I think it's better to get rid of it in separate ticket.

PR #572, opened 2020-01-06
[NOD-546] Report build failures to Discord instead of Telegram

stasatdaglabs:


netsync/interface.go
```diff
@@ -22,7 +22,7 @@ type PeerNotifier interface {
 
 	TransactionConfirmed(tx *util.Tx)
 }
-
+gdfgdfg
```
svarogg: Looks wrong

deploy.sh
```diff
@@ -17,12 +17,12 @@ IMAGE_NAME=${ECR_SERVER}/${SERVICE_NAME}
...
-# Sends a Telegram notification with some details about the failure
+# Sends a Discord notification with some details about the failure
 # All variables in this function are set by Jenkins
-notify_telegram() {
-  echo "./telegram.sh \
-    '${TELEGRAM_API_TOKEN}' \
-    '${TELEGRAM_CHAT_ID}' \
+notify_discord() {
+  echo "./discord.sh \
+    '${DISCORD_CLIENT_ID}' \
```
svarogg: Please do this only if DISCORD_CLIENT_ID and DISCORD_API_TOKEN are set.

PR #576, opened 2020-01-07
[NOD-505] Try reading the response in case of upnp error

svarogg:


server/serverutils/upnp.go
```diff
@@ -313,8 +315,13 @@ func soapRequest(url, function, message string) (replyXML []byte, err error) {
 	}
 
 	if r.StatusCode >= 400 {
-		// log.Stderr(function, r.StatusCode)
-		err = errors.New("Error " + strconv.Itoa(r.StatusCode) + " for " + function)
+		data, readErr := ioutil.ReadAll(r.Body)
+		if readErr != nil {
+			err = errors.Wrapf(readErr, "Error %d for %s", r.StatusCode, function)
+		} else {
+			err = errors.Errorf("Error %d for %s. Response: %s", r.StatusCode, function, data)
```
stasatdaglabs: Wouldn't we prefer to wrap this error as well?

svarogg: There's nothing to wrap, just a response code and data

PR #590, opened 2020-01-19
[NOD-665] Initialize blockNode blueScore to be MaxUint64 by default

stasatdaglabs:


blockdag/blocknode.go
```diff
@@ -130,12 +132,16 @@ func (dag *BlockDAG) newBlockNode(blockHeader *wire.BlockHeader, parents blockSe
...
-		if err != nil {
-			panic(errors.Wrap(err, "unexpected error in GHOSTDAG"))
-		}
+	if len(parents) == 0 {
+		// The genesis block is defined to have a blueScore of 0
+		node.blueScore = 0
+		return node, nil
+	}
+
+	var err error
```
svarogg: You don't need to declare error now, use := instead

PR #592, opened 2020-01-20
[NOD-669] Fix startSync sending a blockLocatorMsg with a zeroHash insead of the peer's selectedTip

stasatdaglabs:


netsync/manager.go
```diff
@@ -189,7 +189,7 @@ func (sm *SyncManager) startSync() {
 		log.Infof("Syncing to block %s from peer %s",
 			bestPeer.SelectedTip(), bestPeer.Addr())
 
-		bestPeer.PushGetBlockLocatorMsg(&daghash.ZeroHash, sm.dagParams.GenesisHash)
+		bestPeer.PushGetBlockLocatorMsg(bestPeer.SelectedTip(), sm.dagParams.GenesisHash)
```
svarogg: While you are there refactoring stuff, how about rename bestPeer to syncPeer, as it's called anywhere else?

PR #593, opened 2020-01-20
[NOD-640] Revamp blueBlocksBetween to return up to maxEntries from lowNode's antiPast to highNode's antiFuture

stasatdaglabs:


blockdag/dag.go
```diff
@@ -1630,30 +1630,83 @@ func (dag *BlockDAG) blueBlocksBetween(lowHash, highHash *daghash.Hash, maxEntri
 	// first iterate on the selected parent chain of the highNode
 	// until we find a new highNode
 	// where highNode.blueScore-lowNode.blueScore+1 <= maxEntries
-
+	//
+	// We also keep track of highNode's selected parent chain child
```
svarogg: In the beginning of this comment we state "In order to get no more then maxEntries of blue blocks from... ".
I believe we are no longer only collecting blue blocks.

It might be also a good idea to state the assumption that most blocks are blue, and therefore `highNode.blueScore-lowNode.blueScore+1` is a good estimation to total amount of blocks we are going to get.

Also it could be nice to have an explanation why moving the highNode one notch lower is a good idea.

blockdag/dag.go
```diff
@@ -1630,30 +1630,83 @@ func (dag *BlockDAG) blueBlocksBetween(lowHash, highHash *daghash.Hash, maxEntri
...
-	nodes = append(nodes, highNode)
-	current := highNode
-	for current.blueScore > lowNode.blueScore {
-		for _, blue := range current.blues {
-			nodes = append(nodes, blue)
+	// Collect every node in highNode's past (including itself) but
+	// NOT in the lowNode's past (excluding itself) into an UpHeap
+	// (a heap sorted by blueScore from lowest to greatest).
+	visited := newSet()
+	highNodeAntiFuture := newUpHeap()
```
svarogg: This is actually highNodeAntiFuture∩lowNodeAntiPast.
I know it will make a very long variable name, but it's important to make this clear.

I'd rather have something generic such as `candidateBlocks` or something with comment above explaining exactly what it is, than `highNodeAntiFuture` which is plain wrong.

blockdag/dag.go
```diff
@@ -1630,30 +1630,83 @@ func (dag *BlockDAG) blueBlocksBetween(lowHash, highHash *daghash.Hash, maxEntri
...
+	if highNodeAntiFuture.Len() < int(maxEntries) {
+		// If we initially (from the function args) received
+		// highNode.blueScore-lowNode.blueScore+1 < maxEntries
+		// then we don't immediately know the selected parent
+		// chain child of highNode. To find it, we iterate
+		// over highNode's children until we stumble upon a
+		// child whose selectedParent is highNode.
+		if highNodeSelectedParentChainChild == nil {
+			for _, child := range highNode.children {
+				if child.selectedParent == highNode {
```
svarogg: There might be a few of those, and they are not necessarily the selected-parent-chain child.
You can use the selected-parent-chain-set to check for this.

However, I think (Please verify I am correct here) you don't actually need the glorified selected-parent-chain-child, but indeed, any node whose selectedParent is highNode will do. In that case - update comments and names to reflect this

blockdag/dag.go
```diff
@@ -1630,30 +1630,83 @@ func (dag *BlockDAG) blueBlocksBetween(lowHash, highHash *daghash.Hash, maxEntri
...
 
-	// Populate and return the found nodes.
-	nodes := make([]*blockNode, 0, highNode.blueScore-lowNode.blueScore+1)
-	nodes = append(nodes, highNode)
-	current := highNode
-	for current.blueScore > lowNode.blueScore {
-		for _, blue := range current.blues {
-			nodes = append(nodes, blue)
+	// Collect every node in highNode's past (including itself) but
+	// NOT in the lowNode's past (excluding itself) into an UpHeap
```
someone235: why UpHeap and not up heap or upHeap?

blockdag/dag.go
```diff
@@ -1620,12 +1624,14 @@ func (dag *BlockDAG) blueBlocksBetween(lowHash, highHash *daghash.Hash, maxEntri
 		return nil, errors.Errorf("Couldn't find high hash %s", highHash)
 	}
 
-	// In order to get no more then maxEntries of blue blocks from
-	// the future of the lowNode (including itself), we iterate
-	// the selected parent chain of the highNode and add the blues
-	// each node (including the highNode itself). This is why the
-	// number of returned blocks will be
+	// In order to get no more then maxEntries blocks from the
+	// future of the lowNode (including itself), we iterate the
```
someone235: we iterate the parents, not selected parent

blockdag/dag.go
```diff
@@ -1635,10 +1641,10 @@ func (dag *BlockDAG) blueBlocksBetween(lowHash, highHash *daghash.Hash, maxEntri
 	}
 
 	// Collect every node in highNode's past (including itself) but
-	// NOT in the lowNode's past (excluding itself) into an UpHeap
+	// NOT in the lowNode's past (excluding itself) into an upHeap
```
someone235: it's still weird to write upHeap in camel case if it doesn't appear in the code as a single word

blockdag/dag.go
```diff
@@ -1651,49 +1657,50 @@ func (dag *BlockDAG) blueBlocksBetween(lowHash, highHash *daghash.Hash, maxEntri
...
 	nodesLen := int(maxEntries)
-	if highNodeAntiFuture.Len() < nodesLen {
-		nodesLen = highNodeAntiFuture.Len()
+	if candidateNodes.Len() < nodesLen {
+		nodesLen = candidateNodes.Len()
 	}
 	nodes := make([]*blockNode, 0, nodesLen)
-	for len(nodes) <= nodesLen {
-		nodes = append(nodes, highNodeAntiFuture.pop())
+	for len(nodes) < nodesLen {
```
someone235: I think it'll be clearer to convert it to `for candidateNodes.Len() > 0 && len(nodes) < maxEntries` and get rid of `nodesLen`

PR #594, opened 2020-01-21
[NOD-629] change GHOSTDAG k to uint8

dereeno:


dagconfig/params.go
```diff
@@ -82,11 +82,16 @@ const (
 	DefinedDeployments
 )
 
+// KSize defines the size of GHOSTDAG consensus algorithm K parameter.
+type KSize uint8
```
svarogg: Maybe KType rather than KSize?

blockdag/blocknode_test.go
```diff
@@ -95,3 +97,36 @@ func TestChainHeight(t *testing.T) {
...
+		DAGParams: &dagconfig.SimnetParams,
+	})
+	if err != nil {
+		t.Fatalf("TestBlueAnticoneSizesSize: Failed to setup DAG instance: %s", err)
+	}
+	defer teardownFunc()
+	blockHeader := dagconfig.SimnetParams.GenesisBlock.Header
+	block, _ := dag.newBlockNode(&blockHeader, newSet())
+
+	maxKsize := dagconfig.KType(math.MaxUint8)
```
someone235: Please explain why do you use max size

blockdag/blocknode_test.go
```diff
@@ -95,3 +97,36 @@ func TestChainHeight(t *testing.T) {
...
+	expected := maxKsize
+	block.bluesAnticoneSizes[daghash.Hash{1}] = expected
+	serializedNode, _ := serializeBlockNode(block)
+	deserializedNode, _ := dag.deserializeBlockNode(serializedNode)
+	if deserializedNode.bluesAnticoneSizes[daghash.Hash{1}] != expected {
+		t.Fatalf("TestBlueAnticoneSizesSize: BlueAnticoneSize should not change when deserilizing. Expected: %v but got %v",
+			expected, deserializedNode.bluesAnticoneSizes[daghash.Hash{1}])
+	}
+
+	block.bluesAnticoneSizes[daghash.Hash{1}]++
```
someone235: Please add comment to explain why do you overflow

blockdag/blocknode_test.go
```diff
@@ -95,3 +97,36 @@ func TestChainHeight(t *testing.T) {
...
+func TestBlueAnticoneSizesSize(t *testing.T) {
+	dag, teardownFunc, err := DAGSetup("TestBlueAnticoneSizesSize", Config{
+		DAGParams: &dagconfig.SimnetParams,
+	})
+	if err != nil {
+		t.Fatalf("TestBlueAnticoneSizesSize: Failed to setup DAG instance: %s", err)
+	}
+	defer teardownFunc()
+	blockHeader := dagconfig.SimnetParams.GenesisBlock.Header
+	block, _ := dag.newBlockNode(&blockHeader, newSet())
```
someone235: we usually call it `node` and not `block`

blockdag/blocknode_test.go
```diff
@@ -95,3 +97,36 @@ func TestChainHeight(t *testing.T) {
...
+
+	maxKsize := dagconfig.KType(math.MaxUint8)
+	expected := maxKsize
+	block.bluesAnticoneSizes[daghash.Hash{1}] = expected
```
someone235: Can you put `daghash.Hash{1}` inside a variable and reuse?

blockdag/blocknode_test.go
```diff
@@ -95,3 +97,36 @@ func TestChainHeight(t *testing.T) {
...
+	serializedNode, _ := serializeBlockNode(block)
+	deserializedNode, _ := dag.deserializeBlockNode(serializedNode)
+	if deserializedNode.bluesAnticoneSizes[daghash.Hash{1}] != expected {
+		t.Fatalf("TestBlueAnticoneSizesSize: BlueAnticoneSize should not change when deserilizing. Expected: %v but got %v",
+			expected, deserializedNode.bluesAnticoneSizes[daghash.Hash{1}])
+	}
+
+	block.bluesAnticoneSizes[daghash.Hash{1}]++
+	size := block.bluesAnticoneSizes[daghash.Hash{1}]
```
someone235: why you called it `expected` in one place and `size` in the other?

blockdag/blocknode_test.go
```diff
@@ -109,24 +117,27 @@ func TestBlueAnticoneSizesSize(t *testing.T) {
...
+	if deserializedNode.bluesAnticoneSizes[hash] != maxKType {
 		t.Fatalf("TestBlueAnticoneSizesSize: BlueAnticoneSize should not change when deserilizing. Expected: %v but got %v",
-			expected, deserializedNode.bluesAnticoneSizes[daghash.Hash{1}])
+			maxKType, deserializedNode.bluesAnticoneSizes[hash])
 	}
 
-	block.bluesAnticoneSizes[daghash.Hash{1}]++
-	size := block.bluesAnticoneSizes[daghash.Hash{1}]
-	serializedNode, _ = serializeBlockNode(block)
+	// Increasing bluesAnticoneSizes by 1 to make it overflow
```
someone235: I think this part of the test will always pass, because 0 is always serializable. I think that the first part might suffice

blockdag/blocknode_test.go
```diff
@@ -95,3 +96,48 @@ func TestChainHeight(t *testing.T) {
...
+	})
+	if err != nil {
+		t.Fatalf("TestBlueAnticoneSizesSize: Failed to setup DAG instance: %s", err)
+	}
+	defer teardownFunc()
+	blockHeader := dagconfig.SimnetParams.GenesisBlock.Header
+	node, _ := dag.newBlockNode(&blockHeader, newSet())
+	hash := daghash.Hash{1}
+
+	// Setting maxKType to maximum value fo KType
```
svarogg: `^dagconfig.KType(0)` is maximum value of KType because we know for sure KType is unsigned (otherwise it would be -1 for signed types).
Please mention in the comment this assumption, and that it's true as long as TestKTypeIsUnsigned doesn't fail.

svarogg: Also: typo: fo -> of

blockdag/blocknode_test.go
```diff
@@ -95,3 +96,48 @@ func TestChainHeight(t *testing.T) {
...
+	maxKType := ^dagconfig.KType(0)
+	node.bluesAnticoneSizes[hash] = maxKType
+	serializedNode, _ := serializeBlockNode(node)
+	deserializedNode, _ := dag.deserializeBlockNode(serializedNode)
+	if deserializedNode.bluesAnticoneSizes[hash] != maxKType {
+		t.Fatalf("TestBlueAnticoneSizesSize: BlueAnticoneSize should not change when deserilizing. Expected: %v but got %v",
```
svarogg: typo: deserilizing -> deseri**a**lizing

blockdag/blocknode_test.go
svarogg: I agree that this test right now will always pass, but I'd argue the first part will not suffice.
The first part will catch the case where bluesAnticoneSizes is smaller than KType, but if it's wider it will work just fine.

I think in second part we should set `expected := 0` (Or `maxKType + 1`, which equals 0), and than it will check what we want to check.

PR #600, opened 2020-01-26
[NOD-696] Handle panics on time.AfterFunc

someone235:


util/panics/panics.go
```diff
@@ -44,3 +45,26 @@ func GoroutineWrapperFuncWithPanicHandler(log logs.Logger) func(func(), func())
 		}()
 	}
 }
+
+// AfterFuncWrapperFunc returns a time.AfterFunc wrapper function that handles panics,
```
stasatdaglabs: , -> .

PR #604, opened 2020-01-27
[NOD-700] Convert blockSet to map[*blockNode]struct{}

stasatdaglabs:


blockdag/dag.go
```diff
@@ -1475,8 +1483,13 @@ func (dag *BlockDAG) oldestChainBlockWithBlueScoreGreaterThan(blueScore uint64)
 // parent chain.
 //
 // This method MUST be called with the DAG lock held
-func (dag *BlockDAG) IsInSelectedParentChain(blockHash *daghash.Hash) bool {
-	return dag.virtual.selectedParentChainSet.containsHash(blockHash)
+func (dag *BlockDAG) IsInSelectedParentChain(blockHash *daghash.Hash) (bool, error) {
```
svarogg: Add comment that says that this function works only for blocks that have already been added to the blockIndex.

PR #605, opened 2020-01-27
[NOD-702] Fix netsync slowing down significantly due to excessive allocs in serializeUTXO

stasatdaglabs:


blockdag/utxodiffstore.go
```diff
@@ -19,14 +20,20 @@ type utxoDiffStore struct {
 	dirty  map[daghash.Hash]struct{}
 	loaded map[daghash.Hash]*blockUTXODiffData
 	mtx    *locks.PriorityMutex
+
+	// writeBuffer is a buffer to be used in flushToDB.
+	// It's used to avoid needless allocations/grows
+	// while writing to the utxoDiffStore DB bucket.
+	writeBuffer *bytes.Buffer
```
svarogg: Maybe move it closer to flushToDB, since it's only used there?

blockdag/utxodiffstore.go
```diff
@@ -177,12 +185,23 @@ func (diffStore *utxoDiffStore) clearDirtyEntries() {
...
 // dbStoreDiffData stores the UTXO diff data to the database.
 // This overwrites the current entry if there exists one.
-func dbStoreDiffData(dbTx database.Tx, hash *daghash.Hash, diffData *blockUTXODiffData) error {
-	serializedDiffData, err := serializeBlockUTXODiffData(diffData)
+func dbStoreDiffData(dbTx database.Tx, buffer *bytes.Buffer, hash *daghash.Hash, diffData *blockUTXODiffData) error {
+	// To avoid a ton of allocs, reuse the given buffer.
+	// We expect the buffer, in most cases, to already
+	// be large enough to accommodate the serialized data
+	// without growing.
+	bufferFrom := buffer.Len()
```
svarogg: Didn't we say we want to reset the buffer so that we don't need to do the bufferFrom stuff?

blockdag/utxoio.go
```diff
@@ -19,26 +19,25 @@ import (
 // 	hasDiffChild | Boolean   | Indicates if a diff child exist
 //  diffChild    | Hash      | The diffChild's hash. Empty if hasDiffChild is true.
 //  diff		 | UTXODiff  | The diff data's diff
-func serializeBlockUTXODiffData(diffData *blockUTXODiffData) ([]byte, error) {
-	w := &bytes.Buffer{}
+func serializeBlockUTXODiffData(buffer *bytes.Buffer, diffData *blockUTXODiffData) error {
```
svarogg: I think we want any function that doesn't need the buffer as bytes.Buffer to get it as io.Writer

PR #607, opened 2020-01-27
[NOD-636] Prevent db corruption on crash.

dereeno:


blockdag/dag.go
```diff
@@ -1987,7 +1987,8 @@ func New(config *Config) (*BlockDAG, error) {
 
 	if genesis == nil {
 		genesisBlock := util.NewBlock(dag.dagParams.GenesisBlock)
-		isOrphan, isDelayed, err := dag.ProcessBlock(genesisBlock, BFNone)
+		var isOrphan, isDelayed bool
```
svarogg: Please add a comment explaining why this is required

blockdag/dag.go
```diff
@@ -1987,6 +1987,8 @@ func New(config *Config) (*BlockDAG, error) {
 
 	if genesis == nil {
 		genesisBlock := util.NewBlock(dag.dagParams.GenesisBlock)
+		// This is prevent creation of a new err variable unintentionally,
```
svarogg: Typo: This is **to** prevent.

IMHO also the "This is" is a bit ambigous.

I'd revise to something like "To prevent the creation of a new err variable unintentionally, so the defered function above could read err - declare isOrphan and isDelayed explicitly"

PR #609, opened 2020-01-29
[NOD-647] Create a default config file even if the sample default config file is missing

stasatdaglabs:


config/config.go
```diff
@@ -808,25 +808,40 @@ func createDefaultConfigFile(destinationPath string) error {
...
-	defer src.Close()
+	defer dest.Close()
 
-	dest, err := os.OpenFile(destinationPath,
-		os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
+	// If the sample config file is missing because e.g. kaspad was
+	// installed using go install, simply create the destination
+	// file and write the RPC credentials into it as is.
+	if _, err := os.Stat(sampleConfigPath); os.IsNotExist(err) {
+		lines := []string{rpcUserString, rpcPassString}
```
svarogg: Just write rpcUser\nrpcPass and be done with it

PR #621, opened 2020-02-03
[NOD-545] Remove headers first related logic

someone235:


server/rpc/handle_get_headers.go
```diff
@@ -7,6 +7,8 @@ import (
 	"github.com/kaspanet/kaspad/util/daghash"
 )
 
+const getHeadersMaxHeaders = 2000
```
svarogg: I think we can unify getHeadersMaxHeaders and getTopHeadersMaxHeaders.

Also - add parameter to GetHeaders and GetTopHeaders commands that says maxNumber.
Set default = getHeadersMaxHeaders

PR #623, opened 2020-02-04
[NOD-727] Do not allow delayed blocks from RPC.

dereeno:


blockdag/process.go
```diff
@@ -45,6 +45,9 @@ const (
 	// in the block index but was never fully processed
 	BFWasStored
 
+	// BFBlockFromRPC is set to indicate that a block was submitted through RPC.
+	BFBlockFromRPC
```
someone235: Can you change to BFDisallowDelay?

blockdag/process.go
```diff
@@ -146,6 +149,7 @@ func (dag *BlockDAG) ProcessBlock(block *util.Block, flags BehaviorFlags) (isOrp
 func (dag *BlockDAG) processBlockNoLock(block *util.Block, flags BehaviorFlags) (isOrphan bool, isDelayed bool, err error) {
 	isAfterDelay := flags&BFAfterDelay == BFAfterDelay
 	wasBlockStored := flags&BFWasStored == BFWasStored
+	wasBlockSubmittedThroughRPC := flags&BFBlockFromRPC == BFBlockFromRPC
```
someone235: Change to `allowDelay` (and reverse condition)

blockdag/error.go
```diff
@@ -217,6 +217,10 @@ const (
 	// ErrInvalidParentsRelation indicates that one of the parents of a block
 	// is also an ancestor of another parent
 	ErrInvalidParentsRelation
+
+	// ErrDelayedBlockisNotAllowed indicates that a block with a delayed timestamp was
+	// submitted and should be rejected.
+	ErrDelayedBlockisNotAllowed
```
someone235: BlockisNotAllowed->BlockIsNotAllowed

blockdag/error.go
```diff
@@ -217,6 +217,10 @@ const (
 	// ErrInvalidParentsRelation indicates that one of the parents of a block
 	// is also an ancestor of another parent
 	ErrInvalidParentsRelation
+
+	// ErrDelayedBlockisNotAllowed indicates that a block with a delayed timestamp was
+	// submitted and should be rejected.
```
someone235: instead of saying "and should be rejected" maybe write "in a context that it shouldn't be allowed", or simply write "with BFDisallowDelay flag raised" or something like this?

blockdag/process.go
```diff
@@ -174,6 +179,11 @@ func (dag *BlockDAG) processBlockNoLock(block *util.Block, flags BehaviorFlags)
 			return false, false, err
 		}
 
+		if delay != 0 && disallowDelay {
+			str := fmt.Sprintf("Cannot process blocks beyond the allowed time offset %s", blockHash)
```
someone235: Maybe clarify that it's not allowed only because BFDisallowDelay?

blockdag/process.go
```diff
@@ -178,9 +179,9 @@ func (dag *BlockDAG) processBlockNoLock(block *util.Block, flags BehaviorFlags)
 			return false, false, err
 		}
 
-		if delay != 0 && wasBlockSubmittedThroughRPC {
-			str := fmt.Sprintf("Cannot process blocks from RPC beyond the allowed time offset %s", blockHash)
-			return false, true, ruleError(ErrDelayedBlockFromRPC, str)
+		if delay != 0 && disallowDelay {
+			str := fmt.Sprintf("Cannot process blocks beyond the allowed time offset while the DisallowDelay flag is raised %s", blockHash)
```
someone235: Why not call it `BFDisallowDelay`?

PR #624, opened 2020-02-06
[NOD-717] Fix nodes getting stuck in an infinite loop in addrManager.getAddress

stasatdaglabs:


connmgr/connmanager.go
```diff
@@ -429,6 +428,13 @@ out:
 	log.Trace("Connection handler done")
 }
 
+// ResetFailedAttempts sets the failed connection attempts counter
+// to zero. This method is expected to be called after every time
+// a peer had been successfully connected and marked as good.
+func (cm *ConnManager) ResetFailedAttempts() {
```
someone235: this name exposes implementation details. Instead I would call it "ConnectionRequestComplete" or something like this

PR #625, opened 2020-02-11
[NOD-747] Change FinalityInterval to be 24 hours, isCurrent to be true if the DAG's time is less than 12 hours than the present, and change MaxInvPerMsg to be 1 << 17

someone235:


blockdag/dag.go
```diff
@@ -27,6 +27,8 @@ const (
 	// maxOrphanBlocks is the maximum number of orphan blocks that can be
 	// queued.
 	maxOrphanBlocks = 100
+
+	isDAGCurrentMaxDiff = 12 * time.Hour
```
stasatdaglabs: TargetTimePerBlock and FinalityInterval are in dagparams. I argue that this should be in dagparams too.

someone235:

someone235:

someone235:

PR #626, opened 2020-02-12
[NOD-744] Wrap go routines with spawn

dereeno:


server/rpc/rpcserver.go
```diff
@@ -623,12 +623,12 @@ func (s *Server) Start() {
 
 	for _, listener := range s.cfg.Listeners {
 		s.wg.Add(1)
-		go func(listener net.Listener) {
+		spawn(func() {
 			log.Infof("RPC server listening on %s", listener.Addr())
 			httpServer.Serve(listener)
 			log.Tracef("RPC listener done for %s", listener.Addr())
 			s.wg.Done()
-		}(listener)
```
someone235: This is actually needed

svarogg: @someone235 net.Listener is an interface, so no, it's not doing copying, so no problem.

connmgr/connmanager.go
```diff
@@ -598,9 +598,12 @@ func (cm *ConnManager) Start() {
 	// provided a callback to be invoked when connections are accepted.
 	if cm.cfg.OnAccept != nil {
 		for _, listner := range cm.cfg.Listeners {
```
svarogg: Typo (not yours, but still please fix): listner -> listener

connmgr/connmanager.go
```diff
@@ -594,8 +598,13 @@ func (cm *ConnManager) Start() {
...
+			// Declaring this variable is necessary so the anonymous function below will use it
+			// rather then the last element in the for loop.
+			var l = listner
```
svarogg: Mike L.:
> —> Without referring to the variable declaration convention, using "l" as a variable is not recommended and is disallowed by many conventions, in order not to confuse it with "1"

PR #642, opened 2020-02-26
[NOD-664] Remove version from everything inside kaspad/cmd - use kaspad version instead

someone235:


someone235:

PR #643, opened 2020-02-26
[NOD-552] Add NormalizeRPCServerAddress and use it where needed

someone235:


someone235:

dagconfig/params.go
```diff
@@ -185,6 +186,16 @@ type Params struct {
 	HDCoinType uint32
 }
 
+// NormalizeRPCServerAddress returns addr with the current network default
+// port appended if there is not already a port specified.
+func (p *Params) NormalizeRPCServerAddress(addr string) string {
+	_, _, err := net.SplitHostPort(addr)
```
svarogg: Add a comment about blabla, here and any other way where we might use SplitHostPort in a similar way

util/network/network.go
```diff
@@ -6,22 +6,38 @@ import (
...
-	return removeDuplicateAddresses(addrs)
+	return removeDuplicateAddresses(addrs), nil
 }
 
 // NormalizeAddress returns addr with the passed default port appended if
 // there is not already a port specified.
-func NormalizeAddress(addr, defaultPort string) string {
+func NormalizeAddress(addr, defaultPort string) (string, error) {
 	_, _, err := net.SplitHostPort(addr)
+	// net.SplitHostPort returns an error
```
svarogg: Why so short lines?

PR #650, opened 2020-03-05
[NOD-806] Exit early after panic

someone235:


util/panics/panics.go
```diff
@@ -1,69 +1,61 @@
...
 )
 
 // HandlePanic recovers panics, log them, runs an optional panicHandler,
 // and then initiates a clean shutdown.
-func HandlePanic(log logs.Logger, goroutineStackTrace []byte, panicHandler func()) {
-	if err := recover(); err != nil {
+func HandlePanic(backend *logs.Backend, goroutineStackTrace []byte) {
+	const panicSubsystem = "PANC"
+
+	log := backend.Logger(panicSubsystem)
```
stasatdaglabs: Please move the above two lines to after we've checked that there really is a panic.

util/panics/panics.go
```diff
@@ -1,69 +1,60 @@
...
-	if err := recover(); err != nil {
+func HandlePanic(backend *logs.Backend, goroutineStackTrace []byte) {
+	err := recover()
+	if err == nil {
+		return
+	}
+
+	panicHandlerDone := make(chan struct{})
+	go func() {
+		const panicSubsystem = "PANC"
```
svarogg: Please use subsystem of whomever threw the panic

PR #658, opened 2020-03-15
[NOD-818] Remove time adjustment

someone235:


blockdag/timesource.go
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2013-2014 The btcsuite developers
```
svarogg: I think that it's a good time to remove this copyright message.

blockdag/timesource.go
```diff
@@ -0,0 +1,32 @@
...
+type TimeSource interface {
+	// Now returns the current time.
+	Now() time.Time
+}
+
+// timeSource provides an implementation of the TimeSource interface
+// that simply returns the current local time.
+type timeSource struct{}
+
+// Ensure the timeSource type implements the TimeSource interface.
```
svarogg: This looks somewhat redundant.

I mean, it might be an ok practice to do it everywhere, but we are not doing it anywhere else, so why bother here?

blockdag/dag.go
```diff
@@ -1901,7 +1901,7 @@ type Config struct {
 	// The caller is expected to keep a reference to the time source as well
 	// and add time samples from other peers on the network so the local
 	// time is adjusted to be in agreement with other peers.
```
stasatdaglabs: This comment is no longer correct. Please adjust it accordingly.

PR #666, opened 2020-03-22
[NOD-848] optimize utxo diffs serialize allocations

someone235:


wire/common.go
```diff
@@ -391,9 +403,21 @@ func ReadVarInt(r io.Reader) (uint64, error) {
 	return rv, nil
 }
 
-// WriteVarInt serializes val to w using a variable number of bytes depending
+// WriteVarIntLittleEndian serializes val to w using a variable number of bytes depending
+// on its value using little endian encoding.
+func WriteVarIntLittleEndian(w io.Writer, val uint64) error {
```
stasatdaglabs: I'm not sure I'm a big fan of having two functions, each for one endianness. Compare the following usage:
WriteVarIntLittleEndian(w, num) x 10
vs.
var byteOrder = LittleEndian
WriteVarInt(w, num, byteOrder) x 10

It's much faster to refactor the latter, and there's no sacrifice in readability.

blockdag/utxoio.go
```diff
@@ -217,97 +217,107 @@ func serializeMultiset(w io.Writer, ms *ecc.Multiset) error {
...
+	}
 
-	// Serialize the header code followed by the compressed unspent
-	// transaction output.
-	serialized := make([]byte, size)
-	offset := putVLQ(serialized, headerCode)
-	offset += putCompressedTxOut(serialized[offset:], uint64(entry.Amount()),
-		entry.ScriptPubKey())
+	if useCompression {
+		return putCompressedTxOut(w, entry.Amount(), entry.ScriptPubKey())
```
stasatdaglabs: Please measure how much this compression actually saves us. With vs. without for ~finality utxoentries.
If it's less than 100mb then consider getting rid of it.

blockdag/compress.go
```diff
@@ -238,185 +121,129 @@ func isPubKey(script []byte) (bool, []byte) {
...
 // it will panic.
-func putCompressedScript(target, scriptPubKey []byte) int {
+func putCompressedScript(w io.Writer, scriptPubKey []byte) error {
 	// Pay-to-pubkey-hash script.
 	if valid, hash := isPubKeyHash(scriptPubKey); valid {
-		target[0] = cstPayToPubKeyHash
-		copy(target[1:21], hash)
-		return 21
+		if len(hash) != pubKeyHashLen {
+			panic(errors.Errorf("public key hash is not %d bytes", pubKeyHashLen))
```
stasatdaglabs: I think I prefer that you not make this check rather than panicking explicitly.

blockdag/compress.go
```diff
@@ -238,185 +121,129 @@ func isPubKey(script []byte) (bool, []byte) {
...
-			copy(target[1:33], serializedPubKey[1:33])
-			return 33
-		case 0x04:
-			// Encode the oddness of the serialized pubkey into the
-			// compressed script type.
-			target[0] = pubKeyFormat | (serializedPubKey[64] & 0x01)
-			copy(target[1:33], serializedPubKey[1:33])
-			return 33
+		if len(hash) != scriptHashLen {
+			panic(errors.Errorf("script hash is not %d bytes", scriptHashLen))
```
stasatdaglabs: I think I prefer that you not make this check rather than panicking explicitly.

PR #669, opened 2020-03-26
[NOD-855] Save ECMH for block utxo and not diff utxo

someone235:


blockdag/dag.go
```diff
@@ -593,6 +594,126 @@ func (dag *BlockDAG) connectBlock(node *blockNode,
...
+}
+
+func (node *blockNode) selectParentMultiSet(dag *BlockDAG) (*ecc.Multiset, error) {
+	if node.isGenesis() {
+		return ecc.NewMultiset(ecc.S256()), nil
+	}
+	return dag.multisetStore.multisetByBlockNode(node.selectedParent)
+}
+
+func (node *blockNode) acceptedSelectedParentMultiset(dag *BlockDAG, acceptanceData MultiBlockTxsAcceptanceData) (*ecc.Multiset, error) {
```
svarogg: Add a comment explaining what this does.
Try finding a better name if possible

blockdag/dag.go
```diff
@@ -593,6 +594,126 @@ func (dag *BlockDAG) connectBlock(node *blockNode,
...
+	ms, err := node.selectParentMultiSet(dag)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, blockAcceptanceData := range acceptanceData {
+		if !blockAcceptanceData.BlockHash.IsEqual(node.selectedParent.hash) {
```
svarogg: Better find the selectedParent's acceptance data using MultiBlockTxsAcceptanceData.FindAcceptanceData, than run only on it.

blockdag/dag.go
```diff
@@ -593,6 +594,126 @@ func (dag *BlockDAG) connectBlock(node *blockNode,
...
+		}
+	}
+	return ms, nil
+}
+
+func addTxToMultiset(ms *ecc.Multiset, tx *wire.MsgTx, pastUTXO UTXOSet, blockBlueScore uint64) (*ecc.Multiset, error) {
+	isCoinbase := tx.IsCoinBase()
+	if !isCoinbase {
+		for _, txIn := range tx.TxIn {
+			outpoint := *wire.NewOutpoint(&txIn.PreviousOutpoint.TxID, txIn.PreviousOutpoint.Index)
```
svarogg: Why can't you simply use txIn.PreviousOutpoint?

blockdag/mining.go
```diff
@@ -34,25 +37,37 @@ func (dag *BlockDAG) BlockForMining(transactions []*util.Tx) (*wire.MsgBlock, er
 		msgBlock.AddTransaction(tx.MsgTx())
 	}
 
-	utxoWithTransactions, err := dag.UTXOSet().WithTransactions(msgBlock.Transactions, UnacceptedBlueScore, false)
+	ms, err := dag.NextBlockMultiset(transactions)
```
svarogg: Please use a longer name than ms.
In other places where the multiset is the key thing, I didn't mind too much(although still better to call it multiSet), but here this is more important because ms is no longer the center of all things, and the word ms can mean a ton of things

blockdag/multisetio.go
```diff
@@ -0,0 +1,52 @@
...
+	"math/big"
+)
+
+const multisetPointSize = 32
+
+// serializeMultiset serializes an ECMH multiset. The serialization
+// is done by taking the (x,y) coordinnates of the multiset point and
+// padding each one of them with 32 byte (it'll be 32 byte in most
+// cases anyway except one of the coordinates is zero) and writing
+// them one after the other.
```
stasatdaglabs: Consider just writing the format instead of describing it in English:

> // Format: <x (32 bytes)><y (32 bytes)>

blockdag/dag.go
```diff
@@ -867,7 +994,7 @@ func (dag *BlockDAG) TxsAcceptedByBlockHash(blockHash *daghash.Hash) (MultiBlock
 // It returns the diff in the virtual block's UTXO set.
```
stasatdaglabs: Please update the above comment.

blockdag/dag.go
```diff
@@ -593,6 +594,126 @@ func (dag *BlockDAG) connectBlock(node *blockNode,
...
+					return nil, err
+				}
+
+				ms, err = addUTXOToMultiset(ms, acceptedEntry, &outpoint)
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
+		break
```
stasatdaglabs: This looks wrong. If we break here then the loop over acceptanceData is equivalent to taking the first element.

blockdag/dag.go
```diff
@@ -593,6 +594,126 @@ func (dag *BlockDAG) connectBlock(node *blockNode,
 	return chainUpdates, nil
 }
 
+func (node *blockNode) calcMultiset(dag *BlockDAG, transactions []*util.Tx, acceptanceData MultiBlockTxsAcceptanceData, selectedParentUTXO, pastUTXO UTXOSet) (*ecc.Multiset, error) {
```
stasatdaglabs: Please add a comment.

PR #670, opened 2020-03-26
[NOD-876] Replace ecc with go-secp256k1 for public keys

elichai:
This is no longer blocked on anything :)
After this the `ecc` package will only be used in `util/hdkeychain`. which isn't because we have no wallet.


txscript/opcode.go
```diff
@@ -2214,13 +2212,12 @@ func opcodeCheckMultiSig(op *parsedOpcode, vm *Engine) error {
...
 
 			// Parse the signature.
-			var err error
-			parsedSig, err = ecc.ParseSignature(signature)
-
-			sigInfo.parsed = true
+			parsedSig, err = secp256k1.DeserializeSchnorrSignatureFromSlice(signature)
 			if err != nil {
-				continue
+				return err
```
svarogg: How sure are you this is a good idea?

To me it looks correct as well, but this is the sort of changes I'd like to do really carefully, and preferably not as part of bigger changes.

So unless you know for an absolute fact that this is the correct behavior, I'd rather revert for now and then check it carefully at more peaceful times.

txscript/sign.go
```diff
@@ -5,29 +5,30 @@
...
+	"github.com/kaspanet/go-secp256k1"
 	"github.com/pkg/errors"
 
 	"github.com/kaspanet/kaspad/dagconfig"
-	"github.com/kaspanet/kaspad/ecc"
 	"github.com/kaspanet/kaspad/util"
 	"github.com/kaspanet/kaspad/wire"
 )
 
 // RawTxInSignature returns the serialized ECDSA signature for the input idx of
```
stasatdaglabs: ECDSA -> Schnorr

txscript/sign_test.go
```diff
@@ -166,7 +178,6 @@ func TestSignTxOutput(t *testing.T) {
 				mkGetKey(map[string]addressToKey{
 					address.EncodeAddress(): {key, false},
 				}), mkGetScript(nil), nil); err != nil {
-				t.Error(err)
```
stasatdaglabs: Why did you remove the error log from here?

txscript/opcode.go
elichai: It's good that you pointed this out.
I should really go over the consensus code again to check for correctness instead of workness

elichai: it is actually *wrong* :) this will break op_checkmultisig

svarogg: By "this" you mean "what have been there up until now" or "my change"?

elichai: arghh not really. sorry for spamming let's get this straight:
if you have: P1,P2,P3 in a 1/3 multisig, and you want to sign with the third key, then you need to provide: 
0*32,0*32,SIG. 
meaning the first 2 signatures will be invalid. but here the error is just about the length which is checked the line above so can never fail. non the less, this is a bad idea because if the behavior of `DeserializeSchnorrSignatureFromSlice` will ever change to actually do some verification then this will break op_checkmultisig.

txscript/sign_test.go
elichai: by mistake, this line whole expression was really hard to understand so I de-structured it and accidentally removed the error

txscript/sign.go
elichai: Done

PR #675, opened 2020-03-30
[NOD-867] Migrate database logic in blockdag/dagio.go to dbaccess

stasatdaglabs:


blockdag/dagio.go
```diff
@@ -360,151 +270,135 @@ func (dag *BlockDAG) removeDAGState() error {
...
 
-			// Deserialize the utxo entry
-			entry, err := deserializeUTXOEntry(bytes.NewReader(cursor.Value()))
-			if err != nil {
-				return err
-			}
+	// Determine how many UTXO entries will be loaded into the index so we can
+	// allocate the right amount.
+	var utxoEntryCount int32
+	cursor, err := dbaccess.UTXOSetCursor(dbaccess.NoTx())
```
svarogg: I might be wrong, but it seems to me that not pre-allocating will have a lower overhead than double-iterating over the cursor, and this pre-allocation defenitely adds complexity.

Also, it's in the start-up, so as long as the overhead isn't overly terrible, I don't mind it too much.

On another note: maybe it's possible to ask levelDB to check for length, without actually iterating over the whole thing?

stasatdaglabs: > On another note: maybe it's possible to ask levelDB to check for length, without actually iterating over the whole thing?

Doesn't seem to support such a thing, I'm afraid.

PR #678, opened 2020-03-30
[NOD-881] Don't recalculate subtreesize for children

someone235:


blockdag/reachability_test.go
```diff
@@ -609,6 +609,45 @@ func TestReindexIntervalErrors(t *testing.T) {
...
+		}
+
+		currentTreeNode = childTreeNode
+	}
+
+	remainingIntervalBefore := *root.remainingInterval
+	// After we added subTreeSize nodes, adding the next
+	// node should lead to a reindex from root.
+	fullReindexTriggeringNode := newReachabilityTreeNode(&blockNode{})
+	b.StartTimer()
```
svarogg: A little bit better approach is to do here b.ResetTimer(), and than you don't need b.StopTimer() in the beginning.

PR #679, opened 2020-03-30
[NOD-857] Add generalized profiler package and use it everwhere

svarogg:


util/profiling/log.go
```diff
@@ -0,0 +1,10 @@
+package profiling
+
+import (
+	"github.com/kaspanet/kaspad/logger"
+	"github.com/kaspanet/kaspad/util/panics"
+)
+
+var log, _ = logger.Get(logger.SubsystemTags.PROF)
+var spawn = panics.GoroutineWrapperFunc(log)
+var spawnAfter = panics.AfterFuncWrapperFunc(log)
```
someone235: This is not used

PR #682, opened 2020-03-31
[NOD-866] Migrate database logic in blockdag/indexers package to dbaccess

stasatdaglabs:


blockdag/dag.go
```diff
@@ -1933,6 +1933,20 @@ func (dag *BlockDAG) SubnetworkID() *subnetworkid.SubnetworkID {
 	return dag.subnetworkID
 }
 
+// ForEachHash runs the given fn on every hash that's currently known to
+// the DAG.
+//
+// This function is NOT safe for cuncurrent access.
```
svarogg: How do we ensure it is not run concurrently?

I can think of two ways:
1. Create a dedicated mutex.
2. Assign it to one of the existing mutexes (I guess DagMutex is appropriate?).

I'd really prefer option 2.
This just means to add to the comment "This function must be called with the BlablaMutex held for Reads/Writes"

P.S.
But before deleting - see comment I wrote on where you use ForEachHash

blockdag/indexers/acceptanceindex.go
```diff
@@ -43,122 +30,71 @@ func NewAcceptanceIndex() *AcceptanceIndex {
...
+	defer dbTx.RollbackUnlessClosed()
 
-// Recover is invoked when the indexer wasn't turned on for several blocks
-// and the indexer needs to close the gaps.
-//
-// This is part of the Indexer interface.
-func (idx *AcceptanceIndex) Recover(dbTx database.Tx, currentBlockID, lastKnownBlockID uint64) error {
-	for blockID := currentBlockID + 1; blockID <= lastKnownBlockID; blockID++ {
-		hash, err := blockdag.DBFetchBlockHashByID(dbTx, currentBlockID)
+	err = idx.dag.ForEachHash(func(hash daghash.Hash) error {
```
svarogg: Functional programming of this style is not idiomatic in go.
Why can't we just do 
```
for hash := range dag.index.index{
    // whatever you want to do with each hash
}
```

dbaccess/acceptanceindex.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return acceptanceData, nil
+}
+
+func acceptanceIndexKey(hash *daghash.Hash) []byte {
+	return acceptanceIndexBucket.Key(hash[:])
+}
+
+// ClearAcceptanceIndex completely removes all acceptanceData entries.
+func ClearAcceptanceIndex() error {
```
svarogg: Maybe Drop instead of Clear, to stay true to the lingo in the flag?
(Or if you don't like Drop for any reason, I'd suggest Remove, both here and in the flag).

Clear sounds a bit like "clear so that you can fill it back up with new data"

dbaccess/acceptanceindex.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return err
+	}
+	defer context.RollbackUnlessClosed()
+	accessor, err := context.accessor()
+	if err != nil {
+		return err
+	}
+
+	// Collect all of the keys
```
svarogg: I'd mention the reason for the collect in the comment.

blockdag/indexers/acceptanceindex.go
stasatdaglabs: We spoke about this and decided to leave it as-is.

PR #683, opened 2020-04-01
[NOD-868] Delete the old database package

stasatdaglabs:


blockdag/dagio.go
```diff
@@ -302,85 +304,84 @@ func (dag *BlockDAG) initDAGState() error {
 	}
 
 	// Attempt to load the DAG state from the database.
-	return dag.db.View(func(dbTx database.Tx) error {
-		// Initialize the reachability store
-		log.Infof("Loading reachability data...")
-		err = dag.reachabilityStore.init(dbaccess.NoTx())
+
+	// Initialize the reachability store
```
svarogg, line 308: IMHO these comments are redundant (especially that we have a log that says the same thing right below)

svarogg: On second thought, there are other stages where the code isn't as clear, and there a comment is in place, so for the sake of uniformity might want to leave these comments as well - your call

blockdag/dagio.go
```diff
@@ -302,85 +304,84 @@ func (dag *BlockDAG) initDAGState() error {
...
+	}
+	dag.virtual.SetTips(tips)
+
+	// Set the last finality point
+	dag.lastFinalityPoint = dag.index.LookupNode(dagState.LastFinalityPoint)
+	dag.finalizeNodesBelowFinalityPoint(false)
+
+	// Go over any unprocessed blockNodes and process them now.
+	for _, node := range unprocessedBlockNodes {
+		// Check to see if the block exists in the block DB. If it
```
svarogg, line 346: Just a thought:
These are unprocessed block, theoretically we can boot and work without them.
So let us assume worst case: unprocessed blocks are corrupted AF.
In this case there can be two options:
A) Something else, more critical in the DB is corrupted AF as well.
B) Nothing else is corrupted.

In case A - we'll crash over `something else` pretty soon.
In case B - why not start working without the unprocessed blocks?

svarogg: P.S.
The other side of this argument is:
If something in the DB is corrupted, it might cause very unpredictable behavior in places that might be harder to detect, so we want to crush the moment we have a first indication of the crash.

I see the merit in both sides, and would love to hear your and @someone235 opinions on that.

database/README.md
```diff
@@ -4,29 +4,41 @@ database
...
-- Nested buckets
-- Iteration support including cursors with seek capability
-- Supports registration of backend databases
-- Comprehensive test coverage
+Database
+--------
+This defines the interface of a database that can begin transactions, open cursors,
+and close itself.
 
+Important: This is not part of the DataAccessor interface because the Transaction
```
svarogg, line 29: I think this paragraph is a bit redundant for a README. Can move it to the DataAccessor interface if it's not there yet.

database/README.md
```diff
@@ -4,29 +4,41 @@ database
...
+interface implements it as well. Were we to merge Database with DataAccessor,
+implementors of the Transaction interface would be forced to implement methods
+such as Begin and Close, which is undesirable.
+
+Transaction
+-----------
+This defines the interface of a generic kaspad database transaction.
+
+Note: Transactions provide data consistency over the state of the database as it was
+when the transaction started. There is NO guarantee that if one puts data into the
```
svarogg, line 39: "There is NO guarantee..."
If I remember correctly there's a guarantee that it won't happen.

database/doc.go
```diff
@@ -1,85 +1,39 @@
...
-nested buckets as well as iterate through them.
+Database
 
-The Get, Put, and Delete functions work with key/value pairs, while the Bucket,
-CreateBucket, CreateBucketIfNotExists, and DeleteBucket functions work with
-buckets. The ForEach function allows the caller to provide a function to be
-called with each key/value pair and nested bucket in the current bucket.
+This defines the interface of a database that can begin transactions, open cursors,
+and close itself.
+Important: this is not part of the DataAccessor interface because the Transaction
```
svarogg, line 23: Here same as in README

database/doc.go
```diff
@@ -1,85 +1,39 @@
...
+Transaction
 
-As discussed above, all of the functions which are used to manipulate key/value
-pairs and nested buckets exist on the Bucket interface. The root metadata
-bucket is the upper-most bucket in which data is stored and is created at the
-same time as the database. Use the Metadata function on the Tx interface
-to retrieve it.
+This defines the interface of a generic kaspad database transaction.
+Note: transactions provide data consistency over the state of the database as it was
+when the transaction started. There is NO guarantee that if one puts data into the
```
svarogg, line 32: And here

database/README.md
stasatdaglabs: This comment refers to the database package, whereas the not-happening guarantee is specific to leveldb.

database/doc.go
stasatdaglabs: This comment refers to the database package, whereas the not-happening guarantee is specific to leveldb.

blockdag/dagio.go
stasatdaglabs: We spoke about it an decided to leave it as-is.

blockdag/dagio.go
```diff
@@ -190,10 +186,16 @@ func (dag *BlockDAG) createDAGState(localSubnetworkID *subnetworkid.SubnetworkID
 // database. When the db does not yet contain any DAG state, both it and the
 // DAG state are initialized to the genesis block.
 func (dag *BlockDAG) initDAGState() error {
+	dbTx, err := dbaccess.NewTx()
```
someone235, line 189: Why do you need a transaction for this function?

PR #684, opened 2020-04-01
[NOD-874] If the node is not current call sm.restartSyncIfNeeded() on handleInvMsg

someone235:


netsync/manager.go
```diff
@@ -733,6 +739,14 @@ func (sm *SyncManager) handleInvMsg(imsg *invMsg) {
 	if err != nil {
 		log.Errorf("Failed to send invs from queue: %s", err)
 	}
+
+	if haveUnknownInvBlock && !sm.current() {
+		// If one of the  an inv message is an unknown block
+		// it is an indication that one of our peers has more
+		// up-to-date date than we, so it triggers
+		// sm.restartSyncIfNeeded().
```
stasatdaglabs, line 747: Please reword:
* is more up-to-date that us
* remove "so it triggers..."

netsync/manager.go
```diff
@@ -733,6 +739,13 @@ func (sm *SyncManager) handleInvMsg(imsg *invMsg) {
 	if err != nil {
 		log.Errorf("Failed to send invs from queue: %s", err)
 	}
+
+	if haveUnknownInvBlock && !sm.current() {
+		// If one of the  an inv message is an unknown block
```
svarogg, line 744: Double space between `the  an`. Fix quickly before Stas get's post-trauma from his war with Mr. Collins

PR #689, opened 2020-04-05
[NOD-897] Support rolling back flat-files in database transactions

stasatdaglabs:


stasatdaglabs:
We decided to tackle this issue differently when we implement pruning.

PR #692, opened 2020-04-06
[NOD-885] Use database.Key and database.Bucket instead of byte slices

someone235:


database/helpers.go
```diff
@@ -1,12 +1,44 @@
...
-// sub-buckets, and keys into a single full key-value
+// Key is a helper type meant to combine prefix
+// and keys into a single full key-value
 // database key.
+type Key struct {
+	prefix, key []byte
+}
+
+// FullKey returns the prefix concatenated to the key.
+func (k *Key) FullKey() []byte {
```
stasatdaglabs, line 17: Please consider renaming this to something like FullKeyBytes().

database/helpers.go
```diff
@@ -1,12 +1,44 @@
...
+	copy(keyPath[len(k.prefix):], k.key)
+	return keyPath
+}
+
+func (k *Key) String() string {
+	return string(k.FullKey())
+}
+
+// Key returns the key part of the key.
+func (k *Key) Key() []byte {
```
stasatdaglabs, line 29: Please consider renaming this to something like KeyBytes().

database/helpers.go
```diff
@@ -1,12 +1,44 @@
...
+// FullKey returns the prefix concatenated to the key.
+func (k *Key) FullKey() []byte {
+	keyPath := make([]byte, len(k.prefix)+len(k.key))
+	copy(keyPath, k.prefix)
+	copy(keyPath[len(k.prefix):], k.key)
+	return keyPath
+}
+
+func (k *Key) String() string {
+	return string(k.FullKey())
```
stasatdaglabs, line 25: Please see if you can think of a nicer way of printing the key.

database/helpers.go
```diff
@@ -8,33 +8,38 @@ import (
...
+	prefix, suffix []byte
 }
 
-// FullKeyBytes returns the prefix concatenated to the key.
-func (k *Key) FullKeyBytes() []byte {
-	keyPath := make([]byte, len(k.prefix)+len(k.key))
-	copy(keyPath, k.prefix)
-	copy(keyPath[len(k.prefix):], k.key)
-	return keyPath
+// Bytes returns the prefix concatenated to the key.
```
stasatdaglabs, line 17: The comment no longer makes much sense.

database/helpers.go
```diff
@@ -8,33 +8,38 @@ import (
 var separator = []byte("/")
 
 // Key is a helper type meant to combine prefix
-// and keys into a single full key-value
+// and suffix into a single full key-value
```
stasatdaglabs, line 11: Please don't use the phrase "key-value" as it means something completely different.

database/helpers.go
```diff
@@ -0,0 +1,81 @@
+package database
```
svarogg, line 1: "helpers" is very general name for this file, while the contents are pretty specific.

I'd call it "keys.go" or something. 
Preferably find a name that shows that it's both keys and buckets, but I can't think of a good one, and since buckets are just something to help build keys, I guess it's fine.

database/helpers.go
```diff
@@ -0,0 +1,81 @@
...
+func (b *Bucket) Bucket(bucketBytes []byte) *Bucket {
+	newPath := make([][]byte, len(b.path)+1)
+	copy(newPath, b.path)
+	copy(newPath[len(b.path):], [][]byte{bucketBytes})
+
+	return MakeBucket(newPath...)
+}
+
+// Key returns the key inside of the current bucket.
+func (b *Bucket) Key(key []byte) *Key {
```
svarogg, line 68: `key` is the whole thing, the parameter name should be `suffix`

PR #693, opened 2020-04-07
[NOD-822] Don't return rule errors from utxoset code

someone235:


blockdag/dag.go
```diff
@@ -1058,21 +1058,24 @@ func (dag *BlockDAG) meldVirtualUTXO(newVirtualUTXODiffSet *DiffUTXOSet) error {
 	return newVirtualUTXODiffSet.meldToBase()
 }
 
-func (node *blockNode) diffFromTxs(pastUTXO UTXOSet, transactions []*util.Tx) (*UTXODiff, error) {
-	diff := NewUTXODiff()
+func (node *blockNode) calcUTXO(pastUTXO *DiffUTXOSet, transactions []*util.Tx) (UTXOSet, error) {
```
svarogg, line 1061: `CalcUTXO` is a bit too general IMHO. I'd rename to `applyTransactionsToUTXO` or something

blockdag/dag.go
```diff
@@ -1058,21 +1058,24 @@ func (dag *BlockDAG) meldVirtualUTXO(newVirtualUTXODiffSet *DiffUTXOSet) error {
...
 		if err != nil {
 			return nil, err
 		}
-		diff, err = diff.WithDiff(txDiff)
-		if err != nil {
-			return nil, err
+
+		if !isAccepted {
+			return nil, ruleError(ErrMissingTxOut, fmt.Sprintf(
+				"Transaction %s is invalid because because at least one of its inputs doesn't have a "+
```
svarogg, line 1072: Instead of saying "because at least one of its inputs...", I'd rather have the input's outpoint specified explicitly

blockdag/dag.go
```diff
@@ -1096,11 +1099,11 @@ func (node *blockNode) verifyAndBuildUTXO(dag *BlockDAG, transactions []*util.Tx
 		return nil, nil, nil, nil, err
 	}
 
-	diffFromTxs, err := node.diffFromTxs(pastUTXO, transactions)
-	if err != nil {
-		return nil, nil, nil, nil, err
+	pastUTXODiffSet, ok := pastUTXO.(*DiffUTXOSet)
```
svarogg, line 1102: I don't really like the cast to *DiffUTXOSet, but it looks indeed necessary in the way you do things.
What was bad in the way things worked up until now?

blockdag/dag.go
```diff
@@ -1058,11 +1058,16 @@ func (dag *BlockDAG) meldVirtualUTXO(newVirtualUTXODiffSet *DiffUTXOSet) error {
 	return newVirtualUTXODiffSet.meldToBase()
 }
 
-func (node *blockNode) diffFromTxs(pastUTXO UTXOSet, transactions []*util.Tx) (*UTXODiff, error) {
+// applyAndVerifyBlockTransactionsToUTXO applies a block's transactions to its
+// given past UTXO, and verifies that there are not double spends with his past.
```
stasatdaglabs, line 1062: his -> its

stasatdaglabs: Also, not -> no

PR #695, opened 2020-04-08
[NOD-849] Database tests

stasatdaglabs:


database/ffldb/transaction_test.go
```diff
@@ -0,0 +1,564 @@
...
+
+import (
+	"bytes"
+	"github.com/kaspanet/kaspad/database"
+	"io/ioutil"
+	"strings"
+	"testing"
+)
+
+func TestKeyValueTransactionCommit(t *testing.T) {
```
someone235, line 11: Can you think of a better name?

database/ffldb/transaction_test.go
```diff
@@ -0,0 +1,564 @@
...
+			"unexpectedly failed: %s", err)
+	}
+	if !bytes.Equal(existingValue, value2) {
+		t.Fatalf("TestKeyValueTransactionCommit: Get "+
+			"returned unexpected value. Want: %s, got: %s",
+			string(value2), string(existingValue))
+	}
+}
+
+func TestKeyValueTransactionRollback(t *testing.T) {
```
someone235, line 157: Can you think of a better name?

database/ffldb/transaction_test.go
```diff
@@ -0,0 +1,564 @@
...
+		}
+	}()
+
+	// Begin a new transaction
+	dbTx, err := db.Begin()
+	if err != nil {
+		t.Fatalf("TestTransactionRollbackUnlessClosed: Begin "+
+			"unexpectedly failed: %s", err)
+	}
+	defer func() {
```
someone235, line 467: Do we really needs this defer?

database/ffldb/ldb/transaction_test.go
```diff
@@ -0,0 +1,141 @@
...
+			defer func() {
+				err := dbTx.RollbackUnlessClosed()
+				if err != nil {
+					t.Fatalf("TestTransactionCloseErrors: RollbackUnlessClosed "+
+						"unexpectedly failed: %s", err)
+				}
+			}()
+
+			// Close the transaction
+			err = dbTx.Commit()
```
svarogg, line 113: It might be a good idea to repeat same test-case with dbTx.Rollback, so that we make sure that it's not something only dbTx.Commit does that makes the thing work the way it should

database/ffldb/transaction_test.go
svarogg: And if not - add comment explaining test

database/ffldb/transaction_test.go
```diff
@@ -0,0 +1,564 @@
...
+	"bytes"
+	"github.com/kaspanet/kaspad/database"
+	"io/ioutil"
+	"strings"
+	"testing"
+)
+
+func TestKeyValueTransactionCommit(t *testing.T) {
+	// Open a test db
+	path, err := ioutil.TempDir("", "TestKeyValueTransactionCommit")
```
svarogg, line 13: Most tests here create a temp database, consider extract function.
Can work similar to DAGSetup - a.k.a. return a cleanup function

database/ffldb/transaction_test.go
```diff
@@ -0,0 +1,564 @@
...
+	// Put a new value
+	key2 := database.MakeBucket().Key([]byte("key2"))
+	value2 := []byte("value2")
+	err = dbTx.Put(key2, value2)
+	if err != nil {
+		t.Fatalf("TestKeyValueTransactionRollback: Put "+
+			"unexpectedly failed: %s", err)
+	}
+
+	// Commit the transaction
```
svarogg, line 239: Copypasta: Commit -> Rollback

database/ffldb/transaction_test.go
```diff
@@ -0,0 +1,564 @@
...
+			defer func() {
+				err := dbTx.RollbackUnlessClosed()
+				if err != nil {
+					t.Fatalf("TestTransactionCloseErrors: RollbackUnlessClosed "+
+						"unexpectedly failed: %s", err)
+				}
+			}()
+
+			// Close the transaction
+			err = dbTx.Commit()
```
svarogg, line 411: Here too, consider running both Commit and Rollback

database/ffldb/ldb/cursor_test.go
```diff
@@ -0,0 +1,284 @@
...
+			if !strings.Contains(err.Error(), expectedErrContainsString) {
+				t.Fatalf("TestCursorCloseErrors: %s "+
+					"returned wrong error. Want: %s, got: %s",
+					test.name, expectedErrContainsString, err)
+			}
+		}()
+	}
+}
+
+func TestCursorCloseFirstAndNext(t *testing.T) {
```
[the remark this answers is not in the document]
stasatdaglabs: First and Next don't return errors when they're closed.

database/ffldb/ldb/cursor_test.go
```diff
@@ -0,0 +1,284 @@
...
+	defer func() {
+		err := db.Close()
+		if err != nil {
+			t.Fatalf("TestCursorCloseFirstAndNext: Close "+
+				"unexpectedly failed: %s", err)
+		}
+	}()
+
+	// Write some data to the database
+	for i := 0; i < 10; i++ {
```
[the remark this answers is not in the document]
stasatdaglabs: Such a loop occurs exactly twice. For the sake of simplicity, I'd prefer to keep the loops where they currently are.

database/ffldb/transaction_test.go
```diff
@@ -8,87 +8,93 @@ import (
 	"testing"
 )
 
-func TestKeyValueTransactionCommit(t *testing.T) {
-	// Open a test db
-	path, err := ioutil.TempDir("", "TestKeyValueTransactionCommit")
+func prepareDatabaseForTest(t *testing.T, testName string) (db database.Database, teardownFunc func()) {
```
svarogg, line 11: I think there might me a more appropriate place for this function rather transaction_test.go.
Even if file of it's own.
Even if this is currently used only in transaction_test.

database/ffldb/transaction_test.go
```diff
@@ -374,43 +364,160 @@ func TestTransactionCloseErrors(t *testing.T) {
...
+				if err != nil {
+					t.Fatalf("TestTransactionCommitErrors: %s "+
+						"unexpectedly failed: %s", test.name, err)
+				}
+			}
+		}()
+	}
+}
+
+func TestTransactionRollbackErrors(t *testing.T) {
```
svarogg, line 415: Can't we somehow combine TestTransactionCommitErrors and TestTransactionRollbackErrors?
They both test for basically the same thing - that after tx close you can't do anything? Am I right?

PR #700, opened 2020-04-13
[NOD-858] Don't switch sync peer if the syncing process hasn't yet started with the current sync peer

someone235:


peer/peer.go
```diff
@@ -775,6 +784,7 @@ func (p *Peer) PushAddrMsg(addresses []*wire.NetAddress, subnetworkID *subnetwor
 //
 // This function is safe for concurrent access.
 func (p *Peer) PushGetBlockLocatorMsg(highHash, lowHash *daghash.Hash) {
+	p.shouldSendBlockLocator = true
```
svarogg, line 787: I'd rename to `isBlockLocatorPending` or `wasBlockLocatorRequested` something.
Because shouldSendBlockLocator sound like "should we send blockLocator".

Also, I don't see anywhere that you set it to false. Shouldn't this happen on handling block locator or something?

netsync/manager.go
```diff
@@ -427,24 +431,38 @@ func (sm *SyncManager) current() bool {
...
-			}()
-			if isWaitingForBlocks {
-				return
-			}
-		}
+	sm.syncPeerLock.Lock()
+	defer sm.syncPeerLock.Unlock()
+	shouldReplaceSyncPeer, err := sm.shouldReplaceSyncPeer()
+	if err != nil {
+		panic(err)
```
stasatdaglabs, line 438: Please write a short comment justifying the panic.

svarogg, line 438: I'm grinding my teeth here, but I guess fine, if syncPeer doesn't have peerState, it's indeed a grievous problem that should never happen.

However, in that case maybe move the panic closer to where that happens?

netsync/manager.go
```diff
@@ -427,24 +431,38 @@ func (sm *SyncManager) current() bool {
...
 	sm.syncPeer = nil
 	sm.startSync()
 }
 
+func (sm *SyncManager) shouldReplaceSyncPeer() (bool, error) {
+	if sm.syncPeer == nil {
+		return true, nil
+	}
+
+	syncPeerState, exists := sm.peerStates[sm.syncPeer]
```
svarogg, line 454: Actually, in case we panic anyway, why not let sm.peerStates[sm.syncPeer] panic to start with?
I mean, just do `syncPeerState := sm.peerStates[sm.syncPeer]`.

If you assume this never happens anyway, then you are saving some code complexity, and if you think this should be handled properly, then handle it properly, not with a panic

PR #705, opened 2020-04-27
[NOD-913] Use sync rate in getBlockTemplate

someone235:


blockdag/sync_rate.go
```diff
@@ -0,0 +1,52 @@
+package blockdag
+
+import "time"
+
+const syncRateWindowDuration = 15 * time.Minute
+
+// addBlockProcessTime adds the last the last
```
svarogg, line 7: typo the last the last.

Also I'm not sure why break the comment into such short lines.

blockdag/sync_rate.go
```diff
@@ -0,0 +1,52 @@
...
+// block process time in order to measure the
+// recent sync rate.
+//
+// This function MUST be called with the DAG state lock held (for writes).
+func (dag *BlockDAG) addBlockProcessTime() {
```
svarogg, line 12: I'm thinking how to prevent confusion between "the time the block was processed" vs "the time it took the block to process".

Not sure I have a very satisfying solution for that. Some thoughts: 
1. Use "addBlockProcessDate". This will work on clarifying the situation, but a bit silly because we are talking about 15 minute resolutions.
2. Maybe addBlockProcessTimestamp?
3. Maybe something like logBlockProcessed (but without the word "log" to prevent confusion with actual logging - think about some synonym)

Think about it and discuss with team

blockdag/process.go
```diff
@@ -253,6 +253,8 @@ func (dag *BlockDAG) processBlockNoLock(block *util.Block, flags BehaviorFlags)
 		}
 	}
 
+	dag.addBlockProcessTime()
```
svarogg, line 256: I think we want to call that only if block was processed from sync, not submitBlock. No?

blockdag/sync_rate.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// IsSyncRateBelowThreshold checks whether the sync rate
+// is below an expected threshold.
+func (dag *BlockDAG) IsSyncRateBelowThreshold() bool {
+	if dag.uptime() < syncRateWindowDuration {
+		return false
+	}
+
+	const maxDeviation = 0.05
```
svarogg, line 46: I'd rather if the threshold was determined inside isSyncedForMining, and passed as a parameter.
This way all relevant constants would be in the same area.

server/rpc/handle_get_block_template.go
```diff
@@ -131,6 +129,16 @@ func handleGetBlockTemplate(s *Server, cmd interface{}, closeChan <-chan struct{
 	}
 }
 
+func isSyncedForMining(s *Server) bool {
+	const maxDiff = 5 * time.Minute
```
svarogg, line 133: I'd rename to maxTipAge or something

server/rpc/handle_get_block_template.go
```diff
@@ -131,6 +129,17 @@ func handleGetBlockTemplate(s *Server, cmd interface{}, closeChan <-chan struct{
...
 }
 
+func isSyncedForMining(s *Server) bool {
+	const maxTipAge = 5 * time.Minute
+	isCloseToCurrentTime := s.cfg.DAG.Now().Sub(s.cfg.DAG.SelectedTipHeader().Timestamp) <= maxTipAge
+	if isCloseToCurrentTime {
+		return true
+	}
+
+	const maxDeviation = 1.05
```
svarogg, line 139: I'd rename to maxSyncRateDeviation

blockdag/sync_rate.go
```diff
@@ -0,0 +1,49 @@
...
+// blocks in the last syncRateWindowDuration
+// duration.
+func (dag *BlockDAG) syncRate() float64 {
+	dag.RLock()
+	defer dag.RUnlock()
+	return float64(len(dag.recentBlockProcessTimesRelevantWindow())) / syncRateWindowDuration.Seconds()
+}
+
+// IsSyncRateBelowThreshold checks whether the sync rate
+// is below an expected threshold.
```
stasatdaglabs, line 38: "an" -> "the"
Also, please add an explanation about what this threshold is and how it was chosen.

server/rpc/handle_get_block_template.go
```diff
@@ -131,6 +129,17 @@ func handleGetBlockTemplate(s *Server, cmd interface{}, closeChan <-chan struct{
 	}
 }
 
+func isSyncedForMining(s *Server) bool {
```
stasatdaglabs, line 132: Please add an explanation of how the node chooses whether it's synced for mining.

server/rpc/handle_get_block_template.go
```diff
@@ -131,6 +129,17 @@ func handleGetBlockTemplate(s *Server, cmd interface{}, closeChan <-chan struct{
...
+	const maxTipAge = 5 * time.Minute
```
stasatdaglabs, line 133: Please add a comment here or above the function explaining this five minute choice.

server/rpc/handle_get_block_template.go
```diff
@@ -131,6 +129,17 @@ func handleGetBlockTemplate(s *Server, cmd interface{}, closeChan <-chan struct{
...
+	isCloseToCurrentTime := s.cfg.DAG.Now().Sub(s.cfg.DAG.SelectedTipHeader().Timestamp) <= maxTipAge
+	if isCloseToCurrentTime {
+		return true
+	}
+
+	const maxSyncRateDeviation = 1.05
```
stasatdaglabs, line 139: Please add a comment here or above the function explaining this 1.05 choice.

PR #708, opened 2020-04-28
[NOD-943] Add acceptedBlockHashes to GetBlockVerboseResult

someone235:


server/rpc/common.go
```diff
@@ -263,6 +271,7 @@ func buildGetBlockVerboseResult(s *Server, block *util.Block, isVerboseTx bool)
 		Bits:                 strconv.FormatInt(int64(blockHeader.Bits), 16),
 		Difficulty:           getDifficultyRatio(blockHeader.Bits, params),
 		ChildHashes:          childHashStrings,
+		AcceptedBlockHashes:  acceptedBlockHashesStrings,
```
svarogg, line 274: I think it's appropriate to call `daghash.Strings(acceptedBlockHashes)` here. Similar things are being done all-around, and it will save us the extra line and extra variable.

server/rpc/common.go
```diff
@@ -245,6 +244,12 @@ func buildGetBlockVerboseResult(s *Server, block *util.Block, isVerboseTx bool)
 		return nil, internalRPCError(err.Error(), context)
 	}
 
+	acceptedBlockHashes, err := s.cfg.DAG.BluesByBlockHash(hash)
+	if err != nil {
+		context := "Could not get block accepted blocks"
```
stasatdaglabs, line 249: Please rephrase: "Could not get accepted block hashes for block %s"

PR #711, opened 2020-05-04
[NOD-975] Don't include block transactions inside its UTXO commitment

stasatdaglabs:


blockdag/dag.go
```diff
@@ -605,20 +605,13 @@ func (dag *BlockDAG) connectBlock(node *blockNode,
 	return chainUpdates, nil
 }
 
-// calcMultiset returns the multiset of the UTXO of the given block with the given transactions.
-func (node *blockNode) calcMultiset(dag *BlockDAG, transactions []*util.Tx, acceptanceData MultiBlockTxsAcceptanceData, selectedParentUTXO, pastUTXO UTXOSet) (*secp256k1.MultiSet, error) {
+// calcMultiset returns the multiset of the UTXO of the given block.
+func (node *blockNode) calcMultiset(dag *BlockDAG, acceptanceData MultiBlockTxsAcceptanceData, selectedParentUTXO UTXOSet) (*secp256k1.MultiSet, error) {
```
svarogg, line 609: It seems like calcMultiset becomes redundant if it just calls pastUTXOMultiSet.

stasatdaglabs: Maybe so, but I'd still rather have this function. How a multiset is being calculated is implementation details within calcMultiset.

svarogg: Alright, but in that case I guess it can simply be be 
```
func calcMultiset(lalalala){
    return node.pastUTXOMultiSet(dag, acceptanceData, selectedParentUTXO)`
}
```

blockdag/dag.go
```diff
@@ -688,6 +639,19 @@ func (node *blockNode) pastUTXOMultiSet(dag *BlockDAG, acceptanceData MultiBlock
 	return ms, nil
 }
 
+func (node *blockNode) selectedParentMultiset(dag *BlockDAG) (*secp256k1.MultiSet, error) {
```
svarogg, line 642: From a purist perspective, this is not selectedParentMultiset - in case of genesis it returns an empty multiset, since genesis doesn't have a selectedParent.

For any practical reason this is fine, but I think a comment on top of the function is in order. (And while you are at it, give a complete description of what the function does)

blockdag/dag_test.go
```diff
@@ -1246,3 +1247,65 @@ func TestDoubleSpends(t *testing.T) {
...
+		t.Fatalf("TestUTXOCommitment: Failed to setup dag instance: %v", err)
+	}
+	defer teardownFunc()
+
+	// Build the following DAG:
+	// G <- A <- B <- D
+	//        <- C <-
+	genesis := params.GenesisBlock
+	blockA := PrepareAndProcessBlockForTest(t, dag, []*daghash.Hash{genesis.BlockHash()}, nil)
+	blockB := PrepareAndProcessBlockForTest(t, dag, []*daghash.Hash{blockA.BlockHash()}, nil)
```
svarogg, line 1267: Please add some transactions inside the blocks that also remove UTXOs.

blockdag/dag.go
```diff
@@ -605,93 +605,52 @@ func (dag *BlockDAG) connectBlock(node *blockNode,
...
-
 	return ms, nil
 }
 
-func (node *blockNode) pastUTXOMultiSet(dag *BlockDAG, acceptanceData MultiBlockTxsAcceptanceData, selectedParentUTXO UTXOSet) (*secp256k1.MultiSet, error) {
-	ms, err := node.acceptedSelectedParentMultiset(dag, acceptanceData)
+// selectedParentMultiset returns the multiset of the node's selected
+// parent. If the node is the genesis blockNode then it does not have
+// a selected parent, in which case return a new, empty multiset.
+func (node *blockNode) selectedParentMultiset(dag *BlockDAG) (*secp256k1.MultiSet, error) {
```
someone235, line 644: I think you should just inline this function inside pastUTXOMultiset (remember that for genesis pastUTXOMultiset should be empty as well)

blockdag/dag.go
```diff
@@ -1109,7 +1068,12 @@ func (node *blockNode) verifyAndBuildUTXO(dag *BlockDAG, transactions []*util.Tx
 		return nil, nil, nil, nil, err
 	}
 
-	multiset, err = node.calcMultiset(dag, transactions, txsAcceptanceData, selectedParentUTXO, pastUTXO)
+	selectedParentPastUTXO, err := node.selectedParentPastUTXO(dag)
```
someone235, line 1071: It makes more sense to just save UTXO diffs as past UTXO in the first place than calling pastUTXO twice.

blockdag/dag.go
```diff
@@ -1153,6 +1125,17 @@ func (data MultiBlockTxsAcceptanceData) FindAcceptanceData(blockHash *daghash.Ha
 	return nil, false
 }
 
+func (data MultiBlockTxsAcceptanceData) String() string {
+	str := ""
+	for _, kaka := range data {
```
someone235, line 1130: Please remove kaka references

blockdag/dag_test.go
someone235: And add a double spend with C

blockdag/dag.go
```diff
@@ -605,93 +605,52 @@ func (dag *BlockDAG) connectBlock(node *blockNode,
...
+			tx := txAcceptanceData.Tx.MsgTx()
 
 			var err error
-			ms, err = removeUTXOFromMultiset(ms, unacceptedEntry, &outpoint)
-			if err != nil {
-				return nil, err
-			}
-
-			ms, err = addUTXOToMultiset(ms, acceptedEntry, &outpoint)
+			ms, err = addTxToMultiset(ms, tx, selectedParentPastUTXO, node.blueScore)
```
svarogg, line 632: Please make sure (preferably with a unit-test) that addTxToMultiset doesn't change `node.selectedParentMultiset` in-place in such a way that it affects the MultiSet store.
TBH, I think it does (but maybe there's a copy somewhere in secp256k)

blockdag/dag.go
svarogg, line 1130: kaka baka look like temporary names

blockdag/dag.go
stasatdaglabs: We spoke about this and decided to leave it as is.

blockdag/dag.go
stasatdaglabs: Created a ticket for it.

PR #716, opened 2020-05-12
[NOD-820] When the node isn't synced, make getBlockTemplate return a boolean isSynced instead of an error

stasatdaglabs:


cmd/kaspaminer/mineloop.go
```diff
@@ -207,12 +209,23 @@ func getBlockTemplate(client *minerClient, longPollID string) (*rpcmodel.GetBloc
...
+
 	var stopOldTemplateSolving chan struct{}
 	for template := range newTemplateChan {
 		if stopOldTemplateSolving != nil {
 			close(stopOldTemplateSolving)
 		}
+
+		if !template.IsSynced {
+			if !mineWhenNotSynced {
+				errChan <- errors.Errorf("Got template with isSynced=false")
```
someone235, line 223: Errors shouldn't start with a capital letter

PR #717, opened 2020-05-12
[NOD-982] Log message with level WARN when getting MsgReject

stasatdaglabs:


peer/log.go
```diff
@@ -171,3 +172,12 @@ func messageSummary(msg wire.Message) string {
 	// No summary for other messages.
 	return ""
 }
+
+func messageLogLevel(msg wire.Message) logs.Level {
```
svarogg, line 176: I think this should be in peer.go, right near readMessage.
log.go is infrastructure for logging.
This is pure business-logic.

PR #719, opened 2020-05-13
[NOD-993] Get rid of AssertError + Use %+v when printing startup errors

svarogg:


kaspad.go
```diff
@@ -138,7 +139,7 @@ func kaspadMain(serverChan chan<- *server.Server) error {
 		interrupt)
 	if err != nil {
 		// TODO: this logging could do with some beautifying.
```
stasatdaglabs, line 141: Please don't leave TODO in the code.

blockdag/error.go
```diff
@@ -295,6 +287,6 @@ func (e RuleError) Error() string {
 }
 
 // ruleError creates an RuleError given a set of arguments.
```
stasatdaglabs, line 289: "given a set of arguments" is redundant.

stasatdaglabs: Also, "an" -> "a"

stasatdaglabs: Actually, why does this comment even exist? Please remove it.

svarogg:

PR #720, opened 2020-05-14
[NOD-994] Greatly increase the amount of logs kaspad keeps before rotating them away

svarogg:


logs/logs.go
```diff
@@ -265,15 +266,27 @@ func callsite(flag uint32) (string, int) {
 	return file, line
 }
 
+const (
+	defaultThreasholdKB = 10 * 1024
```
stasatdaglabs, line 270: Typo: Threashold -> Threshold

PR #722, opened 2020-05-18
[NOD-1006] Make use of a pool to avoid excessive allocation of big.Ints 

stasatdaglabs:


mining/mining.go
```diff
@@ -206,10 +207,12 @@ func (g *BlkTmplGenerator) NewBlockTemplate(payToAddress util.Address, extraNonc
 		return nil, err
 	}
 
+	targetDifficulty := big.NewInt(0)
+	util.CompactToBig(msgBlock.Header.Bits, targetDifficulty)
 	log.Debugf("Created new block template (%d transactions, %d in fees, "+
 		"%d mass, target difficulty %064x)",
 		len(msgBlock.Transactions), txsForBlockTemplate.totalFees,
-		txsForBlockTemplate.totalMass, util.CompactToBig(msgBlock.Header.Bits))
+		txsForBlockTemplate.totalMass, targetDifficulty)
```
svarogg, line 215: It might be a good idea to use LogClosure here, so that targetDifficulty will not be calculated if log level is above debug

util/math.go
```diff
@@ -52,7 +52,7 @@ func FastLog2Floor(n uint64) uint8 {
 //
 // The formula to calculate N is:
 // 	N = (-1^sign) * mantissa * 256^(exponent-3)
-func CompactToBig(compact uint32) *big.Int {
+func CompactToBig(compact uint32, out *big.Int) {
```
someone235, line 55: I would call it "dst" or "destination", no?

server/rpc/handle_get_block_template.go
```diff
@@ -784,7 +787,9 @@ func (state *gbtWorkState) blockTemplateResult(dag *blockdag.BlockDAG, useCoinba
 	// implied by the included or omission of fields:
 	//  Including MinTime -> time/decrement
 	//  Omitting CoinbaseTxn -> coinbase, generation
-	targetDifficulty := fmt.Sprintf("%064x", util.CompactToBig(header.Bits))
+	target := big.NewInt(0)
+	util.CompactToBig(header.Bits, target)
```
svarogg, line 791: Maybe create function that wraps CompactToBig that returns the result, in addition to the one with the out param.

You optimized one place, but all other places have become a bit more verbose because of it. Better be able to enjoy both worlds.

util/math.go
```diff
@@ -63,21 +63,18 @@ func CompactToBig(compact uint32) *big.Int {
 	// treat the exponent as the number of bytes and shift the mantissa
 	// right or left accordingly. This is equivalent to:
 	// N = mantissa * 256^(exponent-3)
-	var bn *big.Int
 	if exponent <= 3 {
 		mantissa >>= 8 * (3 - exponent)
-		bn = big.NewInt(int64(mantissa))
+		out = out.SetInt64(int64(mantissa))
```
someone235, line 68: Using assignment for mutable function is confusing

blockdag/blockwindow.go
```diff
@@ -53,13 +54,37 @@ func (window blockWindow) minMaxTimestamps() (min, max int64) {
...
+	target := acquireBigInt()
 	for _, node := range window {
-		target := util.CompactToBig(node.bits)
+		util.CompactToBigWithDestination(node.bits, target)
 		averageTarget.Add(averageTarget, target)
 	}
-	return averageTarget.Div(averageTarget, big.NewInt(int64(len(window))))
+
+	// Reuse `target` to avoid a big.Int allocation
+	windowLen := target
```
svarogg, line 82: Better utilize the pool rather then re-use directly.
Also, please defer the release of all BigInts

blockdag/blockwindow.go
```diff
@@ -53,13 +54,37 @@ func (window blockWindow) minMaxTimestamps() (min, max int64) {
...
+
+func acquireBigInt() *big.Int {
+	return blockWindowBigIntPool.Get().(*big.Int)
+}
+
+func releaseBigInt(toRelease *big.Int) {
+	toRelease.SetInt64(0)
+	blockWindowBigIntPool.Put(toRelease)
+}
+
```
someone235, line 71: If it's a general big int pool, call it with a generic name, if it's not, call the function acquireBigIntFromBlockWindowBigIntPool or something like this

util/bigintpool/pool.go
```diff
@@ -0,0 +1,26 @@
...
+// initializes it to x.
+func Acquire(x int64) *big.Int {
+	bigInt := bigIntPool.Get().(*big.Int)
+	bigInt.SetInt64(x)
+	return bigInt
+}
+
+// Release returns the given big.Int to the pool.
+func Release(toRelease *big.Int) {
+	toRelease.SetInt64(0)
```
someone235, line 24: Why do you need this line?

PR #724, opened 2020-05-18
[NOD-990] Save utxo diffs of past UTXO

someone235:


blockdag/dag.go
```diff
@@ -1153,31 +1131,35 @@ func (node *blockNode) fetchBlueBlocks() ([]*util.Block, error) {
...
-	pastUTXO = acceptedSelectedParentUTXO
-	multiBlockTxsAcceptanceData = MultiBlockTxsAcceptanceData{BlockTxsAcceptanceData{
-		BlockHash:        *node.selectedParent.hash,
-		TxAcceptanceData: selectedParentAcceptanceData,
-	}}
+	selectedParentPastDiffUTXOSet, ok := selectedParentPastUTXO.(*DiffUTXOSet)
+	if !ok {
+		panic("selectedParentPastUTXO is not *DiffUTXOSet")
+	}
+	pastUTXO = NewDiffUTXOSet(selectedParentPastDiffUTXOSet.base, selectedParentPastDiffUTXOSet.UTXODiff.clone())
```
svarogg, line 1141: Why not use utxoSet.clone() for this whole thing?
No errors, no panics, no type assertions.

svarogg: Note: right now DiffUTXOSet.clone actually clones the base as well.
I'm not sure why is it useful, so I suggest changing this behavior. (Ofc, make sure it makes sense in all usages)

blockdag/error.go
```diff
@@ -121,6 +121,11 @@ const (
 	// either does not exist or has already been spent.
 	ErrMissingTxOut
 
+	// ErrDoubleSpendsWithBlockTransaction indicates a transaction
+	// that spends an output that was already spent by another
+	// transaction in the same block.
+	ErrDoubleSpendsWithBlockTransaction
```
svarogg, line 127: I don't like this name too much, is a bit confusing.
How about ErrDoubleSpendWithinBlock? Or ErrDoubleSpendInSameBlock or something else that makes it clear what this is?

blockdag/validate.go
```diff
@@ -545,6 +545,22 @@ func (dag *BlockDAG) checkBlockSanity(block *util.Block, flags BehaviorFlags) (t
 		existingTxIDs[*id] = struct{}{}
 	}
 
+	// Check for duplicate transactions. This check will be fairly quick
+	// since the transaction IDs are already cached due to building the
+	// merkle tree above.
+	usedOutpoints := make(map[wire.Outpoint]struct{})
```
svarogg, line 551: Consider the map being from outpoint to tx that spends it.
This way you can print in the error message which is the doublespending transaction - might save a tons of debugging to somebody in the future

blockdag/validate.go
```diff
@@ -838,6 +854,11 @@ func (dag *BlockDAG) checkConnectToPastUTXO(block *blockNode, pastUTXO UTXOSet,
 			return nil, err
 		}
 
+		err = checkDoubleSpendsWithBlockPast(pastUTXO, transactions)
```
svarogg, line 857: Is there any chance checkDoubleSpendsWithBlockPast makes ensureNoDuplicateTx redundant?

someone235: ensureNoDuplicateTx checks another thing (that you do not have a transaction that has the same ID as another tx in the UTXO, maybe because of hash collision), but maybe you talked about the duplicate transaction check in checkBlockSanity. But if we'll remove it we'll not get errors for hash collisions in the same block.

svarogg: Hmmmm... actually, it seems like this is implicitly validated at least two more times: In CheckBlockMass and when validating the signatures, since both require all relevant UTXOs to exist in the pastUTXO.
In that case, do we really want to validate this explicitly?

PR #731, opened 2020-05-25
[NOD-1012] Disable subnetworks

stasatdaglabs:


dagconfig/params.go
```diff
@@ -380,6 +392,9 @@ var SimnetParams = Params{
 	PrivateKeyID: 0x64, // starts with 4 (uncompressed) or F (compressed)
 	// Human-readable part for Bech32 encoded addresses
 	Prefix: util.Bech32PrefixKaspaSim,
+
+	// EnableNonNativeSubnetworks enables non-native/coinbase transactions
+	EnableNonNativeSubnetworks: true,
```
svarogg, line 397: I'm not sure it's a good idea having simnet EnableNonNativeSubnetworks.

I'd rather if this is false by default in all networks, and changed "manually" only in the tests that care about this.

PR #734, opened 2020-05-26
[NOD-1013] Drain channels forever without blocking when doing graceful shutdown

someone235:


svarogg:
Closed because decided to do something else to fix the issue.

PR #736, opened 2020-05-27
[NOD-1013] Don't wait for syncManager and rebroadcastHandler when doing graceful shutdown

someone235:


svarogg:
Closed because decided to do something else to fix the issue.

PR #741, opened 2020-06-02
[NOD-1004] Make AddrManager.getAddress use only 1 loop to check all address chances and pick one of them

stasatdaglabs:


addrmgr/addrmanager.go
```diff
@@ -226,49 +226,26 @@ func (a *AddrManager) updateAddress(netAddr, srcAddr *wire.NetAddress, subnetwor
 		netAddrCopy := *netAddr
 		ka = &KnownAddress{na: &netAddrCopy, srcAddr: srcAddr, subnetworkID: subnetworkID}
 		a.addrIndex[addr] = ka
-		if subnetworkID == nil {
-			a.nNewFullNodes++
-		} else {
-			a.nNew[*subnetworkID]++
-		}
+		a.incrementNNewNodes(subnetworkID)
 		// XXX time penalty?
```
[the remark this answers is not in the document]
stasatdaglabs: Looks like it had been added during a "move a file elsewhere" commit, so there aren't any clues there. I think it's safe to say that the original intention had been lost to time. I'm removing it.

addrmgr/addrmanager.go
```diff
@@ -288,34 +265,35 @@ func (a *AddrManager) updateAddrNew(bucket int, addr string, ka *KnownAddress) {
...
 
 	if _, ok := a.addrTried[*ka.subnetworkID]; !ok {
 		a.addrTried[*ka.subnetworkID] = &triedBucket{}
 		for i := range a.addrTried[*ka.subnetworkID] {
-			a.addrTried[*ka.subnetworkID][i] = list.New()
+			a.addrTried[*ka.subnetworkID][i] = nil
 		}
 	}
-	a.addrTried[*ka.subnetworkID][bucket].PushBack(ka)
+	a.addrTried[*ka.subnetworkID][bucket] = append(a.addrTried[*ka.subnetworkID][bucket], ka)
```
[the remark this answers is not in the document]
stasatdaglabs: The sizes of the arrays are very limited (64 item IIRC), and I only ever append to the end of the slice, which is cheap. There really is no reason to use a linked list here.

addrmgr/addrmanager.go
```diff
@@ -919,79 +879,93 @@ func (a *AddrManager) GetAddress() *KnownAddress {
...
-			}
-			factor *= 1.2
+	}
+
+	panic("randomValue is exactly 1, which cannot happen")
+}
+
+type bucket interface {
+	name() string
+	randomBucket(random *rand.Rand) []*KnownAddress
```
someone235, line 931: It's quite weird that randomBucket doesn't return a bucket

addrmgr/addrmanager.go
```diff
@@ -1070,13 +1044,17 @@ func (a *AddrManager) Good(addr *wire.NetAddress, subnetworkID *subnetworkid.Sub
...
-				break
+		bucket := a.addrTried[*oldSubnetworkID][triedBucketIndex]
+		toRemoveIndex := -1
+		for i, ka := range bucket {
+			if NetAddressKey(ka.NetAddress()) == addrKey {
+				toRemoveIndex = i
 			}
 		}
+		if toRemoveIndex != -1 {
+			a.addrTried[*oldSubnetworkID][triedBucketIndex] =
```
someone235, line 1055: You can just move this to 1050 and break, and then get rid of weird -1

addrmgr/addrmanager.go
stasatdaglabs: We spoke about this and decided to leave it as is.

addrmgr/addrmanager.go
stasatdaglabs: I cannot, since that would be modifying the bucket while iterating over it.

addrmgr/addrmanager.go
```diff
@@ -331,47 +312,28 @@ func (a *AddrManager) expireNew(bucket *newBucket, idx int, decrNewCounter func(
...
-	var oldestElem *list.Element
-	var lst *list.List
-	if subnetworkID == nil {
-		lst = a.addrTriedFullNodes[bucket]
-	} else {
-		lst = a.addrTried[*subnetworkID][bucket]
-	}
-	for e := lst.Front(); e != nil; e = e.Next() {
-		ka := e.Value.(*KnownAddress)
+	oldestIndex := -1
```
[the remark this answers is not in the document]
stasatdaglabs: I return the index from the function, so a boolean won't save me from that.

addressmanager/addressmanager.go
```diff
@@ -0,0 +1,1363 @@
...
+	accumulatedChance := float64(0)
+	for _, knownAddress := range randomBucket {
+		normalizedChance := knownAddress.chance() / totalChance
+		accumulatedChance += normalizedChance
+		if randomValue < accumulatedChance {
+			return knownAddress
+		}
+	}
+
+	panic("randomValue is exactly 1, which cannot happen")
```
svarogg, line 960: At the same time it could be also > 1.
I'd rephrase the error to "randomValue >= 1, which cannot happen".

addrmgr/addrmanager.go
someone235: I would still prefer that you'll use a boolean

stasatdaglabs:

PR #747, opened 2020-06-07
[NOD-970] Add isSyncing flag

someone235:


netsync/manager.go
```diff
@@ -132,13 +132,13 @@ type requestQueueAndSet struct {
...
-	syncCandidate           bool
-	lastSelectedTipRequest  time.Time
-	isPendingForSelectedTip bool
-	requestQueueMtx         sync.Mutex
-	requestQueues           map[wire.InvType]*requestQueueAndSet
-	requestedTxns           map[daghash.TxID]struct{}
-	requestedBlocks         map[daghash.Hash]struct{}
+	syncCandidate          bool
+	lastSelectedTipRequest time.Time
+	shouldSendSelectedTip  bool
```
svarogg, line 137: shouldSendSelectedTip is very confusing name. I reviewed this whole PR in a wrong way because of it, before realizing what it actually means :|

It's not clear whether we should send or the peer should send.

What was bad with `isPendingForSelectedTip`?

PR #751, opened 2020-06-08
[NOD-1033] Instead of checking number of transactions, check that the block size in bytes does not exceed MaxMassPerBlock

stasatdaglabs:


blockdag/validate.go
```diff
@@ -473,17 +473,16 @@ func (dag *BlockDAG) checkBlockSanity(block *util.Block, flags BehaviorFlags) (t
...
 		return 0, ruleError(ErrNoTransactions, "block does not contain "+
 			"any transactions")
 	}
 
-	// A block must not have more transactions than the max block mass or
-	// else it is certainly over the block mass limit.
-	if numTx > wire.MaxMassPerBlock {
-		str := fmt.Sprintf("block contains too many transactions - "+
-			"got %d, max %d", numTx, wire.MaxMassPerBlock)
+	// A block must not bigger in bytes than the max block mass because
```
svarogg, line 481: typo: ... must not **be** bigger

blockdag/validate.go
```diff
@@ -473,17 +473,16 @@ func (dag *BlockDAG) checkBlockSanity(block *util.Block, flags BehaviorFlags) (t
...
+	// otherwise it is certainly over the block mass limit.
+	if msgBlock.SerializeSize() > wire.MaxMassPerBlock {
```
svarogg, line 483: Please call .SerializeSize() only once.

svarogg:
Please also add test that checks that validation fails if block size is 1 byte too big

blockdag/validate_test.go
```diff
@@ -194,6 +195,30 @@ func TestCheckBlockSanity(t *testing.T) {
 		t.Errorf("CheckBlockSanity: unexpected return %s delay", delay)
 	}
 
+	// Ensure that a block that's bigger in bytes than MaxMassPerBlock fails
+	blockBytes, err := block.Bytes()
+	if err != nil {
+		t.Fatalf("CheckBlockSanity: Bytes unexpectedly failed: %s", err)
+	}
+	tooBigMassBlock, err := util.NewBlockFromBytes(blockBytes)
```
someone235, line 203: Why call it tooBigMass? It's tooBigBytes, no? Or you can just call it tooBigBlock?

PR #756, opened 2020-06-10
[NOD-1048] Make leveldb compaction much less frequent

stasatdaglabs:


blockdag/test_utils.go
```diff
@@ -62,6 +64,11 @@ func DAGSetup(dbName string, openDb bool, config Config) (*BlockDAG, func(), err
 			return nil, nil, errors.Errorf("error creating temp dir: %s", err)
 		}
 
+		originalOptions := ldb.Options
```
svarogg, line 67: Since the scope of this var is a bit wide, I'd rename to originalLDBOptions

PR #762, opened 2020-06-15
[NOD-833] Remove getBlockTemplate capabilites and move mining address to getBlockTemplate

someone235:


someone235:

someone235:

PR #763, opened 2020-06-15
[NOD-1035] Use reachability to check finality

someone235:


someone235:

blockdag/dag.go
```diff
@@ -816,6 +816,24 @@ func (dag *BlockDAG) LastFinalityPointHash() *daghash.Hash {
 	return dag.lastFinalityPoint.hash
 }
 
+// isInSelectedParentChain returns whether a is in the selected parent chain of b
+func (dag *BlockDAG) isInSelectedParentChain(a, b *blockNode) (bool, error) {
```
stasatdaglabs, line 820: Please rename to something like aNode, bNode.

blockdag/dag.go
```diff
@@ -824,12 +842,17 @@ func (dag *BlockDAG) checkFinalityRules(newNode *blockNode) error {
...
+		return nil
+	}
+
+	isInSelectedChain, err := dag.isInSelectedParentChain(dag.lastFinalityPoint, newNode.selectedParent)
+	if err != nil {
+		return err
+	}
+
+	if !isInSelectedChain {
+		return ruleError(ErrFinality, "the last finality point is not in the selected chain of this block")
```
stasatdaglabs, line 855: Please add the word "parent" between the words "selected" and "chain".

blockdag/dag.go
```diff
@@ -816,20 +816,51 @@ func (dag *BlockDAG) LastFinalityPointHash() *daghash.Hash {
 	return dag.lastFinalityPoint.hash
 }
 
+// isInSelectedParentChain returns whether aNode is in the selected parent chain of bNode.
+func (dag *BlockDAG) isInSelectedParentChain(aNode, bNode *blockNode) (bool, error) {
+	// Because reachabilityInterval.isAncestorOf and isInSelectedParentChain
```
svarogg, line 821: If interval.isAncestorOf is exclusive (a.k.a returns false if aNode==b.Node, if I understand correctly what you mean), why do we need to check this explicitly?

And if it's not exclusive in the sense I defined here, didn't we agree we should make it such?

blockdag/dag.go
```diff
@@ -816,20 +816,51 @@ func (dag *BlockDAG) LastFinalityPointHash() *daghash.Hash {
...
 
-	for currentNode := newNode; currentNode != dag.lastFinalityPoint; currentNode = currentNode.selectedParent {
-		// If we went past dag's last finality point without encountering it -
-		// the new block has violated finality.
-		if currentNode.blueScore <= dag.lastFinalityPoint.blueScore {
-			return ruleError(ErrFinality, "The last finality point is not in the selected chain of this block")
-		}
+	// Because newNode doesn't have reachability data we
+	// need to check if the last finality point is in the
+	// selected parent chainof newNode.selectedParent, so
```
svarogg, line 850: typo: chainof

blockdag/dag.go
```diff
@@ -816,20 +816,51 @@ func (dag *BlockDAG) LastFinalityPointHash() *daghash.Hash {
...
+		return false, err
+	}
+
+	return aTreeNode.interval.isAncestorOf(bTreeNode.interval), nil
+}
+
 // checkFinalityRules checks the new block does not violate the finality rules
-// specifically - the new block selectedParent chain should contain the old finality point
+// specifically - the new block selectedParent chain should contain the old finality point.
 func (dag *BlockDAG) checkFinalityRules(newNode *blockNode) error {
```
svarogg, line 842: I'd rename to `checkFinalityPointInSelectedParentChain` to be more explicit regarding what we check (since that's the only finality rule anyway)

svarogg: Or something like `checkFinalityViolation` if you want something shorter, but still don't want to imply there's more then one finality rule.

blockdag/reachability.go
```diff
@@ -155,7 +155,7 @@ func exponentialFractions(sizes []uint64) []float64 {
 // property of reachability intervals that intervals are either completely disjoint,
 // or one strictly contains the other.
 func (ri *reachabilityInterval) isAncestorOf(other *reachabilityInterval) bool {
-	return ri.start <= other.end && other.end <= ri.end
+	return ri.start <= other.end && other.end < ri.end
```
svarogg, line 158: Pretty sure that if `ri.start < other.end && other.end <= ri.end` we should also return true.

In addition to fix, please make sure we have test cases for all:
1. ri == other
2. ri.start == other.start && ri.end != other.end
3. ri.end != other.end && ri.end == other.end
4. ri.start > other.start && ri.end < other.end
5. ri.start < other.start && ri.end > other.end

PR #766, opened 2020-06-16
[NOD-1059] Always call sm.restartSyncIfNeeded() when getting selectedtip message

someone235:


svarogg:

PR #768, opened 2020-06-17
[NOD-1060] Don't sync from misbehaving peer

someone235:


svarogg:

PR #773, opened 2020-06-22
[NOD-1063] Optimize deep reachability tree insertions 

stasatdaglabs:


blockdag/reachability.go
```diff
@@ -366,14 +416,199 @@ func (rtn *reachabilityTreeNode) propagateInterval(subTreeSizeMap map[*reachabil
...
+}
+
+func (rtn *reachabilityTreeNode) reclaimIntervalBeforeChosenChild(
+	commonAncestor *reachabilityTreeNode, commonAncestorChosenChild *reachabilityTreeNode, reindexRoot *reachabilityTreeNode) (
+	map[*reachabilityTreeNode]struct{}, error) {
+
+	modifiedTreeNodes := make(map[*reachabilityTreeNode]struct{})
+
+	current := commonAncestorChosenChild
+	for !current.hasSlackIntervalBefore() {
```
svarogg, line 453: Please add comment explaining the two loops

blockdag/reachability.go
```diff
@@ -366,14 +416,199 @@ func (rtn *reachabilityTreeNode) propagateInterval(subTreeSizeMap map[*reachabil
...
+		if current == reindexRoot {
```
svarogg, line 454: The code here is not linear. Anything happening inside this if clause actually happens in the end of the loop.

Therefore I suggest to either move this whole thing into a function, or break out of here and do this outside the loop

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
 	}
+
+	// Update the reindex root.
+	// Note that we check for blue score here in order to find out
+	// whether the new node is going to be the virtual's selected
+	// parent. We don't check node == virtual.selectedParent because
+	// at this stage the virtual had not yet been updated.
+	if node.blueScore > dag.SelectedTipBlueScore() {
+		updateStartTime := time.Now()
+		modifiedTreeNodes, err := dag.updateReachabilityReindexRoot(newTreeNode)
```
svarogg, line 822: I'd rather if we check for whether we should or not update reachability root at this point

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+	if err != nil {
+		return nil, nil, false, err
+	}
+
+	return chosenReindexRootChild, modifiedTreeNodes, true, nil
+}
+
+// findReachabilityTreeAncestorInChildren finds the reachability tree child
+// of rtn that is the ancestor of node.
+func (rtn *reachabilityTreeNode) findReachabilityTreeAncestorInChildren(node *reachabilityTreeNode) (*reachabilityTreeNode, error) {
```
michaelsutton, line 891: The whole point is to make the search in children logarithmic, so if this line copies the entire list into a futureCoveringSet it becomes linear anyway, so might as well just search linearly in the original children list... instead I suggest extracting findIndex logic into a global function which can be used here as well

blockdag/reachability.go
```diff
@@ -227,23 +283,17 @@ func (rtn *reachabilityTreeNode) addChild(child *reachabilityTreeNode) ([]*reach
...
 		return nil, err
 	}
 	child.setInterval(allocated)
-	rtn.remainingInterval = remaining
-	return []*reachabilityTreeNode{rtn, child}, nil
+	return map[*reachabilityTreeNode]struct{}{rtn: {}, child: {}}, nil
 }
 
 // setInterval sets the reachability interval for this node.
 func (rtn *reachabilityTreeNode) setInterval(interval *reachabilityInterval) {
```
michaelsutton, line 295: The code sometimes uses setInterval and sometimes sets interval directly. Since setInterval no longer sets the 'remaining' field, we should either use it all the time or delete and not use it at all

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+	modifiedTreeNodes := make(map[*reachabilityTreeNode]struct{})
+
+	reindexRootChildNodesBeforeChosen, reindexRootChildNodesAfterChosen, err :=
+		reindexRoot.splitChildrenAroundChosenChild(chosenReindexRootChild)
+	if err != nil {
+		return nil, err
+	}
+
+	reindexRootChildNodesBeforeChosenSizesSum, modifiedNodes, err :=
+		dag.tightenReachabilityTreeIntervalsBeforeChosenReindexRootChild(reindexRoot, reindexRootChildNodesBeforeChosen)
```
michaelsutton, line 915: maybe tightenReachabilityTreeIntervalsBeforeReindexRoot**Chosen**Child

also I think TreeIntervals can be Intervals

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+	if err != nil {
+		return nil, err
+	}
+	for node := range modifiedNodes {
+		modifiedTreeNodes[node] = struct{}{}
+	}
+
+	reindexRootChildNodesAfterChosenSizesSum, modifiedNodes, err :=
+		dag.tightenReachabilityTreeIntervalsAfterChosenReindexRootChild(reindexRoot, reindexRootChildNodesAfterChosen)
```
michaelsutton, line 924: maybe tightenReachabilityTreeIntervalsAfterReindexRoot**Chosen**Child

also I think TreeIntervals can be Intervals

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+
+	reindexRootStart := reindexRoot.interval.start
+	reindexRootEnd := reindexRoot.interval.end
+	targetRangeForReindexRootChildStart :=
+		reindexRootStart + reindexRootChildNodesBeforeChosenSizesSum + reachabilityReindexSlack
+	targetRangeForReindexRootChildEnd :=
+		reindexRootEnd - reindexRootChildNodesAfterChosenSizesSum - reachabilityReindexSlack - 1
+	newReindexRootChildInterval := newReachabilityInterval(targetRangeForReindexRootChildStart, targetRangeForReindexRootChildEnd)
+
+	if targetRangeForReindexRootChildStart > reindexRootStart || targetRangeForReindexRootChildEnd < reindexRootEnd {
```
michaelsutton, line 1014: perhaps general interval logic such as interval contains/doesNotContain/isBefore/isAfter can be extracted to interval methods. For instance:
   `def doesNotContain(interval this, interval other):
       return this.start > other.start || this.end < other.end
`

which will be used here:
`if newReindexRootChildInterval.doesNotContain(reindexRoot.interval)`

blockdag/reachability.go
```diff
@@ -187,33 +198,78 @@ type reachabilityTreeNode struct {
...
 
 func newReachabilityTreeNode(blockNode *blockNode) *reachabilityTreeNode {
 	// Please see the comment above reachabilityTreeNode to understand why
 	// we use these initial values.
 	interval := newReachabilityInterval(1, math.MaxUint64-1)
-	// We subtract 1 from the end of the remaining interval to prevent the node from allocating
+	return &reachabilityTreeNode{blockNode: blockNode, interval: interval}
+}
+
+func (rtn *reachabilityTreeNode) childIntervalAllocationRange() *reachabilityInterval {
```
michaelsutton, line 210: Maybe child**ren**IntervalAllocationRange ?

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+		modifiedTreeNodes[node] = struct{}{}
+	}
+
+	return modifiedTreeNodes, nil
+}
+
+func (rtn *reachabilityTreeNode) splitChildrenAroundChosenChild(chosenChild *reachabilityTreeNode) (
+	nodesBeforeChosen []*reachabilityTreeNode, nodesAfterChosen []*reachabilityTreeNode, err error) {
+
+	chosenIndex := -1
```
michaelsutton, line 947: findIndex logic from futureCoveringSet can be used here as well to find the index of chosen in log time (although this is probably in overkill because children are accessed linearly in the calling method anyway)

blockdag/reachability.go
michaelsutton, line 1014: **CRITICAL ERROR**:  the comparison here should be with `prevReindexRootChildStart` and `prevReindexRootChildEnd`
which should be set `prevReindexRootChildStart := chosenReindexRootChild.interval.start` and `prevReindexRootChildEnd := chosenReindexRootChild.interval.end`

blockdag/reachability.go
```diff
@@ -382,6 +617,23 @@ func (rtn *reachabilityTreeNode) isAncestorOf(other *reachabilityTreeNode) bool
 	return rtn.interval.isAncestorOf(other.interval)
 }
 
+// findCommonAncestor finds the most recent reachability tree ancestor
+// common to both rtn and other.
+func (rtn *reachabilityTreeNode) findCommonAncestor(other *reachabilityTreeNode) *reachabilityTreeNode {
+	currentThis := rtn
+	currentOther := other
+	for {
+		if currentThis.parent == nil || currentThis.isAncestorOf(other) {
```
svarogg, line 626: `currentThis.parent == nil` implies that currentThis is genesis, right?
But in that case `currentThis.isAncestorOf(other)` will also be true, wouldn't it?

In that case I'd remove the `currentThis.parent == nil` condition.

Then when we get to `currentThis = currentThis.parent` and one of the guys has no parents - panic :fearful: 

Why here I'm OK with panicing while in other cases I was against (it's still a shift in my position, but please don't make it a slippery slope) - this is an assumption that the reachability module has - all tree nodes have a common ancestor, unlike most cases where we wanted to assert a more general assumption in a specific library.
I'd still rather return an error in that case, but that will propagate error-returning to a ton of functions, so I'm fine with panicing this one time.

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+			modifiedTreeNodes[node] = struct{}{}
+		}
+		nextReindexRoot = candidateReindexRoot
+	}
+
+	dag.reachabilityReindexRoot = nextReindexRoot
+	return modifiedTreeNodes, nil
+}
+
+func (dag *BlockDAG) tryMovingReachabilityReindexRoot(
```
svarogg, line 865: This is not "try" because it does not imply "do something, if you fail - we're fine". It's a "maybe" at the very least.
But anyway, as mentioned in comment before - if possible and makes sense - better to restructure this whole thing

blockdag/reachability.go
svarogg: I think `ReachabilityTreeIntervals` can be `Intervals`, in all names, not only here.

The thing is, those functions can't be dag functions in that case.
However, from a quick check, it seems like we are never actually using the dag, only dag.reachabilityStore, so maybe we can scope all these functions to reachabilityStore? (We might want to rename it to reachabilityManager or something to be more general)

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+	for node := range modifiedNodes {
+		modifiedTreeNodes[node] = struct{}{}
+	}
+
+	reindexRootChildNodesAfterChosenSizesSum, modifiedNodes, err :=
+		dag.tightenReachabilityTreeIntervalsAfterChosenReindexRootChild(reindexRoot, reindexRootChildNodesAfterChosen)
+	if err != nil {
+		return nil, err
+	}
+	for node := range modifiedNodes {
```
svarogg, line 928: Yet another reason to alias map[*reachabilityTreeNode]struct{} to reachabilityTreeNodeSet is that then you can add a method that will do this loop instead of repeating it a few times over this function

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+	}
+
+	return rootChildrenFutureCoveringSet[i-1].treeNode, nil
+}
+
+func (dag *BlockDAG) concentrateReachabilityTreeIntervalAroundReindexRootChild(
+	reindexRoot *reachabilityTreeNode, chosenReindexRootChild *reachabilityTreeNode) (
+	map[*reachabilityTreeNode]struct{}, error) {
+
+	modifiedTreeNodes := make(map[*reachabilityTreeNode]struct{})
```
svarogg, line 906: It's a bit hard for a casual reader to understand what's the diff between `modifiedTreeNodes` and the two `modifiedNodes`s.
I'd suggest renaming to `allModifiedTreeNodes`, `beforeModifiedTreeNodes` and `afterModifiedTreeNodes` or something

blockdag/reachability.go
svarogg: I agree, but isn't it already done in `isAncestorOf`, so just replace these places with it?

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+	return reindexRootChildNodesAfterChosenSizesSum, modifiedTreeNodes, nil
+}
+
+func (dag *BlockDAG) expandReachabilityTreeIntervalInChosenReindexRootChild(reindexRoot *reachabilityTreeNode,
+	chosenReindexRootChild *reachabilityTreeNode, reindexRootChildNodesBeforeChosenSizesSum uint64,
+	reindexRootChildNodesAfterChosenSizesSum uint64) (map[*reachabilityTreeNode]struct{}, error) {
+
+	modifiedTreeNodes := make(map[*reachabilityTreeNode]struct{})
+
+	reindexRootStart := reindexRoot.interval.start
```
svarogg, line 1006: Here, and in other places, I'd rather inline `reindexRoot.interval.start` instead of assign it to variable.
This way it's clearer where the var comes from.

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+	reindexRoot *reachabilityTreeNode, reindexRootChildNodesBeforeChosen []*reachabilityTreeNode) (
+	reindexRootChildNodesBeforeChosenSizesSum uint64, modifiedTreeNodes map[*reachabilityTreeNode]struct{}, err error) {
+
+	reindexRootChildNodesBeforeChosenSizes, reindexRootChildNodesBeforeChosenSubtreeSizeMaps, reindexRootChildNodesBeforeChosenSizesSum :=
+		calcReachabilityTreeNodeSizes(reindexRootChildNodesBeforeChosen)
+
+	reindexRootStart := reindexRoot.interval.start
+	targetRangeBeforeReindexRootStart := reindexRootStart + reachabilityReindexSlack
+	targetRangeBeforeReindexRootEnd := targetRangeBeforeReindexRootStart + reindexRootChildNodesBeforeChosenSizesSum - 1
+	intervalBeforeReindexRootStart := newReachabilityInterval(targetRangeBeforeReindexRootStart, targetRangeBeforeReindexRootEnd)
```
svarogg, line 970: I think the 3 above lines would be much more readable if you do:
```
intervalBeforeReindexRootStart := newReachabilityInterval(
     reindexRootStart + reachabilityReindexSlack,
     targetRangeBeforeReindexRootStart + reindexRootChildNodesBeforeChosenSizesSum - 1)
```

Same for tightenReachabilityTreeIntervalsAfterChosenReindexRootChild

blockdag/reachability.go
michaelsutton, line 626: the test for nil is redundant because genesis will always return true for `.isAncestorOf(other)`
(same goes for 3 lines below)

blockdag/reachability_test.go
```diff
@@ -624,26 +624,26 @@ func BenchmarkReindexInterval(b *testing.B) {
...
 		b.StartTimer()
-		_, err := currentTreeNode.addChild(fullReindexTriggeringNode)
+		_, err := currentTreeNode.addChild(fullReindexTriggeringNode, root)
 		b.StopTimer()
 		if err != nil {
 			b.Fatalf("addChild: %s", err)
 		}
 
-		if *root.remainingInterval == remainingIntervalBefore {
+		if *root.remainingIntervalAfter() == remainingIntervalBefore {
```
michaelsutton, line 646: Before and After here are confusing... (Before means before the `addChild `call while After `root.remainingIntervalAfter)` is related to interval logic

blockdag/reachability.go
```diff
@@ -547,9 +811,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+		reindexRootEnd - reindexRootChildNodesAfterChosenSizesSum - reachabilityReindexSlack - 1
+	newReindexRootChildInterval := newReachabilityInterval(targetRangeForReindexRootChildStart, targetRangeForReindexRootChildEnd)
+
+	if targetRangeForReindexRootChildStart > reindexRootStart || targetRangeForReindexRootChildEnd < reindexRootEnd {
+		// New interval doesn't contain the previous one, propagation is required
+		chosenReindexRootChild.interval = newReachabilityInterval(
+			targetRangeForReindexRootChildStart+reachabilityReindexSlack,
+			targetRangeForReindexRootChildEnd-reachabilityReindexSlack-1,
+		)
+		modifiedNodes, err := chosenReindexRootChild.countSubtreesAndPropagateInterval()
```
michaelsutton, line 1020: After the propagate call, interval should be set back to:
`		chosenReindexRootChild.interval = newReachabilityInterval(
			targetRangeForReindexRootChildStart,
			targetRangeForReindexRootChildEnd,
		)`
see [here ](https://github.com/kaspanet/pocs/blob/master/ghostdag/block.py#L387) for reference

blockdag/reachability.go
michaelsutton: `isAncestorOf` is kind of the same logic but the name would be confusing if used here. Here it's more about simple interval "algebra"

blockdag/reachability.go
```diff
@@ -547,9 +814,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+
+func (dag *BlockDAG) tightenReachabilityTreeIntervalsAfterChosenReindexRootChild(
+	reindexRoot *reachabilityTreeNode, reindexRootChildNodesAfterChosen []*reachabilityTreeNode) (
+	reindexRootChildNodesAfterChosenSizesSum uint64, modifiedTreeNodes map[*reachabilityTreeNode]struct{}, err error) {
+
+	reindexRootChildNodesAfterChosenSizes, reindexRootChildNodesAfterChosenSubtreeSizeMaps, reindexRootChildNodesAfterChosenSizesSum :=
+		calcReachabilityTreeNodeSizes(reindexRootChildNodesAfterChosen)
+
+	reindexRootEnd := reindexRoot.interval.end
+	targetRangeAfterReindexRootEnd := reindexRootEnd - reachabilityReindexSlack
```
someone235, line 993: For clarity, redefine it so targetRangeAfterReindexRootEnd will be subtracted by -1 in the first place

blockdag/reachability.go
```diff
@@ -547,9 +814,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+
+	reindexRootStart := reindexRoot.interval.start
+	reindexRootEnd := reindexRoot.interval.end
+	targetRangeForReindexRootChildStart :=
+		reindexRootStart + reindexRootChildNodesBeforeChosenSizesSum + reachabilityReindexSlack
+	targetRangeForReindexRootChildEnd :=
+		reindexRootEnd - reindexRootChildNodesAfterChosenSizesSum - reachabilityReindexSlack - 1
+	newReindexRootChildInterval := newReachabilityInterval(targetRangeForReindexRootChildStart, targetRangeForReindexRootChildEnd)
+
+	if targetRangeForReindexRootChildStart > reindexRootStart || targetRangeForReindexRootChildEnd < reindexRootEnd {
```
someone235, line 1019: Isn't it always true?

blockdag/reachability_test.go
```diff
@@ -50,7 +50,7 @@ func TestAddChild(t *testing.T) {
 	if tipInterval != 1 {
 		t.Fatalf("TestAddChild: unexpected tip interval size: want: 1, got: %d", tipInterval)
 	}
-	tipRemainingInterval := lastChild.remainingInterval.size()
+	tipRemainingInterval := lastChild.remainingIntervalAfter().size()
```
someone235, line 53: Don't we wanna check the beforeInterval as well?

blockdag/reachability_test.go
```diff
@@ -624,26 +624,26 @@ func BenchmarkReindexInterval(b *testing.B) {
...
+			_, err := currentTreeNode.addChild(childTreeNode, root)
 			if err != nil {
 				b.Fatalf("addChild: %s", err)
 			}
 
 			currentTreeNode = childTreeNode
 		}
 
-		remainingIntervalBefore := *root.remainingInterval
+		remainingIntervalBefore := *root.remainingIntervalAfter()
```
someone235, line 635: The name is now ambiguous

blockdag/dag.go
```diff
@@ -149,13 +149,14 @@ type BlockDAG struct {
...
 
-	lastFinalityPoint       *blockNode
-	reachabilityReindexRoot *reachabilityTreeNode
+	lastFinalityPoint *blockNode
 
 	utxoDiffStore     *utxoDiffStore
 	reachabilityStore *reachabilityStore
 	multisetStore     *multisetStore
 
+	reachabilityTree *reachabilityTree
```
someone235, line 158: Consider putting any dag reachability methods inside reachabilityTree

blockdag/dag.go
```diff
@@ -2043,6 +2044,7 @@ func New(config *Config) (*BlockDAG, error) {
 	dag.utxoDiffStore = newUTXODiffStore(dag)
 	dag.reachabilityStore = newReachabilityStore(dag)
 	dag.multisetStore = newMultisetStore(dag)
+	dag.reachabilityTree = newReachabilityTree(nil)
```
someone235, line 2047: Why pass nil if you never pass something else

blockdag/reachability.go
```diff
@@ -547,9 +814,276 @@ func (dag *BlockDAG) updateReachability(node *blockNode, selectedParentAnticone
...
+	if node.blueScore > dag.SelectedTipBlueScore() {
+		updateStartTime := time.Now()
+		modifiedTreeNodes, err := dag.updateReachabilityReindexRoot(newTreeNode)
+		if err != nil {
+			return err
+		}
+		if len(modifiedTreeNodes) > 0 {
+			updateTimeElapsed := time.Since(updateStartTime)
+			log.Debugf("Reachability reindex root updated to %s. "+
+				"Modified %d tree nodes and took %dms.",
```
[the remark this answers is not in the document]
stasatdaglabs: I argue that if it's under 1ms it's not interesting to look at. We only care if reachability takes a long time.

blockdag/reachability.go
```diff
@@ -227,23 +283,17 @@ func (rtn *reachabilityTreeNode) addChild(child *reachabilityTreeNode) ([]*reach
 	}
 
 	// Allocate from the remaining space
-	allocated, remaining, err := rtn.remainingInterval.splitInHalf()
+	allocated, _, err := remaining.splitInHalf()
```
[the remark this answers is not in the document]
stasatdaglabs: We talked about it and decided to leave it as is.

blockdag/reachability.go
```diff
@@ -366,14 +416,199 @@ func (rtn *reachabilityTreeNode) propagateInterval(subTreeSizeMap map[*reachabil
...
+	childrenBeforeChosen, _, err := rtn.splitChildrenAroundChosenChild(chosenChild)
+	if err != nil {
+		return nil, err
+	}
+
+	childrenBeforeChosenSizes, childrenBeforeChosenSubtreeSizeMaps, childrenBeforeChosenSizesSum :=
+		calcReachabilityTreeNodeSizes(childrenBeforeChosen)
+
+	// Apply a tight interval
+	newIntervalEnd := childrenBeforeChosen[len(childrenBeforeChosen)-1].interval.end + 1
```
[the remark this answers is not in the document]
michaelsutton: You are probably correct, but it's worth a validation line (not checked in...) and running tests before changing 
e.g. `assert childrenBeforeChosen[len(childrenBeforeChosen)-1].interval.end + 1 == chosenChild.interval.start - 1`

blockdag/reachability.go
```diff
@@ -382,6 +617,26 @@ func (rtn *reachabilityTreeNode) isAncestorOf(other *reachabilityTreeNode) bool
 	return rtn.interval.isAncestorOf(other.interval)
 }
 
+// findCommonAncestor finds the most recent reachability tree ancestor
+// common to both rtn and other.
+func (rtn *reachabilityTreeNode) findCommonAncestor(other *reachabilityTreeNode) *reachabilityTreeNode {
+	currentThis := rtn
```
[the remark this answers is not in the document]
michaelsutton: How?

blockdag/reachability.go
stasatdaglabs: I think I prefer keeping it the way it is, since findIndex searches for ancestors rather than direct parent-child relationships.

blockdag/reachability.go
stasatdaglabs: That's a bit ungrammatical. I renamed it to intervalRangeForChildAllocation to make it more clear.

blockdag/reachability.go
stasatdaglabs: I can't think of a way of restructuring it as you propose without completely rewriting the logic.
Perhaps we should open a ticket and handle it later?

blockdag/reachability.go
```diff
@@ -8,6 +8,39 @@ import (
...
+	// tree nodes not in the selected parent chain. Note that this is not
+	// a constant for testing purposes.
+	reachabilityReindexSlack uint64 = 1 << 12
+)
+
+// modifiedTreeNodes are a set of reachabilityTreeNodes that's bubbled up
+// from any function that modifies them, so that the original caller may
+// update the database accordingly. This is a set rather than a slice due
+// to frequent duplicate treeNodes between operations.
+type modifiedTreeNodes map[*reachabilityTreeNode]struct{}
```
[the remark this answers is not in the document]
stasatdaglabs: We spoke about it and decided to leave it as is.

blockdag/reachability.go
```diff
@@ -665,125 +671,127 @@ func (rtn *reachabilityTreeNode) String() string {
...
 
-// findIndex finds the index of the block with the maximum start that is below
-// the given block.
-func (fb futureCoveringBlockSet) findIndex(block *futureCoveringBlock) int {
-	blockInterval := block.treeNode.interval
+// findAncestorIndexOfNode finds the index of the reachability tree
+// ancestor of `node` among the nodes in `tns`. It does so by finding
+// the index of the block with the maximum start that is below the
+// given block.
+func (tns treeNodeSet) findAncestorIndexOfNode(node *reachabilityTreeNode) (int, bool) {
```
someone235, line 767: please return uint

blockdag/reachability.go
```diff
@@ -665,125 +671,127 @@ func (rtn *reachabilityTreeNode) String() string {
...
 		if end < middleInterval.start {
 			high = middle
 		} else {
 			low = middle + 1
 		}
 	}
-	return low
+
+	if low == 0 {
+		return -1, false
```
someone235: Please return 0, false

blockdag/reachability.go
```diff
@@ -665,125 +671,127 @@ func (rtn *reachabilityTreeNode) String() string {
...
-	if i > 0 {
-		candidate := (*fb)[i-1]
-		candidateInterval := candidate.treeNode.interval
+func (fb *futureCoveringTreeNodeSet) insertNode(node *reachabilityTreeNode) {
+	blockInterval := node.interval
+	ancestorIndex, ok := treeNodeSet(*fb).findAncestorIndexOfNode(node)
+	if ok {
+		candidate := (*fb)[ancestorIndex]
+		candidateInterval := candidate.interval
 		if candidateInterval.isAncestorOf(blockInterval) {
```
someone235: No need to use intervals

blockdag/reachability.go
```diff
@@ -665,125 +671,127 @@ func (rtn *reachabilityTreeNode) String() string {
...
-			// block is an ancestor of candidate, and can thus replace it
-			(*fb)[i-1] = block
+			// node is an ancestor of candidate, and can thus replace it
+			(*fb)[ancestorIndex] = node
 			return
 		}
 	}
 
-	// Insert block in the correct index to maintain futureCoveringBlockSet as
+	// Insert node in the correct index to maintain futureCoveringTreeNodeSet as
```
someone235, line 726: Explicitly take care of `!ok`

blockdag/reachability.go
```diff
@@ -1013,19 +1020,16 @@ func (rt *reachabilityTree) expandIntervalInReindexRootChosenChild(reindexRoot *
```
someone235, line 1024: chosenReindexRootChild should be renamed to reindexRootChosenChild (like the call site)

blockdag/reachability.go
```diff
@@ -1013,19 +1020,16 @@ func (rt *reachabilityTree) expandIntervalInReindexRootChosenChild(reindexRoot *
...
+		reindexRoot.interval.end-reindexRootChildNodesAfterChosenSizesSum-reachabilityReindexSlack-1,
+	)
 
-	if targetRangeForReindexRootChildStart > reindexRootStart || targetRangeForReindexRootChildEnd < reindexRootEnd {
+	if !newReindexRootChildInterval.contains(chosenReindexRootChild.interval) {
 		// New interval doesn't contain the previous one, propagation is required
 		chosenReindexRootChild.interval = newReachabilityInterval(
-			targetRangeForReindexRootChildStart+reachabilityReindexSlack,
-			targetRangeForReindexRootChildEnd-reachabilityReindexSlack-1,
+			newReindexRootChildInterval.start+reachabilityReindexSlack,
```
someone235, line 1035: Please add a comment that explains why you temporarily change the interval

blockdag/dag.go
```diff
@@ -712,9 +713,10 @@ func (dag *BlockDAG) saveChangesFromBlock(block *util.Block, virtualUTXODiff *UT
 
 	// Update DAG state.
 	state := &dagState{
-		TipHashes:         dag.TipHashes(),
-		LastFinalityPoint: dag.lastFinalityPoint.hash,
-		LocalSubnetworkID: dag.subnetworkID,
+		TipHashes:               dag.TipHashes(),
+		LastFinalityPoint:       dag.lastFinalityPoint.hash,
+		ReachabilityReindexRoot: dag.reachabilityTree.reindexRoot.blockNode.hash,
```
someone235, line 718: Pls put ReachabilityReindexRoot somewhere else and fetch it from db when you init reachabilityTree

blockdag/reachability.go
stasatdaglabs: This function explicitly returns an index, which must be int.

blockdag/reachability.go
```diff
@@ -348,11 +415,11 @@ func (rtn *reachabilityTreeNode) countSubtrees(subTreeSizeMap map[*reachabilityT
 // Subtree intervals are recursively allocated according to subtree sizes and
 // the allocation rule in splitWithExponentialBias. This method returns
 // a list of reachabilityTreeNodes modified by it.
-func (rtn *reachabilityTreeNode) propagateInterval(subTreeSizeMap map[*reachabilityTreeNode]uint64) ([]*reachabilityTreeNode, error) {
-	// We set the interval to reset its remainingInterval, so we could reallocate it while reindexing.
-	rtn.setInterval(rtn.interval)
+func (rtn *reachabilityTreeNode) propagateInterval(subTreeSizeMap map[*reachabilityTreeNode]uint64) (
+	modifiedTreeNodes, error) {
+
+	modifiedTreeNodes := newModifiedTreeNodes()
```
svarogg, line 421: This hides the type modifiedTreeNodes. This is not very good.
Please rename either the type or all the instances of this type

blockdag/reachabilitystore.go
```diff
@@ -57,22 +57,26 @@ func (store *reachabilityStore) setBlockAsDirty(blockHash *daghash.Hash) {
 	store.dirty[*blockHash] = struct{}{}
 }
 
-func reachabilityNotFoundError(node *blockNode) error {
-	return errors.Errorf("Couldn't find reachability data for block %s", node.hash)
+func reachabilityNotFoundError(hash *daghash.Hash) error {
+	return errors.Errorf("Couldn't find reachability data for block %s", hash)
```
someone235, line 61: Convert to small case

stasatdaglabs:

PR #777, opened 2020-06-28
[NOD-1063] Fix a bug in which a block is pointing directly to a block in the selected parent chain below the reindex root

stasatdaglabs:


blockdag/reachability_test.go
```diff
@@ -818,6 +822,37 @@ func TestIsInPast(t *testing.T) {
 	}
 }
 
+func TestAddChildToSelectedParentChainBelowReindexRoot(t *testing.T) {
```
someone235, line 825: Please rename to "points directly to the selected parent chain"

PR #778, opened 2020-06-30
[NOD-1098] Change timestamp precision to one millisecond

someone235:


someone235:

blockdag/validate.go
```diff
@@ -412,17 +413,6 @@ func (dag *BlockDAG) checkBlockHeaderSanity(block *util.Block, flags BehaviorFla
 		}
 	}
 
-	// A block timestamp must not have a greater precision than one second.
```
svarogg, line 415: Don't we want to validate precision is not greater then millisecond now?

util/mstime/mstime.go
```diff
@@ -0,0 +1,121 @@
...
+	return t.time.After(u.time)
+}
+
+// Before reports whether the time instant t is before u.
+func (t Time) Before(u Time) bool {
+	return t.time.Before(u.time)
+}
+
+// Add returns the time t+d.
+// It panics if d has a precision lesser than one millisecond.
```
svarogg, line 63: I think the correct was to say is "precision greater then"

PR #782, opened 2020-07-02
[NOD-684] change simnet block rate to block per ms

someone235:


svarogg:

PR #787, opened 2020-07-07
[NOD-1124] Implement the Flow thread model and architecture

stasatdaglabs:


netadapter/netadapter.go
```diff
@@ -0,0 +1,56 @@
...
+				panic(err)
+			}
+			peer.router.RouteMessage(message)
+		}
+	}
+}
+
+// SetRouterInitializer sets the routerInitializer function
+// for the net adapter
+func (na *NetAdapter) SetRouterInitializer(routerInitializer func(peer *Peer) *Router) {
```
svarogg, line 49: Also create type for routerInitializer

netadapter/netadapter.go
```diff
@@ -0,0 +1,56 @@
...
+		server: server,
+	}
+
+	newConnectionHandler := adapter.buildNewConnectionHandler()
+	server.SetNewConnectionHandler(newConnectionHandler)
+
+	return &adapter, nil
+}
+
+func (na *NetAdapter) buildNewConnectionHandler() func(connection server.Connection) {
```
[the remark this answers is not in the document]
svarogg: Actually, since it returns a hardcoded function, why can't this just be this function and that's it?
Why does the function need a constructor?

netadapter/server/grpc/grpcserver.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	return nil
+}
+
+type gRPCConnection struct{}
+
+// Send sends the given message through the connection
+// This is part of the Connection interface
+func (c *gRPCConnection) Send(message wire.Message) error {
+	return nil
```
svarogg, line 58: Add a TODO comment so it's not lost.
Here in all other unimplemented functions.

Also, preferably panic out of here, unless you need it for testing.

protocol/protocol.go
```diff
@@ -2,19 +2,45 @@ package protocol
...
+	netAdapter.SetRouterInitializer(routerInitializer)
+
+	protocolManager := ProtocolManager{
+		netAdapter: netAdapter,
+	}
+	return &protocolManager, nil
 }
 
-func startDummy(server p2pserver.Server, connection p2pserver.Connection, dag *blockdag.BlockDAG) chan<- wire.Message {
+func buildRouterInitializer(netAdapter *netadapter.NetAdapter,
```
svarogg, line 30: build -> new

protocol/protocol.go
```diff
@@ -2,19 +2,45 @@ package protocol
...
+	dag *blockdag.BlockDAG) func(peer *netadapter.Peer) *netadapter.Router {
```
svarogg, line 31: All func types should have types defined for them.

netadapter/router.go
```diff
@@ -0,0 +1,20 @@
...
+
+type Router struct {
+	routes map[string]chan<- wire.Message
+}
+
+// AddRoute registers the messages of types `messageTypes` to
+// be routed to the given `inChannel`
+func (r *Router) AddRoute(messageTypes []string, inChannel chan<- wire.Message) {
+	for _, messageType := range messageTypes {
+		r.routes[messageType] = inChannel
```
someone235, line 13: return error if a route for this message type already exists

protocol/peerstate.go
```diff
@@ -0,0 +1,7 @@
+package protocol
+
+import "github.com/kaspanet/kaspad/util/daghash"
+
+type PeerState struct {
+	requestedBlocks map[*daghash.Hash]struct{}
```
someone235, line 6: This field is not needed in PeerState

protocol/peerstate.go
```diff
@@ -0,0 +1,7 @@
+package protocol
+
+import "github.com/kaspanet/kaspad/util/daghash"
+
+type PeerState struct {
```
someone235, line 5: A dependency cycle will be created if it's inside protocol package

netadapter/netadapter.go
```diff
@@ -0,0 +1,56 @@
...
+	return func(connection server.Connection) {
+		peer := Peer{}
+		peer.connection = connection
+		peer.router = na.routerInitializer(&peer)
+
+		for {
+			message, err := peer.connection.Receive()
+			if err != nil {
+				// TODO: properly handle error
```
someone235, line 40: Use `TODO(libp2p)` so it'll be easier to find it later

netadapter/peer.go
```diff
@@ -0,0 +1,15 @@
+package netadapter
+
+import (
+	"github.com/kaspanet/kaspad/netadapter/server"
+	"github.com/kaspanet/kaspad/wire"
+)
+
+type Peer struct {
+	connection server.Connection
+	router     *Router
```
someone235, line 10: why do we need it?

netadapter/netadapter.go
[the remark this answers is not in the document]
stasatdaglabs: > Why the build prefix?
I didn't want to call it newNewConnectionHandler, and since it's a constructor, I didn't want to omit the verb as you do with e.g. getters in Go.
> Why does the function need a constructor?
It needs `(na *NetAdapter)`.

protocol/protocol.go
stasatdaglabs: Please see https://github.com/kaspanet/kaspad/pull/787#discussion_r450910307

svarogg:

stasatdaglabs:

stasatdaglabs:

PR #789, opened 2020-07-08
[NOD-1124] Implement the Flow thread model and architecture

stasatdaglabs:


netadapter/id.go
```diff
@@ -0,0 +1,36 @@
...
+	bytes := make([]byte, idLength)
+	_, err := rand.Read(bytes)
+	if err != nil {
+		return nil, err
+	}
+	return CreateID(bytes)
+}
+
+// CreateID creates an ID from the given bytes
+func CreateID(bytes []byte) (*ID, error) {
```
svarogg, line 27: It's a convention that all constructor names start with `new`

netadapter/netadapter.go
```diff
@@ -43,31 +61,118 @@ func (na *NetAdapter) Start() error {
 
 // Stop safely closes the NetAdapter
 func (na *NetAdapter) Stop() error {
+	if atomic.AddInt32(&na.stop, 1) != 1 {
+		log.Warnf("Net adapter stopped more than once")
+		return nil
+	}
 	return na.server.Stop()
```
svarogg, line 67: I don't think we want to `Stop` na.server twice either, bad things might happen.
So we need to either add the atomic stuff to server and anything under it as well, or insert the na.server.Stop() into the if. I prefer the later

netadapter/netadapter.go
```diff
@@ -43,31 +61,118 @@ func (na *NetAdapter) Start() error {
...
+		}
+		connection.SetOnDisconnectedHandler(func() error {
+			na.unregisterConnection(connection)
+			return router.Close()
+		})
+		router.SetOnIDReceivedHandler(func(id *ID) {
+			na.registerConnection(connection, router, id)
+		})
+
+		na.startReceiveLoop(connection, router)
```
svarogg, line 84: I prefer when the function is already the loop, and then the caller spawns it

netadapter/netadapter.go
```diff
@@ -43,31 +61,118 @@ func (na *NetAdapter) Start() error {
...
+				if err != nil {
+					log.Warnf("Failed to disconnect from %s: %s", connection, err)
+				}
+				return
+			}
+
+			message, err := connection.Receive()
+			if err != nil {
+				log.Warnf("Failed to receive from %s: %s", connection, err)
+				err := connection.Disconnect()
```
svarogg, line 121: Since you are disconnecting, you should probably return afterwards, otherwise the loop is stuck forever

netadapter/netadapter.go
```diff
@@ -43,31 +61,118 @@ func (na *NetAdapter) Start() error {
...
+				return
+			}
+
+			message := router.TakeOutputMessage()
+			err := connection.Send(message)
 			if err != nil {
-				// TODO(libp2p): properly handle error
-				panic(err)
+				log.Warnf("Failed to send to %s: %s", connection, err)
+				err := connection.Disconnect()
```
svarogg, line 146: Here too

netadapter/router.go
```diff
@@ -5,27 +5,83 @@ import (
 	"github.com/pkg/errors"
 )
 
+// OnIDReceivedHandler is a function that is to be called
+// once a new Connection is sends us its ID.
```
svarogg, line 9: Grammar: is sends us

netadapter/router.go
```diff
@@ -5,27 +5,83 @@ import (
...
 		}
-		r.routes[messageType] = inChannel
+		r.inputRoutes[messageType] = inputChannel
+	}
+	return nil
+}
+
+// RemoveRoute unregisters the messages of types `messageTypes` from
+// the router
+func (r *Router) RemoveRoute(messageTypes []string) error {
```
svarogg, line 48: Will this ever be used?

netadapter/router.go
```diff
@@ -5,27 +5,83 @@ import (
...
+}
+
+// Close shuts down the router by closing all registered
+// input channels
+func (r *Router) Close() error {
+	inChannels := make(map[chan<- wire.Message]struct{})
+	for _, inChannel := range r.inputRoutes {
+		inChannels[inChannel] = struct{}{}
+	}
+	for inChannel := range inChannels {
```
svarogg, line 83: Why the double loop?

netadapter/server/grpcserver/grpcserver.go
```diff
@@ -48,7 +48,9 @@ func (s *gRPCServer) Connections() []server.Connection {
 	panic("unimplemented")
 }
 
-type gRPCConnection struct{}
+type gRPCConnection struct {
+	onDisconnectedHandler server.OnDisconnectedHandler
```
svarogg, line 52: Just to make sure: I'm expected to call onDisconnectedHandler when the other side disconnects - right?
Should I also call it when .Disconnect is called()?

netadapter/netadapter.go
```diff
@@ -108,48 +107,48 @@ func (na *NetAdapter) unregisterConnection(connection server.Connection) {
 func (na *NetAdapter) startReceiveLoop(connection server.Connection, router *Router) {
 	spawn(func() {
 		for {
-			if atomic.LoadInt32(&na.stop) != 0 {
-				err := connection.Disconnect()
-				if err != nil {
-					log.Warnf("Failed to disconnect from %s: %s", connection, err)
-				}
-				return
+			if atomic.LoadUint32(&na.stop) != 0 {
```
someone235: maybe move to the for condition?

netadapter/netadapter.go
```diff
@@ -43,31 +61,117 @@ func (na *NetAdapter) Start() error {
...
+				log.Warnf("Failed to receive from %s: %s", connection, err)
+				break
 			}
-			router.RouteMessage(message)
 		}
-	}
+
+		err := connection.Disconnect()
+		if err != nil {
+			log.Warnf("Failed to disconnect from %s: %s", connection, err)
```
someone235: This will always happen if the other peer initiate a disconnection, no?

netadapter/router.go
stasatdaglabs: Yep, Ori asked for it.

netadapter/server/grpcserver/grpcserver.go
stasatdaglabs: Yep!

netadapter/router.go
stasatdaglabs: Routes generally point N messages to 1 channel and I don't want to call close() on a channel more than once.

PR #792, opened 2020-07-13
[NOD-1137] Implement handshake protocol

someone235:


protocol/sendversion/sendversion.go
```diff
@@ -0,0 +1,62 @@
...
+// SendVersion sends a version to a peer and waits for verack.
+func SendVersion(msgChan <-chan wire.Message, router *netadapter.Router, netAdapter *netadapter.NetAdapter,
+	dag *blockdag.BlockDAG) error {
+
+	selectedTipHash := dag.SelectedTipHash()
+	subnetworkID := config.ActiveConfig().SubnetworkID
+
+	// Version message.
+	msg := wire.NewMsgVersion(netAdapter.GetBestLocalAddress(), netAdapter.ID(), selectedTipHash, subnetworkID)
+	err := msg.AddUserAgent(userAgentName, userAgentVersion, config.ActiveConfig().UserAgentComments...)
```
svarogg, line 40: MsgVersion.AddUserAgent is called only when sending the user agent.
As such, why do we really want to validate it, and then panic if it fails? Why not just accept it as is 

(We can write a unit test to validate it, to make sure the default one is OK, and then on start we should validate it to make sure that a custom one is OK)

wire/common.go
```diff
@@ -128,6 +129,19 @@ func ReadElement(r io.Reader, element interface{}) error {
 		}
 		return nil
 
+	case **id.ID:
```
svarogg, line 132: Why would we ever pass a double-pointer here?

protocol/sendversion/sendversion.go
someone235: You also validate it on Decode and Encode, but I don't care if it'll be validated only there

wire/common.go
someone235: Otherwise we need to copy the whole ID, but I guess it's ok

netadapter/netadapter.go
```diff
@@ -165,3 +167,34 @@ func (na *NetAdapter) Broadcast(connectionIDs []*id.ID, message wire.Message) er
 	}
 	return nil
 }
+
+// GetBestLocalAddress returns the most appropriate local address to use
+// for the given remote address.
+func (na *NetAdapter) GetBestLocalAddress() (*wire.NetAddress, error) {
+	//TODO(libp2p) Reimplement this, and check reachability to the other node
+	listenAddr := config.ActiveConfig().Listeners[0]
```
stasatdaglabs, line 175: Please rename listenAddr to listenAddress.

netadapter/netadapter.go
```diff
@@ -165,3 +167,34 @@ func (na *NetAdapter) Broadcast(connectionIDs []*id.ID, message wire.Message) er
...
+	_, portStr, err := net.SplitHostPort(listenAddr)
```
stasatdaglabs, line 176: Please rename portStr to portString.

netadapter/netadapter.go
```diff
@@ -165,3 +167,34 @@ func (na *NetAdapter) Broadcast(connectionIDs []*id.ID, message wire.Message) er
...
+	if err != nil {
+		portStr = config.ActiveConfig().NetParams().DefaultPort
+	}
+
+	portInt, err := strconv.Atoi(portStr)
+	if err != nil {
+		return nil, err
+	}
+
+	addrs, err := net.InterfaceAddrs()
```
stasatdaglabs, line 186: Please rename addrs to addresses.

netadapter/netadapter.go
```diff
@@ -165,3 +167,34 @@ func (na *NetAdapter) Broadcast(connectionIDs []*id.ID, message wire.Message) er
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, addr := range addrs {
```
stasatdaglabs, line 191: Please rename addr to address.

netadapter/netadapter.go
```diff
@@ -165,3 +167,34 @@ func (na *NetAdapter) Broadcast(connectionIDs []*id.ID, message wire.Message) er
...
+		ifaceIP, _, err := net.ParseCIDR(addr.String())
```
stasatdaglabs, line 192: Please rename ifaceIP to interfaceIP.

protocol/handshake.go
```diff
@@ -0,0 +1,103 @@
...
+	case err := <-errChan:
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	case <-locks.ReceiveFromChanWhenDone(func() { wg.Wait() }):
+	}
+
+	err = peerpkg.AddToReadyPeers(peer)
+	if errors.Is(err, peerpkg.ErrPeerWithSameIDExists) {
```
stasatdaglabs, line 78: Please move this to within the err != nil check.

protocol/peer/peer.go
```diff
@@ -92,17 +101,46 @@ func (p *Peer) String() string {
...
 }
 
-// minUint32 is a helper function to return the minimum of two uint32s.
-// This avoids a math import and the need to cast to floats.
-func minUint32(a, b uint32) uint32 {
-	if a < b {
-		return a
+var (
+	readyPeers    = make(map[*id.ID]*Peer, 0)
+	readyPeersMtx sync.RWMutex
```
stasatdaglabs, line 106: Please rename this to readyPeersMutex.

protocol/receiveversion/receiveversion.go
```diff
@@ -0,0 +1,79 @@
...
+	// connected peer may support.
+	minAcceptableProtocolVersion = wire.ProtocolVersion
+)
+
+// ReceiveVersion waits for the peer to send a version message, sends a
+// verack in response, and updates its info accordingly.
+func ReceiveVersion(incomingRoute *router.Route, outgoingRoute *router.Route, peer *peerpkg.Peer,
+	dag *blockdag.BlockDAG) (addr *wire.NetAddress, routeClosed bool, err error) {
+
+	msg, isClosed := incomingRoute.Dequeue()
```
stasatdaglabs, line 27: Please rename msg to message.

stasatdaglabs: Dequeue returns isOpen, actually.
Please go over the code and make sure you didn't make the same error anywhere else.

protocol/peer/peer.go
```diff
@@ -92,17 +101,46 @@ func (p *Peer) String() string {
...
 }
 
 // GetReadyPeerIDs returns the peer IDs of all the ready peers.
 func GetReadyPeerIDs() []*id.ID {
-	// TODO(libp2p)
-	panic("unimplemented")
+	readyPeersMutex.RLock()
+	defer readyPeersMutex.RUnlock()
+	peerIDs := make([]*id.ID, len(readyPeers))
+	i := 0
```
stasatdaglabs, line 135: This variable never seems to be incremented.

protocol/handshake.go
```diff
@@ -0,0 +1,103 @@
...
+	"sync"
+	"sync/atomic"
+)
+
+func handshake(router *routerpkg.Router, netAdapter *netadapter.NetAdapter, peer *peerpkg.Peer,
+	dag *blockdag.BlockDAG, addressManager *addrmgr.AddrManager) (closed bool, err error) {
+
+	receiveVersionRoute, err := router.AddIncomingRoute([]string{wire.CmdVersion})
+	if err != nil {
+		panic(err)
```
svarogg, line 23: No panic, ruleError vs normal error, as agreed on zoom

protocol/handshake.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/kaspanet/kaspad/wire"
+	"github.com/pkg/errors"
+	"sync"
+	"sync/atomic"
+)
+
+func handshake(router *routerpkg.Router, netAdapter *netadapter.NetAdapter, peer *peerpkg.Peer,
+	dag *blockdag.BlockDAG, addressManager *addrmgr.AddrManager) (closed bool, err error) {
+
+	receiveVersionRoute, err := router.AddIncomingRoute([]string{wire.CmdVersion})
```
svarogg, line 21: Do we ever remove these routes?
What will happen if peer sends version/verack after handshake is over?

wire/message_test.go
```diff
@@ -34,6 +35,16 @@ func makeHeader(kaspaNet KaspaNet, command string,
 	return buf
 }
 
+func idFromBytes(idBytes []byte) *id.ID {
```
svarogg, line 38: Maybe move this method to `id` package?
Sounds like someone somewhere will need it.

protocol/handshake.go
```diff
@@ -0,0 +1,103 @@
...
+			return
+		}
+		peerAddress = address
+	})
+
+	spawn(func() {
+		defer wg.Done()
+		closed, err := sendversion.SendVersion(sendVersionRoute, router.OutgoingRoute(), netAdapter, dag)
+		if err != nil || closed {
+			log.Errorf("error from SendVersion: %s", err)
```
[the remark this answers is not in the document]
someone235: It's ok

protocol/handshake.go
```diff
@@ -0,0 +1,103 @@
...
+			if atomic.AddUint32(&errChanUsed, 1) != 1 {
```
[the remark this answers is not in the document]
someone235: I prefer to use atomic because it's similar to the other other flows

netadapter/netadapter.go
```diff
@@ -165,3 +167,58 @@ func (na *NetAdapter) Broadcast(connectionIDs []*id.ID, message wire.Message) er
...
+			if err != nil {
+				return nil, err
+			}
+			ip = net.ParseIP(hostAddrs[0])
+			if ip == nil {
+				return nil, errors.Errorf("Cannot resolve IP address for host '%s'", host)
+			}
+		}
+		return wire.NewNetAddressIPPort(ip, uint16(portInt), wire.SFNodeNetwork), nil
+
```
stasatdaglabs, line 197: This part and the next part are almost entirely identical. Please extract it to a function.
Also, there's a superfluous empty line here.

netadapter/router/route.go
```diff
@@ -61,7 +64,7 @@ func (r *Route) EnqueueWithTimeout(message wire.Message, timeout time.Duration)
 	}
 	select {
 	case <-time.After(timeout):
-		return false, errors.New("timeout expired")
+		return false, ErrTimeout
```
stasatdaglabs: Dequeue should return this error as well.

protocol/handlerelayinvs/handle_relay_invs.go
```diff
@@ -141,6 +142,9 @@ func readMsgBlock(incomingRoute *router.Route, invsQueue *[]*wire.MsgInvRelayBlo
 		const timeout = 30 * time.Second
 		message, isOpen, err := incomingRoute.DequeueWithTimeout(timeout)
 		if err != nil {
+			if !errors.Is(err, router.ErrTimeout) {
+				panic(err)
+			}
```
stasatdaglabs: Please consider not making this check.

protocol/sendversion/sendversion.go
```diff
@@ -0,0 +1,81 @@
...
+	spawn(func() {
+		_, isOpen := incomingRoute.Dequeue()
+		if !isOpen {
+			closeChan <- struct{}{}
+			return
+		}
+		close(gotVerack)
+	})
+
+	const timeout = 30 * time.Second
```
stasatdaglabs, line 72: Please use DequeueWithTimeout instead.

PR #800, opened 2020-07-16
[NOD-1125] Implement the IBD flow

stasatdaglabs:


wire/msgibdblock.go
```diff
@@ -0,0 +1,44 @@
...
+// MaxPayloadLength returns the maximum length the payload can be for the
+// receiver. This is part of the Message interface implementation.
+func (msg *MsgIBDBlock) MaxPayloadLength(pver uint32) uint32 {
+	return MaxMessagePayload
+}
+
+// NewMsgIBDBlock returns a new kaspa ibdblock message that conforms to the
+// Message interface. See MsgIBDBlock for details.
+func NewMsgIBDBlock(msgBlock *MsgBlock) *MsgIBDBlock {
+	return &MsgIBDBlock{*msgBlock}
```
someone235, line 43: Why does it get a pointer instead of getting it by value, if you use the value later?

protocol/ibd/handle_get_block_locator.go
```diff
@@ -0,0 +1,66 @@
...
+	if !isOpen {
+		return nil, nil, false, nil
+	}
+	msgGetBlockLocator := message.(*wire.MsgGetBlockLocator)
+
+	return msgGetBlockLocator.LowHash, msgGetBlockLocator.HighHash, true, nil
+}
+
+func sendBlockLocator(outgoingRoute *router.Route, locator blockdag.BlockLocator) (shouldContinue bool, err error) {
+	msgBlockLocator := wire.NewMsgBlockLocator()
```
svarogg, line 52: `blockdag.BlockLocator` is `[]*daghash.Hash`
`MsgBlockLocator` is `struct {	BlockLocatorHashes []*daghash.Hash }`

I suggest we move BlockLocator to wire package (However, leave NewBlockLocator in blockdag, since it is dag-dependant), then wire.NewMsgBlockLocator will simply assign the array to the inner field, and we'll save the loop below.

protocol/ibd/handle_get_blocks.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		msgIBDBlocks[i] = wire.NewMsgIBDBlock(block.MsgBlock())
+	}
+
+	return msgIBDBlocks, nil
+}
+
+func sendMsgIBDBlocks(outgoingRoute *router.Route, msgIBDBlocks []*wire.MsgIBDBlock) (shouldContinue bool, err error) {
+	for _, msgIBDBlock := range msgIBDBlocks {
+		isOpen, err := outgoingRoute.EnqueueWithTimeout(msgIBDBlock, common.DefaultTimeout)
```
svarogg, line 71: This is dangerous, since we are sending here quite a mass of blocks, we run a real chance of clogging the buffer, and the only thing that might save us is that we'll be able to finish all the sending within common.DefaultTimeout.

I don't think we should rely on DefaultTimeout in this case. 
Even more then that - we shouldn't rely on the buffer at all, since other messages might be coming simultaneously from other flows.

Instead, we need some indication that block was actually sent before sending the next one.

protocol/ibd/ibd.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// HandleIBD waits for IBD start and handles it when IBD is triggered for this peer
+func HandleIBD(incomingRoute *router.Route, outgoingRoute *router.Route,
+	peer *peerpkg.Peer, dag *blockdag.BlockDAG) error {
+
+	for {
+		peer.WaitForIBDStart()
+
+		// We run the flow inside a func so that the defer is called at its end
```
svarogg, line 66: I suggest creating a separate func, so that this one is only flow control (waiting for start, then calling the actual flow)

protocol/ibd/ibd.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+		if !shouldContinue {
+			return nil, false, nil
+		}
+
+		// We check whether the locator's highest hash is in the local DAG.
+		// If it isn't, we need to narrow our getBlockLocator request and
+		// try again.
+		locatorHighHash := blockLocatorHashes[0]
+		if !dag.IsInDAG(locatorHighHash) {
```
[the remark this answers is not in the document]
svarogg: I tend to agree with Ori. The "stop condition" should be in the if

protocol/ibd/selected_tip.go
```diff
@@ -0,0 +1,133 @@
...
+func RequestSelectedTip(incomingRoute *router.Route,
+	outgoingRoute *router.Route, peer *peerpkg.Peer, dag *blockdag.BlockDAG) error {
+	for {
+		peer.WaitForSelectedTipRequests()
+
+		// We run the flow inside a func so that the defer is called at its end
+		err := func() error {
+			defer peer.FinishRequestingSelectedTip()
+
+			shouldContinue, err := requestSelectedTip(outgoingRoute)
```
svarogg, line 44: I think requestSelectedTip and receiveSelectedTip can be combined to a single function, unless we ever want to call them separately, which I highly doubt

protocol/ibd/selected_tip.go
```diff
@@ -0,0 +1,133 @@
...
+
+// RequestSelectedTip waits for selected tip requests and handles them
+func RequestSelectedTip(incomingRoute *router.Route,
+	outgoingRoute *router.Route, peer *peerpkg.Peer, dag *blockdag.BlockDAG) error {
+	for {
+		peer.WaitForSelectedTipRequests()
+
+		// We run the flow inside a func so that the defer is called at its end
+		err := func() error {
+			defer peer.FinishRequestingSelectedTip()
```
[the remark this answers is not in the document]
svarogg: In case it's decided that the defer is important - you should move it to after the call for requestSelectedTip

protocol/ibd/selected_tip.go
```diff
@@ -0,0 +1,133 @@
...
+		}()
+		if err != nil {
+			return err
+		}
+	}
+}
+
+func requestSelectedTip(outgoingRoute *router.Route) (shouldContinue bool, err error) {
+	msgGetSelectedTip := wire.NewMsgGetSelectedTip()
+	isOpen, err := outgoingRoute.EnqueueWithTimeout(msgGetSelectedTip, common.DefaultTimeout)
```
svarogg, line 71: You can `return outgoingRoute.EnqueueWithTimeout(//...)` at this point, will be equivalent 
(assuming you are not combining requestSelectedTip and receiveSelectedTip as requested in previous comment)

protocol/ibd/selected_tip.go
```diff
@@ -0,0 +1,133 @@
...
+	return requestSelectedTips(dag)
+}
+
+func hasRecentlyReceivedBlock(dag *blockdag.BlockDAG) bool {
+	return dag.Now().Sub(dag.SelectedTipHeader().Timestamp) > minDurationToRequestSelectedTips
+}
+
+func requestSelectedTips(dag *blockdag.BlockDAG) error {
+	for _, peer := range peerpkg.ReadyPeers() {
+		peer.RequestSelectedTipIfRequired()
```
svarogg, line 29: I might be missing something, but I couldn't find the place where these selectedTips are being received.

protocol/peer/peer.go
```diff
@@ -169,3 +186,52 @@ func IDExists(peerID *id.ID) bool {
...
+// request flow.
+func (p *Peer) RequestSelectedTipIfRequired() {
+	if atomic.LoadUint32(&p.isSelectedPeerRequested) != 0 {
+		return
+	}
+	if mstime.Since(p.lastSelectedTipRequest) < minGetSelectedTipInterval {
+		return
+	}
+
+	atomic.StoreUint32(&p.isSelectedPeerRequested, 1)
```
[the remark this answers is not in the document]
svarogg: Better use SwapUint32 then adding one then subtracting one

protocol/peer/peer.go
```diff
@@ -31,6 +34,20 @@ type Peer struct {
 	lastPingNonce    uint64        // The nonce of the last ping we sent
 	lastPingTime     time.Time     // Time we sent last ping
 	lastPingDuration time.Duration // Time for last ping to return
+
+	isSelectedPeerRequested uint32
```
svarogg, line 38: It seems like here and in `selectedPeerRequestChan` you typo'd `Peer` instead of `tip`

protocol/peer/peer.go
```diff
@@ -169,3 +186,52 @@ func IDExists(peerID *id.ID) bool {
...
+	<-p.selectedPeerRequestChan
+}
+
+// FinishRequestingSelectedTip finishes requesting the selected
+// tip from this peer
+func (p *Peer) FinishRequestingSelectedTip() {
+	atomic.StoreUint32(&p.isSelectedPeerRequested, 0)
+}
+
+// StartIBD notifies the peer that starting IBD is required.
```
svarogg, line 227: You are not notifying the peer, you are notifying the IBD flow

wire/msgibdblock.go
svarogg: It is correct. The fact that MsgIBDBlock embeds MsgBlock by value is not the caller's concern, and the caller is going to have a pointer.

protocol/ibd/ibd.go
```diff
@@ -0,0 +1,244 @@
...
+	isOrphan, isDelayed, err := dag.ProcessBlock(block, blockdag.BFNone)
+	if err != nil {
+		return err
+	}
+	if isOrphan {
+		return protocolerrors.Errorf(true, "received orphan block %s "+
+			"during IBD", block.Hash())
+	}
+	if isDelayed {
+		return protocolerrors.Errorf(true, "received delayed block %s "+
```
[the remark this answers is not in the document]
stasatdaglabs: We don't expect delayed blocks in IBD, so yes.

protocol/ibd/selected_tip.go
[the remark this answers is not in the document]
stasatdaglabs: We spoke about this and decided to keep this as is.

protocol/ibd/ibd.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/kaspanet/kaspad/protocol/protocolerrors"
+	"github.com/kaspanet/kaspad/util"
+	"github.com/kaspanet/kaspad/util/daghash"
+	"github.com/kaspanet/kaspad/wire"
+	"sync"
+	"sync/atomic"
+)
+
+var (
+	isIBDRunning  uint32
```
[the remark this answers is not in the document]
stasatdaglabs: I'm keeping it as is for now, since I don't have a Manager object available at this time.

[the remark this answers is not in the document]
someone235: So please add todo

protocol/ibd/selected_tip.go
stasatdaglabs: We spoke about this and decided to leave it as is.

protocol/ibd/selected_tip.go
stasatdaglabs: 1. `RequestSelectedTip` waits on `peer.selectedTipRequestChan`
2. `RequestSelectedTipIfRequired` does: `peer.selectedTipRequestChan <- struct{}{}`
3. In the end of `RequestSelectedTip`: `peer.SetSelectedTipHash(peerSelectedTipHash)`

protocol/peer/peer.go
```diff
@@ -165,8 +152,52 @@ func GetReadyPeerIDs() []*id.ID {
...
+	<-p.selectedTipRequestChan
+}
+
+// FinishRequestingSelectedTip finishes requesting the selected
+// tip from this peer
+func (p *Peer) FinishRequestingSelectedTip() {
+	atomic.SwapUint32(&p.isSelectedTipRequested, 0)
+}
+
+// StartIBD notifies the peer that starting its IBD flow is required.
```
svarogg, line 193: "StartIBD starts the IBD process with this peer" or something similar

protocol/flows/ibd/selected_tip.go
```diff
@@ -0,0 +1,128 @@
...
+	for _, peer := range peerpkg.ReadyPeers() {
+		peer.RequestSelectedTipIfRequired()
+	}
+}
+
+// RequestSelectedTip waits for selected tip requests and handles them
+func RequestSelectedTip(incomingRoute *router.Route,
+	outgoingRoute *router.Route, peer *peerpkg.Peer, dag *blockdag.BlockDAG) error {
+	for {
+		peer.WaitForSelectedTipRequests()
```
svarogg, line 37: Move into runSelectedTipRequest

protocol/flows/ibd/handle_get_blocks.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	msgGetBlocks := message.(*wire.MsgGetBlocks)
+
+	return msgGetBlocks.LowHash, msgGetBlocks.HighHash, false, nil
+}
+
+func buildMsgIBDBlocks(lowHash *daghash.Hash, highHash *daghash.Hash,
+	dag *blockdag.BlockDAG) ([]*wire.MsgIBDBlock, error) {
+
+	const maxHashesInMsgIBDBlocks = wire.MaxInvPerMsg
```
svarogg, line 48: Make this equal to MaxBlockLocators

protocol/flows/ibd/ibd.go
```diff
@@ -20,40 +20,39 @@ var (
 
 // StartIBDIfRequired selects a peer and starts IBD against it
 // if required
-func StartIBDIfRequired(dag *blockdag.BlockDAG) error {
+func StartIBDIfRequired(dag *blockdag.BlockDAG) {
 	startIBDMutex.Lock()
```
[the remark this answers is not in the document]
stasatdaglabs: I prefer to keep it this way for clarity.

PR #802, opened 2020-07-19
[NOD-1160] Convert *config.Config from singleton to an object that is being passed around 

svarogg:


kaspad.go
```diff
@@ -39,62 +39,60 @@ type kaspad struct {
...
-	err := s.networkAdapter.Stop()
+	err := k.networkAdapter.Stop()
 	if err != nil {
 		log.Errorf("Error stopping the p2p protocol: %+v", err)
 	}
 
-	// Shutdown the RPC server if it's not disabled.
-	if !config.ActiveConfig().DisableRPC {
-		err := s.rpcServer.Stop()
+	// Shutdown the RPC server if it'k not disabled.
```
someone235, line 93: wrong rename

PR #804, opened 2020-07-19
[NOD-1161] Name goroutines and log them by the name

someone235:


util/panics/panics.go
```diff
@@ -74,3 +75,12 @@ func exit(log *logs.Logger, reason string, currentThreadStackTrace []byte, gorou
 	os.Exit(1)
 	fmt.Print("After os.Exit(1)")
 }
+
+func handleSpawnedFunction(log *logs.Logger, stackTrace []byte, spawnedFunctionName string, spawnedFunction func()) {
+	goroutineID := atomic.AddUint64(&goroutineLastID, 1)
```
stasatdaglabs, line 80: According to this, goroutineID starts with 1 instead of 0. Consider working around that somehow.

PR #805, opened 2020-07-20
[NOD-1170] Return a custom error when a route is closed

someone235:


netadapter/router/route.go
```diff
@@ -13,7 +13,12 @@ const (
 )
 
 // ErrTimeout signifies that one of the router functions had a timeout.
-var ErrTimeout = errors.New("timeout expired")
+var (
+	ErrTimeout = errors.New("timeout expired")
```
stasatdaglabs, line 17: The comment to ErrTimeout should be inside the var().

PR #806, opened 2020-07-21
[NOD-1164] Remove the singleton from dbaccess, to enable multiple db connections in same run

svarogg:


blockdag/dag.go
```diff
@@ -162,6 +163,103 @@ type BlockDAG struct {
 	startTime                       mstime.Time
 }
 
+// New returns a BlockDAG instance using the provided configuration details.
+func New(config *Config) (*BlockDAG, error) {
+	// Enforce required config fields.
```
stasatdaglabs, line 168: I would argue that a databaseContext is also required. Please a nil check here. 
Alternatively, remove all the nil checks and let it panic.

dbaccess/context.go
```diff
@@ -11,24 +11,28 @@ type Context interface {
...
-func (*noTxContext) accessor() (database.DataAccessor, error) {
-	return db()
+type noTxContext struct {
+	backend *DatabaseContext
 }
 
-// NoTx creates and returns an instance of dbaccess.Context without an attached database transaction
-func NoTx() Context {
-	return noTxContextSingleton
+func (ctx *noTxContext) accessor() (database.DataAccessor, error) {
```
someone235, line 18: Do we maybe want to get rid of this function and implement it directly as part of `DatabaseContext`?

addrmgr/addrmanager.go
```diff
@@ -58,6 +59,22 @@ type AddrManager struct {
 	localSubnetworkID  *subnetworkid.SubnetworkID
 }
 
+// New returns a new Kaspa address manager.
+// Use Start to begin processing asynchronous address updates.
```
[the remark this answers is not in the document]
svarogg: Why incorrect?

dbaccess/context.go
svarogg: Nope, it makes the API of noTxContext and TxContext similar, allowing the rest of dbaccess function to be agonstic to tx/noTx

PR #808, opened 2020-07-21
[NOD-1168] Add context interfaces for flows

someone235:


protocol/flows/ping/ping.go
```diff
@@ -11,9 +11,13 @@ import (
 	"github.com/kaspanet/kaspad/wire"
 )
 
+// Context is the interface for the context needed for the Ping and Pong flows.
```
svarogg, line 14: Maybe no need for Context if nothing in there?
Still might be good as a placeholder, not sure.

Anyway, if decide to keep it, please rename as in all other places [FlowName]Context

protocol/flowcontext/flow_context.go
```diff
@@ -0,0 +1,55 @@
...
+	addressManager *addrmgr.AddrManager, txPool *mempool.TxPool, netAdapter *netadapter.NetAdapter) *FlowContext {
+	return &FlowContext{
+		cfg:                         cfg,
+		netAdapter:                  netAdapter,
+		dag:                         dag,
+		addressManager:              addressManager,
+		txPool:                      txPool,
+		sharedRequestedTransactions: relaytransactions.NewSharedRequestedTransactions(),
+		sharedRequestedBlocks:       blockrelay.NewSharedRequestedBlocks(),
+	}
```
stasatdaglabs, line 54: readyPeers needs to be initialized here as well.

stasatdaglabs: Likewise transactionsToRebroadcast.

protocol/flows/relaytransactions/shared_requested_transactions.go
```diff
@@ -37,7 +37,7 @@ func (s *SharedRequestedTransactions) addIfNotExists(txID *daghash.TxID) (exists
 	return false
 }
 
-// NewSharedRequestedTransactions returns a new instance of SharedRequestedTransactions.
+// NewSharedRequestedTransactions returns a new instance of *SharedRequestedTransactions.
```
stasatdaglabs: Whether you return it by reference or by value is probably redundant to specify in a comment.

protocol/flows/handshake/handshake.go
```diff
@@ -5,21 +5,30 @@ import (
...
 	routerpkg "github.com/kaspanet/kaspad/netadapter/router"
-	"github.com/kaspanet/kaspad/protocol/flows/ibd"
 	peerpkg "github.com/kaspanet/kaspad/protocol/peer"
 	"github.com/kaspanet/kaspad/util/locks"
 	"github.com/kaspanet/kaspad/wire"
 	"github.com/pkg/errors"
 )
 
+// Context is the interface for the context needed for the HandleHandshake flow.
+type Context interface {
```
stasatdaglabs: Please rename this to HandshakeContext.

PR #809, opened 2020-07-22
[NOD-1175] Implement AddBlock

stasatdaglabs:


protocol/flowcontext/blocks.go
```diff
@@ -49,6 +51,15 @@ func (f *FlowContext) SharedRequestedBlocks() *blockrelay.SharedRequestedBlocks
 
 // AddBlock adds the given block to the DAG and propagates it.
 func (f *FlowContext) AddBlock(block *util.Block) error {
-	// TODO(libp2p): unimplemented
-	panic("unimplemented")
+	isOrphan, isDelayed, err := f.DAG().ProcessBlock(block, blockdag.BFDisallowDelay)
+	if err != nil {
+		return err
+	}
+	if isOrphan {
```
svarogg, line 58: It is a bit problematic that the name of the function is so general - AddBlock, while the implementation implies specifics (since this was submitted from RPC, we do not allow orphan/delay).

The naming/parameters/something should make clear that this function does not allow delayeds and orphans.

In addition, from reading ProcessBlock, it seems like if this block is orphan - it will be added to the orphan pool, which is a problem.

PR #811, opened 2020-07-22
[NOD-1176] Implement a struct for each flow to share flow data

someone235:


protocol/protocol.go
```diff
@@ -139,13 +140,13 @@ func (m *Manager) addIBDFlows(router *routerpkg.Router, stopped *uint32, stop ch
 
 	addFlow("RequestSelectedTip", router, []wire.MessageCommand{wire.CmdSelectedTip}, stopped, stop,
```
stasatdaglabs, line 141: This should probably be moved to a function called addSelectedTipFlows.

protocol/protocol.go
```diff
@@ -139,13 +140,13 @@ func (m *Manager) addIBDFlows(router *routerpkg.Router, stopped *uint32, stop ch
...
 		func(incomingRoute *routerpkg.Route) error {
-			return ibd.RequestSelectedTip(m.context, incomingRoute, outgoingRoute, peer)
+			return selectedtip.RequestSelectedTip(m.context, incomingRoute, outgoingRoute, peer)
 		},
 	)
 
 	addFlow("HandleGetSelectedTip", router, []wire.MessageCommand{wire.CmdGetSelectedTip}, stopped, stop,
```
stasatdaglabs, line 147: This should probably be moved to a function called addSelectedTipFlows.

protocol/flows/ping/receive.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/kaspanet/kaspad/netadapter/router"
+	"github.com/kaspanet/kaspad/wire"
+)
+
+// ReceivePingsContext is the interface for the context needed for the ReceivePings flow.
+type ReceivePingsContext interface {
+}
+
+type receivePingsFlow struct {
+	incomingRoute, outgoingRoute *router.Route
```
svarogg, line 13: I think you want to embed the context into the flow struct.
Since we introduced the context to "comply to the standard" - let's comply with the standard all the way.

protocol/flows/ping/send.go
```diff
@@ -1,46 +1,37 @@
...
-			return err
-		}
-	}
-}
-
 // SendPingsContext is the interface for the context needed for the SendPings flow.
 type SendPingsContext interface {
 }
 
+type sendPingsFlow struct {
```
svarogg, line 17: Same here

someone235:

PR #829, opened 2020-07-28
[NOD-1198] Make router a property of netConnection, and remove map from connection to router in netAdapter

svarogg:


netadapter/netconnection.go
```diff
@@ -54,3 +88,19 @@ func (c *NetConnection) NetAddress() *wire.NetAddress {
 func (c *NetConnection) SetOnInvalidMessageHandler(onInvalidMessageHandler server.OnInvalidMessageHandler) {
 	c.connection.SetOnInvalidMessageHandler(onInvalidMessageHandler)
 }
+
+func (c *NetConnection) setOnDisconnectedHandler(onDisconnectedHandler server.OnDisconnectedHandler) {
```
someone235, line 92: maybe move it to constructor?

netadapter/server/grpcserver/grpc_connection.go
```diff
@@ -89,7 +89,9 @@ func (c *gRPCConnection) Disconnect() error {
 
 	log.Debugf("Disconnected from %s", c)
 
-	return c.onDisconnectedHandler()
+	c.onDisconnectedHandler()
+
+	return nil
```
someone235, line 94: make the function `void`

netadapter/netconnection.go
svarogg: Impossible, since netAdapter does `delete(na.connections, netConnection)`, a.k.a. needs a reference to the netConnection in the disconnectHandler

PR #836, opened 2020-08-02
[NOD-1211] Transaction relay integration test + fixes to flow

svarogg:


integration/tx_relay_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"time"
+
+	"github.com/kaspanet/go-secp256k1"
+	"github.com/kaspanet/kaspad/txscript"
+	"github.com/kaspanet/kaspad/util"
+	"github.com/kaspanet/kaspad/wire"
+)
+
+func TestTxRelay(t *testing.T) {
+	payer, mediator, payee, teardown := standardSetup(t)
```
someone235, line 16: I'm not sure I like the payer/payee terminology, because the node doesn't really pay to each other

netadapter/server/grpcserver/protowire/common.go
```diff
@@ -26,6 +27,9 @@ func protoHashesToWire(protoHashes []*Hash) ([]*daghash.Hash, error) {
 }
 
 func wireHashToProto(hash *daghash.Hash) *Hash {
+	if hash == nil {
```
someone235, line 30: I prefer if it'll be outside of wireHashToProto so you won't get it for required fields

protocol/flows/relaytransactions/handle_relayed_transactions.go
```diff
@@ -145,42 +145,53 @@ func (flow *handleRelayedTransactionsFlow) broadcastAcceptedTransactions(accepte
...
 
-// readMsgTx returns the next msgTx in incomingRoute, and populates invsQueue with any inv messages that meanwhile arrive.
+// readMsgTxOrNotFound returns the next msgTx or msgTransactionNotFound in incomingRoute,
+// returning only one of the message types at a time.
 //
-// Note: this function assumes msgChan can contain only wire.MsgInvTransaction and wire.MsgBlock messages.
-func (flow *handleRelayedTransactionsFlow) readMsgTx() (
-	msgTx *wire.MsgTx, err error) {
+// and populates invsQueue with any inv messages that meanwhile arrive.
+func (flow *handleRelayedTransactionsFlow) readMsgTxOrNotFound() (
```
someone235, line 152: I'm not sure, but I think maybe it should just return wire.Message

wire/message.go
```diff
@@ -43,6 +43,7 @@ const (
 	CmdInvTransaction
 	CmdRequestTransactions
 	CmdIBDBlock
+	CmdNotFound
```
someone235, line 46: Please change (here and in stringer) to "CmdTransactionNotFound" (and in stringer too)

wire/msgtransactionnotfound.go
```diff
@@ -0,0 +1,29 @@
...
+	ID *daghash.TxID
+}
+
+// Command returns the protocol command string for the message. This is part
+// of the Message interface implementation.
+func (msg *MsgTransactionNotFound) Command() MessageCommand {
+	return CmdNotFound
+}
+
+// NewMsgTransactionNotFound returns a new kaspa transactionsnotfound message that conforms to the
```
someone235, line 23: Change transactionsnotfound->TransactionNotFound

integration/tx_relay_test.go
svarogg: appHarness is not only node, it contains an address and privateKey

protocol/flows/relaytransactions/handle_relayed_transactions.go
svarogg: I prefer two return values over having to do runtime type-checks afterwards

wire/msgtransactionnotfound.go
svarogg: I wrote transactionsnotfound because that's what's going on in rest of wire package.

Anyway, fixed here, because it looks better, even if non-compliant.

netadapter/server/grpcserver/protowire/message_transaction.go
```diff
@@ -93,7 +93,10 @@ func (x *TransactionMessage) fromWireMessage(msgTx *wire.MsgTx) {
 		LockTime:     msgTx.LockTime,
 		SubnetworkID: wireSubnetworkIDToProto(&msgTx.SubnetworkID),
 		Gas:          msgTx.Gas,
-		PayloadHash:  wireHashToProto(msgTx.PayloadHash),
 		Payload:      msgTx.Payload,
 	}
+
+	if msgTx.PayloadHash != nil {
```
someone235, line 99: To be more consistent with the current design, can you prepare the payload hash (nil or not) above `*x = TransactionMessage{`?

PR #838, opened 2020-08-02
[NOD-1214] Added test for 64 incoming connections to single node 

svarogg:


svarogg:

integration/64_incoming_connections_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"testing"
+	"time"
+
+	"github.com/kaspanet/kaspad/util/locks"
+
+	"github.com/kaspanet/kaspad/wire"
+)
+
+func Test64IncomingConnections(t *testing.T) {
+	const numBullies = 64 // Much more then 64 hosts creates a risk of running out of available file descriptors
```
someone235, line 15: then->than

integration/64_incoming_connections_test.go
```diff
@@ -0,0 +1,55 @@
...
+	params := make([]*harnessParams, numBullies+1)
```
someone235, line 16: Everywhere else in the code params is singular of dagParams

integration/64_incoming_connections_test.go
someone235, line 15: Please mention that it's leveldb file descriptors

integration/64_incoming_connections_test.go
```diff
@@ -0,0 +1,55 @@
...
+	blockAddedWG := sync.WaitGroup{}
+	blockAddedWG.Add(numBullies)
+	for _, bully := range bullies {
+		err := bully.rpcClient.NotifyBlocks()
+		if err != nil {
+			t.Fatalf("Error from NotifyBlocks: %+v", err)
+		}
+
+		bully.rpcClient.onBlockAdded = func(header *wire.BlockHeader) {
+			blockAddedWG.Done()
```
someone235, line 44: Can you check that it's not called twice?

PR #843, opened 2020-08-04
[NOD-1226] Properly handle ErrIsClosed during handshake

svarogg:


protocol/flows/handshake/handshake.go
```diff
@@ -93,3 +95,20 @@ func HandleHandshake(context HandleHandshakeContext, netConnection *netadapter.N
 
 	return peer, nil
 }
+
+// Handshake is different from other flows, since in it should consider ErrRouteClosed as ProtocolError
+// Therefore we implement a separate handleError for handshake
+func handleError(err error, flowName string, isStopping *uint32, errChan chan error) {
+	if errors.Is(err, router.ErrRouteClosed) {
+		errChan <- protocolerrors.Wrap(false, err, "Disconnected on handshake")
```
someone235, line 103: You forgot to return

PR #851, opened 2020-08-06
[NOD-1221] Explicitly add a maximum message size in gRPC

stasatdaglabs:


netadapter/server/grpcserver/grpc_server.go
```diff
@@ -21,11 +21,13 @@ type gRPCServer struct {
 	server             *grpc.Server
 }
 
+const maxReceiveMessageSize = 1024 * 1024 * 5 // 5MB
```
svarogg, line 24: Didn't we agree on 10 MB?

netadapter/server/grpcserver/grpc_server.go
```diff
@@ -21,11 +21,13 @@ type gRPCServer struct {
...
+
 // NewGRPCServer creates and starts a gRPC server, listening on the
 // provided addresses/ports
 func NewGRPCServer(listeningAddrs []string) (server.Server, error) {
 	s := &gRPCServer{
-		server:         grpc.NewServer(),
+		server:         grpc.NewServer(grpc.MaxRecvMsgSize(maxReceiveMessageSize)),
```
svarogg, line 30: I think will also want to limit send size

PR #854, opened 2020-08-09
[NOD-1204] Add timestamp and message number to domain messages

someone235:


domainmessage/base_message.go
```diff
@@ -0,0 +1,24 @@
+package domainmessage
+
+import "time"
+
+type baseMessage struct {
+	messageNumber uint64
+	receivedAt    time.Time
```
stasatdaglabs, line 7: receivedAt is not used anywhere. Please remove it.

PR #858, opened 2020-08-10
[NOD-1190] Refactor process.go

stasatdaglabs:


blockdag/process.go
```diff
@@ -153,60 +30,82 @@ func (dag *BlockDAG) ProcessBlock(block *util.Block, flags BehaviorFlags) (isOrp
...
 
-	if !isAfterDelay {
-		// Perform preliminary sanity checks on the block and its transactions.
-		delay, err := dag.checkBlockSanity(block, flags)
+	log.Debugf("Accepted block %s", blockHash)
+
+	return false, false, nil
+}
+
+func (dag *BlockDAG) processBlockCheckDelayedAndOrphanBlocks(block *util.Block, flags BehaviorFlags) (isOrphan bool, isDelayed bool, err error) {
```
svarogg, line 66: Why the `processBlock` prefix here and in `processBlockCheckBlockDelay` and `processBlockCheckMissingParents`?

blockdag/process.go
```diff
@@ -219,69 +118,339 @@ func (dag *BlockDAG) processBlockNoLock(block *util.Block, flags BehaviorFlags)
...
+// performs several validation checks which depend on its position within
+// the block DAG before adding it. The block is expected to have already
+// gone through ProcessBlock before calling this function with it.
+//
+// The flags are also passed to checkBlockContext and connectBlock. See
+// their documentation for how the flags modify their behavior.
+//
+// This function MUST be called with the dagLock held (for writes).
+func (dag *BlockDAG) maybeAcceptBlock(block *util.Block, flags BehaviorFlags) error {
+	err := dag.checkBlockContext(block, flags)
```
svarogg, line 157: I'm not 100% convinced that checkBlockContext should be in maybeAcceptBlock, and not in processBlock right before it.

TBH the distinction between processBlock and maybeAcceptBlock is not very clear.
One possible distinction is that maybeAcceptBlock is handling all the blockNode stuff. Therefore checkBlockContext should be outside.

LMK what you think.

blockdag/process.go
```diff
@@ -219,69 +118,339 @@ func (dag *BlockDAG) processBlockNoLock(block *util.Block, flags BehaviorFlags)
...
+	if err != nil {
+		return dag.handleConnectBlockError(err, newNode)
+	}
+
+	dag.notifyBlockAccepted(block, chainUpdates, flags)
+
+	return nil
+}
+
+func (dag *BlockDAG) createNewBlockNode(block *util.Block) (
```
svarogg, line 177: `createNewBlockNode` is somewhat confusing vs dag.newBlockNode. Not entirely clear what's the difference.

I couldn't think of a better name though, so if you can't either, simply add comment explaining what exactly does this function do... something along the lines of
```
// createNewBlockNode generates a new block node for given block and stores it in the block index....
```

blockdag/validate.go
```diff
@@ -735,6 +733,33 @@ func (dag *BlockDAG) checkBlockContext(block *util.Block, parents blockSet, flag
...
+			if err != nil {
+				return err
+			}
+		}
+		return err
+	}
+	return nil
+}
+
+func (dag *BlockDAG) checkBlockTxsFinalized(block *util.Block, node *blockNode, flags BehaviorFlags) error {
```
svarogg, line 751: From you I would expect it to be called checkBlockTreansactionsFinalized

blockdag/validate.go
```diff
@@ -717,11 +706,20 @@ func (dag *BlockDAG) validateParents(blockHeader *domainmessage.BlockHeader, par
 // for how the flags modify its behavior.
 //
 // This function MUST be called with the dag state lock held (for writes).
-func (dag *BlockDAG) checkBlockContext(block *util.Block, parents blockSet, flags BehaviorFlags) error {
+func (dag *BlockDAG) checkBlockContext(block *util.Block, flags BehaviorFlags) error {
+	err := dag.checkBlockParents(block)
+	if err != nil {
+		return err
+	}
+	parents, err := lookupParentNodes(block, dag)
```
svarogg, line 714: You just did lookupParentNodes in checkBlockParents. 
Why not return it from there and use here?

blockdag/validate.go
```diff
@@ -735,6 +733,33 @@ func (dag *BlockDAG) checkBlockContext(block *util.Block, parents blockSet, flag
...
+	}
+	return nil
+}
+
+func (dag *BlockDAG) checkBlockTxsFinalized(block *util.Block, node *blockNode, flags BehaviorFlags) error {
+	fastAdd := flags&BFFastAdd == BFFastAdd || dag.index.NodeStatus(node).KnownValid()
+	if fastAdd {
+		return nil
+	}
+	parents, err := lookupParentNodes(block, dag)
```
svarogg, line 756: Why not node.parents? Or actually, you do parents.bluest() a couple line below - instead we should probably use node.selectedParent (Please check me on this one)

blockdag/block_utxo.go
```diff
@@ -0,0 +1,305 @@
...
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+
+	multiset, err = node.calcMultiset(dag, txsAcceptanceData, selectedParentPastUTXO)
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+
+	calculatedMultisetHash := daghash.Hash(*multiset.Finalize())
```
svarogg, line 108: It could be nice if this block turned into a function as well

blockdag/process.go
svarogg: processBlockCheckMissingParents doesn't have a corresponding something in validate.go, so I think it can be safely renamed to checkMissingParents.

So we are left with processBlockCheckBlockDelay and processBlockCheckDelayedAndOrphanBlocks the can be confused with checkBlockDelayed

I suggest to rename checkBlockDelayed to shouldBlockBeDelayed.
Then remove prefix from processBlockCheckBlockDelay and processBlockCheckDelayedAndOrphanBlocks

blockdag/orphaned_blocks.go
```diff
@@ -0,0 +1,235 @@
+package blockdag
```
someone235, line 1: Maybe change to `orphan_blocks`?

blockdag/process.go
```diff
@@ -153,60 +30,82 @@ func (dag *BlockDAG) ProcessBlock(block *util.Block, flags BehaviorFlags) (isOrp
...
-		if delay != 0 {
-			err = dag.addDelayedBlock(block, delay)
-			if err != nil {
-				return false, false, err
-			}
-			return false, true, nil
+func (dag *BlockDAG) processBlockCheckBlockDelay(block *util.Block, flags BehaviorFlags) (isDelayed bool, err error) {
+	delay, isDelayed := dag.checkBlockDelayed(block)
+	if isDelayed && isBehaviorFlagRaised(flags, BFDisallowDelay) {
+		str := fmt.Sprintf("Cannot process blocks beyond the "+
```
someone235, line 82: use small case

blockdag/process.go
```diff
@@ -153,60 +30,82 @@ func (dag *BlockDAG) ProcessBlock(block *util.Block, flags BehaviorFlags) (isOrp
...
 	for _, parentHash := range block.MsgBlock().Header.ParentHashes {
 		if !dag.IsInDAG(parentHash) {
 			missingParents = append(missingParents, parentHash)
 		}
 	}
-	if len(missingParents) > 0 && disallowOrphans {
-		str := fmt.Sprintf("Cannot process orphan blocks while the BFDisallowOrphans flag is raised %s", blockHash)
+
+	if len(missingParents) > 0 && isBehaviorFlagRaised(flags, BFDisallowOrphans) {
+		str := fmt.Sprintf("Cannot process orphan blocks while the "+
```
someone235, line 107: use small case

blockdag/process.go
someone235: I would call it createBlockNodeFromBlock

blockdag/process_test.go
```diff
@@ -103,6 +104,7 @@ func TestProcessDelayedBlocks(t *testing.T) {
 
 	blockDelay := time.Duration(dag1.Params.TimestampDeviationTolerance)*dag1.Params.TargetTimePerBlock + 5*time.Second
 	delayedBlock.Header.Timestamp = initialTime.Add(blockDelay)
+	delayedBlock.Header.Nonce = 1
```
someone235, line 107: Why?

stasatdaglabs: Even if you pass BFNoPOW to ProcessBlock, PoW is always checked in the un-delaying process.

blockdag/process.go
stasatdaglabs: The only reason maybeAcceptBlock still exists is because it's used inside of processOrphans.
Ideally processOrphans would use processBlock (as processDelayedBlocks does, for example), but that would require a change that's outside the scope of this ticket.

blockdag/validate.go
```diff
@@ -707,11 +769,7 @@ func (dag *BlockDAG) validateParents(blockHeader *domainmessage.BlockHeader, par
 //
 // This function MUST be called with the dag state lock held (for writes).
 func (dag *BlockDAG) checkBlockContext(block *util.Block, flags BehaviorFlags) error {
-	err := dag.checkBlockParents(block)
-	if err != nil {
-		return err
-	}
-	parents, err := lookupParentNodes(block, dag)
+	parents, err := dag.checkBlockParents(block)
```
someone235, line 772: Can you calculate the parents outside and give it as an argument?

blockdag/process_test.go
```diff
@@ -104,7 +104,7 @@ func TestProcessDelayedBlocks(t *testing.T) {
 
 	blockDelay := time.Duration(dag1.Params.TimestampDeviationTolerance)*dag1.Params.TargetTimePerBlock + 5*time.Second
 	delayedBlock.Header.Timestamp = initialTime.Add(blockDelay)
-	delayedBlock.Header.Nonce = 1
+	delayedBlock.Header.Nonce = 2
```
someone235, line 107: Please add a comment to explain why you set nonce

blockdag/process_test.go
someone235: Open a ticket to fix it

PR #860, opened 2020-08-11
[NOD-1233] Go over all TODO(libp2p)s and either fix them or create tickets for them

stasatdaglabs:


rpc/handle_get_net_totals.go
```diff
@@ -7,7 +7,6 @@ import (
 
 // handleGetNetTotals implements the getNetTotals command.
 func handleGetNetTotals(s *Server, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
-	// TODO(libp2p): fill this up with real values
 	reply := &model.GetNetTotalsResult{
```
someone235, line 10: Can you please write a comment that explains the values here are not real?

PR #862, opened 2020-08-11
[NOD-1257] Disable difficulty adjustment on simnet

someone235:


dagconfig/params.go
```diff
@@ -394,6 +397,8 @@ var SimnetParams = Params{
 
 	// EnableNonNativeSubnetworks enables non-native/coinbase transactions
 	EnableNonNativeSubnetworks: false,
+
+	DisableDifficultyAdjustment: true,
```
stasatdaglabs, line 401: Please set this explicitly in all dag params.

PR #869, opened 2020-08-13
[NOD-1249] Add pruning related constants

someone235:


domain/blockdag/finality.go
```diff
@@ -18,6 +18,12 @@ func (dag *BlockDAG) FinalityInterval() uint64 {
 	return uint64(dag.Params.FinalityDuration / dag.Params.TargetTimePerBlock)
 }
 
+// pruningDepth is used to determine the depth from the virtual block where the pruning point is set once updated.
```
stasatdaglabs, line 21: Please explain the equation in the comment.

domain/blockdag/finality.go
```diff
@@ -18,6 +18,12 @@ func (dag *BlockDAG) FinalityInterval() uint64 {
...
+func (dag *BlockDAG) pruningDepth() uint64 {
+	k := uint64(dag.Params.K)
+	return 2*dag.FinalityInterval() + 4*mergeSetSizeLimit*k + 2*k + 2
```
svarogg, line 24: Please add unit test that makes sure that this equation evaluates to what we expect it to evaluate in all standard nets.

PR #873, opened 2020-08-17
[NOD-1275] Fix onNewBlock not being called from RPC submitBlock

stasatdaglabs:


network/protocol/flowcontext/blocks.go
```diff
@@ -39,8 +39,10 @@ func (f *FlowContext) broadcastTransactionsAfterBlockAdded(block *util.Block, tr
 	for i, tx := range transactionsAcceptedToMempool {
 		txIDsToBroadcast[i] = tx.ID()
 	}
-
-	copy(txIDsToBroadcast[len(transactionsAcceptedToMempool):], txIDsToBroadcast)
+	offset := len(transactionsAcceptedToMempool)
+	for i, tx := range txIDsToRebroadcast {
```
someone235, line 43: tx -> txID

PR #903, opened 2020-08-27
[NOD-1151] Update DNSSeeder to use grpc

oudeis:


someone235 (changes requested):

app/app.go
```diff
@@ -144,15 +144,15 @@ func New(cfg *config.Config, databaseContext *dbaccess.DatabaseContext, interrup
 
 func (a *App) maybeSeedFromDNS() {
 	if !a.cfg.DisableDNSSeed {
-		if a.cfg.GRPCSeed != "" {
+		switch {
+		case a.cfg.GRPCSeed != "" && a.cfg.DNSSeed != "":
```
someone235, line 148: Can you move this check to config/config.go where we validate command line arguments?

someone235 (changes requested):

PR #918, opened 2020-09-07
[NOD-1313] Refactor AddressManager

sept-en:


infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -5,1264 +5,254 @@
...
-		if oldest.referenceCount == 0 {
-			am.decrementNewAddressCount(subnetworkID)
-			delete(am.addressIndex, addressKey)
-		}
-	}
+// NetAddressKey returns a key of the ip address to use it in maps.
+func netAddressKey(netAddress *appmessage.NetAddress) AddressKey {
+	key := make([]byte, len(netAddress.IP), len(netAddress.IP)+2)
+	copy(key, netAddress.IP)
+	return AddressKey(append(key, byte(netAddress.Port), byte(netAddress.Port>>8)))
```
svarogg, line 73: I don't like the bit-magic happening with the port number.
Better use `binary.LittleEndian.PutUint16` instead.

infrastructure/network/randomaddress/address_randomize.go
```diff
@@ -0,0 +1,43 @@
+package randomaddress
```
svarogg, line 1: I might be missing something, but I don't understand why do we need a separate package for RandomAddress?

Note that later down the road, most probably the address selection algorithm will get more complex, including managing multiple address lists within the address manager (similar or not to the New/Tried distinction + buckets in the legacy address manager).

Therefore a generic object that gets an address list and returns a random address out of it will hurt our extensibility rather then helping it.

infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -5,1264 +5,254 @@
...
+		addresses: map[AddressKey]*netAddressWrapper{},
+		random:    randomaddress.NewAddressRandomize(),
+		cfg:       cfg,
+	}, nil
 }
 
-// PeersStateForSerialization returns the data model that is used to serialize the peers state to any encoding.
-func (am *AddressManager) PeersStateForSerialization() (*PeersStateForSerialization, error) {
+// AddAddresses adds addresses to the address manager
+func (am *AddressManager) AddAddresses(addresses ...*appmessage.NetAddress) {
```
svarogg, line 97: Please add the `AddAddress` function as specified. Then use it inside the loop of AddAddresses.

Note: You will have to have two version of AddAddress: one that locks am.mutex (for the exported AddAddress) and another one that doesn't lock `am.mutex` (for usage inside AddAddresses loop).
In such cases we usually name the non-locking function `addAddressNoLock`.

infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -5,1264 +5,254 @@
...
-		for i := range am.subnetworkTriedAddresBucketArrays[subnetworkID] {
-			peersState.SubnetworkTriedAddressBucketArrays[subnetworkIDStr][i] = make([]AddressKey, len(am.subnetworkTriedAddresBucketArrays[subnetworkID][i]))
-			j := 0
-			for _, knownAddress := range am.subnetworkTriedAddresBucketArrays[subnetworkID][i] {
-				peersState.SubnetworkTriedAddressBucketArrays[subnetworkIDStr][i][j] = NetAddressKey(knownAddress.netAddress)
-				j++
+		key := netAddressKey(address)
+		if _, ok := am.addresses[key]; !ok {
+			am.addresses[key] = &netAddressWrapper{
+				netAddress: address,
```
svarogg, line 109: Please set IsBanned and IsLocal (in case you keep it) explicitly.

infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -5,1264 +5,254 @@
...
-		am.subnetworkTriedAddresBucketArrays[*knownAddress.subnetworkID] = &triedAddressBucketArray{}
-		for i := range am.subnetworkTriedAddresBucketArrays[*knownAddress.subnetworkID] {
-			am.subnetworkTriedAddresBucketArrays[*knownAddress.subnetworkID][i] = nil
-		}
-	}
-	am.subnetworkTriedAddresBucketArrays[*knownAddress.subnetworkID][bucketIndex] = append(am.subnetworkTriedAddresBucketArrays[*knownAddress.subnetworkID][bucketIndex], knownAddress)
+type netAddressWrapper struct {
+	netAddress *appmessage.NetAddress
+	isBanned   bool
+	isLocal    bool
```
svarogg, line 58: I think it makes more sense to keep two lists: one for local, one for other addresses.
They are never used interchageably, and it will save us quite a lot of looping inside BestLocalAddress

infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -5,1264 +5,254 @@
...
-	}
-
-	return peersState, nil
 }
 
-// loadPeers loads the known address from the database. If missing,
-// just don't load anything and start fresh.
-func (am *AddressManager) loadPeers() error {
+// AddLocalAddresses adds local netAddresses to the address manager
+func (am *AddressManager) AddLocalAddresses(addresses ...*appmessage.NetAddress) {
```
svarogg, line 116: During the refactor, we have lost the functionality that was calling `AddLocalAddress` (under addressManager.initListeners), this functionality should be restored.

infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -5,1264 +5,254 @@
...
-}
-
-// numAddresses returns the number of addresses that belongs to a specific subnetwork id
-// which are known to the address manager.
-func (am *AddressManager) numAddresses(subnetworkID *subnetworkid.SubnetworkID) int {
-	if subnetworkID == nil {
-		return am.fullNodeNewAddressCount + am.fullNodeTriedAddressCount
+	result := make([]*appmessage.NetAddress, 0, len(am.addresses))
+	for key, address := range am.addresses {
+		if !address.isBanned && !exceptionsKeys[key] {
```
svarogg, line 156: It might make sense to keep all the banned addresses in a separate list as well, instead of filtering them out every time we need an address

infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -5,1264 +5,254 @@
...
-	var bucketArray addressBucketArray
-	if triedAddressCount > 0 && (newAddressCount == 0 || am.random.Intn(2) == 0) {
-		bucketArray = triedAddressBucketArray
-	} else if newAddressCount > 0 {
-		bucketArray = newAddressBucketArray
-	} else {
-		// There aren't any addresses in any of the buckets
+	key := netAddressKey(address)
+	if address, ok := am.addresses[key]; ok {
+		address.isBanned = true
```
svarogg, line 216: It is idiomatic in go to have the error-case inside the if.
Among other things, the compiler uses this assumption to do branch-prediction.

This means you'll have to declare `var address` outside the if, but it's still preferable.

(Same for Unban, IsBanned)

infrastructure/network/addressmanager/network.go
```diff
@@ -77,6 +78,37 @@ var (
 	heNet = ipNet("2001:470::", 32, 128)
 )
 
+const (
+	// TriedBucketCount is the number of buckets we split tried
+	// addresses over.
+	TriedBucketCount = 64
```
svarogg, line 84: It seems like all these consts are now redundant (after KnownAddress is deleted)

svarogg (changes requested):

oudeis:

infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -5,1585 +5,225 @@
...
-	Addresses []*serializedKnownAddress
-
-	SubnetworkNewAddressBucketArrays   map[string]*serializedNewAddressBucketArray // string is Subnetwork ID
-	FullNodeNewAddressBucketArray      serializedNewAddressBucketArray
-	SubnetworkTriedAddressBucketArrays map[string]*serializedTriedAddressBucketArray // string is Subnetwork ID
-	FullNodeTriedAddressBucketArray    serializedTriedAddressBucketArray
-}
-
-type localAddress struct {
+type addressEntry struct {
```
svarogg, line 21: If we only have netAddress inside addressEntry, do we really need a especial struct for it? 
Why not store appmessage.NetAddress directly?

infrastructure/network/connmanager/outgoing_connections.go
```diff
@@ -24,46 +28,26 @@ func (c *ConnectionManager) checkOutgoingConnections(connSet connectionSet) {
 
 	connectionsNeededCount := c.targetOutgoing - len(c.activeOutgoing)
 	connectionAttempts := connectionsNeededCount * 2
```
svarogg, line 30: We no longer need the `* 2` here.
We needed it before, because much of the connectionsAttempts were discovered as existing/banned.
Since RandomAddresses no longer returns existing/banned addresses, we should be fine.

infrastructure/network/connmanager/outgoing_connections.go
```diff
@@ -24,46 +28,26 @@ func (c *ConnectionManager) checkOutgoingConnections(connSet connectionSet) {
...
-	for i := 0; i < connectionAttempts; i++ {
+	netAddresses := c.addressManager.RandomAddresses(connectionAttempts, connectedAddresses)
```
svarogg, line 31: When you get to call RandomAddresses, you have to make sure that connectedAddresses also includes the incoming addresses.

infrastructure/network/connmanager/outgoing_connections.go
```diff
@@ -24,46 +28,26 @@ func (c *ConnectionManager) checkOutgoingConnections(connSet connectionSet) {
...
+
+	for _, netAddress := range netAddresses {
 		// Return in case we've already reached or surpassed our target
 		if len(c.activeOutgoing) >= c.targetOutgoing {
```
svarogg, line 35: Since we no longer do the `* 2` (see separate comment above), we no longer need this clause.

infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -5,1585 +5,225 @@
...
-		// Send something unroutable if nothing suitable.
-		var ip net.IP
-		if !IsIPv4(remoteAddress) {
-			ip = net.IPv6zero
-		} else {
-			ip = net.IPv4zero
-		}
-		services := appmessage.SFNodeNetwork | appmessage.SFNodeBloom
-		bestAddress = appmessage.NewNetAddressIPPort(ip, 0, services)
+	if ok {
```
svarogg, line 201: As previously mentioned - it's idiomatic in go to have the error-case in the if.
So please, invert the condition here.

svarogg (changes requested):

app/rpc/rpchandlers/get_peer_addresses.go
```diff
@@ -11,9 +11,11 @@ import (
 
 // HandleGetPeerAddresses handles the respectively named RPC command
 func HandleGetPeerAddresses(context *rpccontext.Context, _ *router.Router, _ appmessage.Message) (appmessage.Message, error) {
-	netAaddresses := context.AddressManager.Addresses()
-	addresses := make([]*appmessage.GetPeerAddressesKnownAddressMessage, len(netAaddresses))
-	for i, netAddress := range netAaddresses {
+	netAddresses := context.AddressManager.Addresses()
+	netAddresses = append(netAddresses, context.AddressManager.BannedAddresses()...)
```
svarogg, line 15: As previously mentioned, the RPC needs to return the banned addresses in a separate list.
Please add new field in the RPC response for "Banned Addresses", and place the output of `context.AddressManager.BannedAddresses` in it.

svarogg (changes requested):

svarogg (dismissed):

svarogg (approved):

PR #946, opened 2020-10-05
Add more tests for Hash and for SubnetworkID

elichai:
Hi,
The motivation here is that even after #912 it didn't catch when I made a mistake modifying the ordering, so I added even more tests that explicitly try to test endian ordering

I also copied all of Hash's tests to subnetworkID and modified them to work because it had no tests and almost the same interface

### DIFF
--- a/util/subnetworkid/subnetworkid_test.go
+++ b/util/subnetworkid/subnetworkid_test.go
@@ -1,0 +1,441 @@
+// Copyright (c) 2013-2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package subnetworkid
+
+import (
+	"bytes"
+	"encoding/hex"
+	"errors"
+	"math/big"
+	"math/rand"
+	"reflect"
+	"sort"
+	"testing"
+)
+
+// TestSubnetworkID tests the SubnetworkID API.
+func TestSubnetworkID(t *testing.T) {
+	subnetworkIDStr := "a3eb3f82edc878cea25ec41d6b790744e5daeef"
+	subnetworkID, err := NewFromStr(subnetworkIDStr)
+	if err != nil {
+		t.Errorf("NewFromStr: %v", err)
+	}
+	subnetworkID2Bytes := []byte{
+		0x79, 0xa6, 0x1a, 0xdb, 0xc6,
+		0xe5, 0xa2, 0xe1, 0x39, 0xd2,
+		0x71, 0x3a, 0x54, 0x6e, 0xc7,
+		0xc8, 0x75, 0x63, 0x2e, 0x75,
+	}
+
+	subnetworkID2, err := New(subnetworkID2Bytes)
+	if err != nil {
+		t.Errorf("New: unexpected error %v", err)
+	}
+
+	// Ensure proper size.
+	if len(subnetworkID2) != IDLength {
+		t.Errorf("New: SubnetworkID length mismatch - got: %v, want: %v",
+			len(subnetworkID2), IDLength)
+	}
+
+	// Ensure contents match.
+	if !bytes.Equal(subnetworkID2[:], subnetworkID2Bytes) {
+		t.Errorf("New: SubnetworkID contents mismatch - got: %v, want: %v",
+			subnetworkID2[:], subnetworkID2Bytes)
+	}
+
+	if subnetworkID2.IsEqual(subnetworkID) {
+		t.Errorf("IsEqual: SubnetworkID contents should not match - got: %v, want: %v",
+			subnetworkID2, subnetworkID)
+	}
+
+	// Set SubnetworkID from byte slice and ensure contents match.
+	err = subnetworkID2.SetBytes(subnetworkID.CloneBytes())
+	if err != nil {
+		t.Errorf("SetBytes: %v", err)
+	}
+	if !subnetworkID2.IsEqual(subnetworkID) {
+		t.Errorf("IsEqual: SubnetworkID contents mismatch - got: %v, want: %v",
+			subnetworkID2, subnetworkID)
+	}
+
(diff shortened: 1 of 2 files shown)


util/subnetworkid/subnetworkid_test.go
```diff
@@ -0,0 +1,443 @@
...
+	})
+
+	SubnetworkIDStr := SubnetworkID.String()
+	if SubnetworkIDStr != wantStr {
+		t.Errorf("String: wrong SubnetworkID string - got %v, want %v",
+			SubnetworkIDStr, wantStr)
+	}
+}
+
+func TestSubnetworkIDesStrings(t *testing.T) {
```
svarogg, line 106: typo: TestSubnetworkID**e**sStrings

util/subnetworkid/subnetworkid_test.go
```diff
@@ -0,0 +1,443 @@
...
+		}
+	}
+}
+
+// TestNewFromStr executes tests against the NewFromStr function.
+func TestNewFromStr(t *testing.T) {
+	tests := []struct {
+		in   string
+		want SubnetworkID
+		err  error
```
svarogg, line 142: Please rename to expectedErr/wantErr or something that makes it clear this is what we expect

util/subnetworkid/subnetworkid_test.go
```diff
@@ -0,0 +1,443 @@
...
+			SubnetworkID([IDLength]byte{
+				0x01, 0x00, 0x00, 0x00, 0x00,
+				0x00, 0x00, 0x00, 0x00, 0x00,
+				0x00, 0x00, 0x00, 0x00, 0x00,
+				0x00, 0x00, 0x00, 0x00, 0x00,
+			}),
+			nil,
+		},
+
+		// Block 203707 with stripped leading zeros.
```
svarogg, line 164: Again, please no mentions of specific blocks.

util/subnetworkid/subnetworkid_test.go
```diff
@@ -0,0 +1,443 @@
...
+			t.Errorf(unexpectedResultStr, i, result, &test.want)
+			continue
+		}
+	}
+}
+
+// TestAreEqual executes tests against the AreEqual function.
+func TestAreEqual(t *testing.T) {
+	SubnetworkID0, _ := NewFromStr("0000000000000000000000000000000000000000")
+	SubnetworkID1, _ := NewFromStr("1111111111111111111111111111111111111111")
```
svarogg, line 213: In many contexts SubnetworkID 1 (a.k.a. coinbase) is 0000000000000000000000000000000000000001.
So I'd rather, to avoid confusion, if the value was the above (here and in all subnetworkIDN's), or you rename to something like `SubnetworkIDAll1s`

util/subnetworkid/subnetworkid_test.go
```diff
@@ -0,0 +1,443 @@
...
+	SubnetworkID2, _ := NewFromStr("2222222222222222222222222222222222222222")
+	SubnetworkID3, _ := NewFromStr("3333333333333333333333333333333333333333")
+	SubnetworkIDes0To2 := []SubnetworkID{*SubnetworkID0, *SubnetworkID1, *SubnetworkID2}
```
svarogg, line 216: So it seems like this was not a typo, but rather systematic - we pluralize ID as IDs, not IDes. Please comply

util/subnetworkid/subnetworkid_test.go
```diff
@@ -0,0 +1,443 @@
...
+		})
+
+		if !reflect.DeepEqual(test.SubnetworkIDes, test.expected) {
+			t.Errorf("unexpected Sort result for"+
+				" test \"%s\". Expected: %v, got: %v.", test.name, test.expected, test.SubnetworkIDes)
+		}
+	}
+}
+
+func SubnetworkIDFlipBit(SubnetworkID SubnetworkID, bit int) SubnetworkID {
```
svarogg, line 389: Name of agument SubnetworkID should not be capitalized

util/subnetworkid/subnetworkid_test.go
```diff
@@ -0,0 +1,443 @@
...
+		}
+	}
+}
+
+func BenchmarkSubnetworkID_Cmp(b *testing.B) {
+	SubnetworkID0, err := NewFromStr("3333333333333333333333333333333333333333")
+	if err != nil {
+		b.Fatal(err)
+	}
+	for n := 0; n < b.N; n++ {
```
svarogg, line 430: It is negligible most of the time, but I still prefer to b.ResetTimer() between the setup anf the benchmark loop

svarogg (changes requested):

elichai:
@svarogg Done https://github.com/kaspanet/kaspad/compare/a88278f08a47c79c278363c46fda3c16d9d859d1..315418d8a90cb889b9a9e6ffbc5add29f4ee9830

util/subnetworkid/subnetworkid_test.go
```diff
@@ -0,0 +1,441 @@
...
+)
+
+// TestSubnetworkID tests the SubnetworkID API.
+func TestSubnetworkID(t *testing.T) {
+	subnetworkIDStr := "a3eb3f82edc878cea25ec41d6b790744e5daeef"
+	subnetworkID, err := NewFromStr(subnetworkIDStr)
+	if err != nil {
+		t.Errorf("NewFromStr: %v", err)
+	}
+	SubnetworkIDAll2Bytes := []byte{
```
svarogg, line 25: This is not All2 as far as I can tell

svarogg (changes requested):

util/subnetworkid/subnetworkid_test.go
elichai: :)

oudeis:

PR #949, opened 2020-10-11
[NOD-1413] Decouple the model package from everything

stasatdaglabs:


domain/consensus/model/blockghostdagdata.go
```diff
@@ -1,17 +1,15 @@
...
 
-import "github.com/kaspanet/kaspad/util/daghash"
-
 // BlockGHOSTDAGData represents GHOSTDAG data for some block
 type BlockGHOSTDAGData struct {
 	blueScore      uint64
-	selectedParent *daghash.Hash
-	mergeSetBlues  []*daghash.Hash
-	mergeSetReds   []*daghash.Hash
+	selectedParent *DomainHash
```
someone235, line 6: Please remove methods and export the members

domain/consensus/model/time.go
```diff
@@ -0,0 +1,26 @@
...
+
+import "time"
+
+const (
+	nanosecondsInAMillisecond = int64(time.Millisecond / time.Nanosecond)
+)
+
+// DomainTime is domain representation of time.Time that guarantees that all
+// of its methods will return times with millisecond precision.
+type DomainTime struct {
```
someone235, line 11: Why do we have a struct with methods on model?

someone235 (changes requested):

someone235 (approved):

PR #954, opened 2020-10-13
[NOD-1413] Make some additional interface changes

stasatdaglabs:


domain/consensus/model/interface_processes_consensusstatemanager.go
```diff
@@ -3,6 +3,11 @@ package model
 // ConsensusStateManager manages the node's consensus state
 type ConsensusStateManager interface {
 	UTXOByOutpoint(outpoint *DomainOutpoint) *UTXOEntry
-	ValidateTransaction(transaction *DomainTransaction, utxoEntries []*UTXOEntry) error
-	CalculateConsensusStateChanges(block *DomainBlock) *ConsensusStateChanges
+	CalculateConsensusStateChanges(block *DomainBlock, parents []*DomainHash,
+		transactions []*DomainTransaction, isDisqualified bool) (stateChanges *ConsensusStateChanges,
```
svarogg, line 7: Why does it get both block and parents+transactions - don't block already contain those?

domain/consensus/model/interface_processes_transactionalidator.go
```diff
@@ -0,0 +1,10 @@
+package model
+
+// TransactionValidator exposes a set of validation classes, after which
+// it's possible to determine whether a transaction is valid
+type TransactionValidator interface {
+	ValidateTransactionInIsolation(transaction *DomainTransaction) error
+	ValidateTransactionInContext(transaction *DomainTransaction) error
+	ValidateTransactionAndCalculateFee(transaction *DomainTransaction, utxoEntries []*UTXOEntry) (fee uint64, err error)
+	ValidateTransactionAgainstUTXO(transaction *DomainTransaction) error
```
svarogg, line 9: This is not the way I envisioned these functions. See submitTransaction graph in EA, I think it explains what I ment, and if not, lets talk about this.

domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -7,13 +7,13 @@ import (
...
-	pruningPointStore   model.PruningPointStore
+	pruningPointStore   model.PruningStore
 }
 
 // New instantiates a new PruningManager
 func New(
 	dagTraversalManager model.DAGTraversalManager,
-	pruningPointStore model.PruningPointStore) *PruningManager {
+	pruningPointStore model.PruningStore) *PruningManager {
 	return &PruningManager{
```
svarogg, line 17: Also needs access to DAGTopology, BlockStatuses and ConsensusStateManager

svarogg (commented):
Two more things that I don't have a line to comment on:
1. ConsensusStateStore should also have function UpdateTips (Or alternatively - update should also receive newTips).
2. We might move the blockExists() from blockIndex to blockStatuses, or blockStore, or something like that.
At the very least blockProcessor is going to need something like that.

svarogg (changes requested):
a

domain/consensus/model/interface_processes_transactionalidator.go
```diff
@@ -3,8 +3,5 @@ package model
 // TransactionValidator exposes a set of validation classes, after which
```
svarogg, line 3: The filename has a typo: transaction**a**lidator

svarogg (approved):

PR #958, opened 2020-10-19
[NOD-1462] Simplify consensus external API

stasatdaglabs:


domain/consensus/model/externalapi/coinbasedata.go
```diff
@@ -0,0 +1,8 @@
+package externalapi
+
+// CoinbaseData contains data by which a coinbase transaction
+// is built in a new block
```
someone235: why new?

someone235 (approved):

PR #965, opened 2020-10-25
[NOD-1476] Make further design changes

stasatdaglabs:


svarogg (changes requested):
Shove acceptanceManager back to CSM

svarogg (approved):

PR #967, opened 2020-10-26
[NOD-1477] Add selected parent to merge set

someone235:


domain/consensus/processes/ghostdagmanager/compare.go
```diff
@@ -45,26 +46,13 @@ func (gm *ghostdagManager) ChooseSelectedParent(blockHashA *externalapi.DomainHa
 	blockABlueScore := blockAGHOSTDAGData.BlueScore
 	blockBBlueScore := blockBGHOSTDAGData.BlueScore
 	if blockABlueScore == blockBBlueScore {
-		if hashesLess(blockHashA, blockHashB) {
-			return blockHashB, nil
+		if hashes.Less(blockHashA, blockHashB) {
+			return blockHashA, nil
 		}
-		return blockHashA, nil
+		return blockHashB, nil
```
elichai, line 52: I believe this is a consensus change and will break all the ghostDAG tests

elichai: Before we took the higher block and here we take the lower:
https://github.com/kaspanet/kaspad/blob/eaf911722584b8ce732306c4250b220b51499d0f/domain/blockdag/blockset.go#L108-L109

stasatdaglabs (approved):

elichai (approved):

PR #972, opened 2020-10-27
Add new ErrMissingTxOut  and ErrInvalidTransactionsInNewBlock errors

elichai:
Should I leave the UTXOEntries checks scattered across consensus or should we have a sanity check and later assume it is not nil?

### DIFF
--- a/domain/consensus/ruleerrors/rule_error.go
+++ b/domain/consensus/ruleerrors/rule_error.go
@@ -276,3 +276,26 @@ func NewErrMissingTxOut(missingOutpoints []externalapi.DomainOutpoint) error {
 		inner:   ErrMissingTxOut{missingOutpoints},
 	})
 }
+
+// ErrInvalidTransactionsInNewBlock indicates that some transactions in a new block are invalid
+type ErrInvalidTransactionsInNewBlock struct {
+	InvalidTransactions []struct {
+		*externalapi.DomainTransaction
+		error
+	}
+}
+
+func (e ErrInvalidTransactionsInNewBlock) Error() string {
+	return fmt.Sprint(e.InvalidTransactions)
+}
+
+// NewErrInvalidTransactionsInNewBlock Creates a new ErrInvalidTransactionsInNewBlock error wrapped in a RuleError
+func NewErrInvalidTransactionsInNewBlock(invalidTransactions []struct {
+	*externalapi.DomainTransaction
+	error
+}) error {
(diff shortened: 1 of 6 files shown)


domain/consensus/ruleerrors/rule_error.go
```diff
@@ -232,13 +234,43 @@ var (
...
+type ErrMissingTxOut struct {
+	MissingOutpoints []externalapi.DomainOutpoint
+}
+
+func (e ErrMissingTxOut) Error() string {
+	return fmt.Sprint(e.MissingOutpoints)
+}
+
+// WrapInRuleError wraps any error inside a RuleError
+func WrapInRuleError(err error) error {
```
stasatdaglabs, line 273: Instead of externalizing WrapInRuleError, I would've preferred a function called NewErrMissingTxOut that creates `ErrMissingTxOut`, wraps it in a `RuleError`, and returns it as an `error`.

stasatdaglabs (changes requested):

domain/consensus/ruleerrors/rule_error.go
elichai: And then we will have multiple of those per error? (on top of the `Error()` impls)

stasatdaglabs: You mean for other errors? If so, then yes. I think I prefer that over the possibility of someone accidentally returning a not-wrapped ErrXxx.

elichai: Done

domain/consensus/ruleerrors/rule_error.go
```diff
@@ -232,13 +234,68 @@ var (
...
+func NewErrMissingTxOut(missingOutpoints []externalapi.DomainOutpoint) error {
+	return errors.WithStack(RuleError{
+		message: "ErrMissingTxOut",
+		inner:   ErrMissingTxOut{missingOutpoints},
+	})
+}
+
+// ErrInvalidTransactionsInNewBlock indicates that some transactions in a new block are invalid
+type ErrInvalidTransactionsInNewBlock struct {
+	InvalidTransactions []struct {
```
stasatdaglabs, line 282: Please make a new type for this struct. Suggested name: InvalidTransaction.

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #991, opened 2020-11-02
[NOD-1497] Add additional methods to consensus' external API

stasatdaglabs:


domain/consensus/consensus.go
```diff
@@ -70,42 +79,69 @@ func validateTransactionInContextAndPopulateMassAndFeeVirtualBlockHash() *extern
...
-	panic("implement me")
+	blockInfo := &externalapi.BlockInfo{}
+
+	exists, err := s.blockStatusStore.Exists(s.databaseContext, blockHash)
+	if err != nil {
+		return nil, err
+	}
+	blockInfo.Exists = exists
+	if !exists {
+		return blockInfo, nil
```
svarogg, line 99: We should probably return something that is not an empty BlockInfo to signal that block doesn't exist (I'd say return nil, but you'll shoot me for such insolent behavior)

domain/consensus/consensus.go
```diff
@@ -71,41 +80,68 @@ func validateTransactionInContextAndPopulateMassAndFeeVirtualBlockHash() *extern
...
 func (s *consensus) GetVirtualSelectedParent() (*externalapi.DomainBlock, error) {
-	panic("implement me")
+	virtualGHOSTDAGData, err := s.ghostdagDataStore.Get(s.databaseContext, model.VirtualBlockHash)
+	if err != nil {
+		return nil, err
+	}
+	return s.GetBlock(virtualGHOSTDAGData.SelectedParent)
 }
 
 func (s *consensus) CreateBlockLocator(lowHigh, highHash *externalapi.DomainHash) (*externalapi.BlockLocator, error) {
```
svarogg, line 141: typo: lowHigh

domain/consensus/processes/syncmanager/syncmanager.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (s syncManager) GetHashesBetween(lowHigh, highHash *externalapi.DomainHash) ([]*externalapi.DomainHash, error) {
+	panic("implement me")
+}
+
+func (s syncManager) GetMissingBlockBodyHashes(highHash *externalapi.DomainHash) ([]*externalapi.DomainHash, error) {
+	panic("implement me")
+}
+
+func (s syncManager) CreateBlockLocator(lowHigh, highHash *externalapi.DomainHash) (*externalapi.BlockLocator, error) {
```
svarogg, line 27: lowHigh here too

svarogg (changes requested):

domain/consensus/consensus.go
stasatdaglabs: We do return blockInfo.Exists = false.

svarogg (approved):

PR #994, opened 2020-11-03
[NOD-1503] Pruning Manager

elichai:
This currently deletes P.Past, I think I also need to delete P.AC that aren't in V.Past, will check with research how it's best to do


domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -1,51 +1,226 @@
...
+			return nil, err
+		}
+
+	}
+	return pruningPoint, nil
 }
 
 // SerializedUTXOSet returns the serialized UTXO set of the
 // current pruning point
 func (pm *pruningManager) SerializedUTXOSet() ([]byte, error) {
```
stasatdaglabs, line 150: This function is no longer relevant, as anyone who needs the serialized UTXOSet can simply directly get it from the store. Please remove it.

domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -1,51 +1,226 @@
...
-	return nil, nil
+	return pm.pruningStore.PruningPointSerializedUTXOSet(pm.databaseContext)
+}
+
+func (pm *pruningManager) deletePastBlocks(pruningPoint *externalapi.DomainHash) error {
+	// Collect every node in highNode's past (including itself) but
```
stasatdaglabs, line 155: Please don't call blocks "nodes" anymore.

domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -1,51 +1,226 @@
...
+	pm.multiSetStore.Delete(blockHash)
+	pm.acceptanceDataStore.Delete(blockHash)
+	pm.blocksStore.Delete(blockHash)
+	pm.utxoDiffStore.Delete(blockHash)
+
+	pm.blockStatusStore.Stage(blockHash, externalapi.StatusHeaderOnly)
+	return false, nil
+}
+
+func pruningDepth(k, finalityInterval, mergeSetSizeLimit uint64) uint64 {
```
stasatdaglabs, line 216: Please move this function into dagconfig/params.go (next to FinalityDepth()) and pass it into PruningManager the same way you pass K.

domain/consensus/utils/hashserialization/utxo.go
```diff
@@ -49,25 +85,21 @@ func serializeOutpoint(w io.Writer, outpoint *externalapi.DomainOutpoint) error
...
-	buf := [8 + 1 + 8]byte{}
+	buf := [8 + 1 + 8 + 8]byte{}
 	// Encode the blueScore.
 	binary.LittleEndian.PutUint64(buf[:8], entry.BlockBlueScore)
 
 	buf[8] = serializeUTXOEntryFlags(entry)
 
 	binary.LittleEndian.PutUint64(buf[9:], entry.Amount)
 
+	binary.LittleEndian.PutUint64(buf[17:], uint64(len(entry.ScriptPublicKey)))
```
stasatdaglabs, line 96: This function has seen been changed in @svarogg's branch. Please make sure to align it with how he has done it.

stasatdaglabs (changes requested):

domain/consensus/datastructures/pruningstore/pruningstore.go
```diff
@@ -118,3 +118,7 @@ func (ps *pruningStore) deserializeUTXOSetBytes(dbPruningPointUTXOSetBytes []byt
 
 	return dbPruningPointUTXOSet.Bytes, nil
 }
+
+func (ps *pruningStore) HasPruningPoint(dbContext model.DBReader) bool {
+	panic("implement me")
```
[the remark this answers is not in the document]
elichai: Yes, sorry this is part of the problem with going back and forth between tasks, when I wrote it the rest of the Store was also not implemented

domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -1,51 +1,226 @@
...
+				return nil, err
+			}
+			serializedUtxo, err := serializeUTXOSetIterator(utxoGenesisIter)
+			if err != nil {
+				return nil, err
+			}
+			pm.pruningStore.Stage(genesis, serializedUtxo)
+			pruningPoint = genesis
+		} else {
+			return nil, err
```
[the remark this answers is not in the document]
elichai: I wanted to avoid 1 Database call, but you're right

domain/consensus/utils/hashserialization/utxo.go
elichai: It's aligned. it's just weird that this was the only place in the file that used `WriteElement` so I replaced it to be like the rest (which were also done for performance reasons)

stasatdaglabs (approved):

PR #1010, opened 2020-11-08
[NOD-1433] Write specified unit tests for GHOSTDAG.

talelbaz:


domain/consensus/processes/ghostdag2/ghostdagimpl.go
```diff
@@ -352,8 +354,8 @@ func (gh *ghostdagHelper) sortByBlueScore(arr []*externalapi.DomainHash) error {
 			err = error
 			return false
 		}
-		var isSmaller bool = false
-		if blockRight.BlueScore < blockLeft.BlueScore {
+		var isSmaller = false
+		if blockLeft.BlueScore < blockRight.BlueScore {
```
elichai, line 358: I think this is more readable:
```go
if blockLeft.BlueScore == blockRight.BlueScore {
    return ismoreHash(arr[j], arr[i])
} else {
    return blockLeft < blockRight
}
```

elichai: Also, maybe change `ismoreHash` to `isHashLess` and then the left/right sides would be consistent with the comparison about

elichai (dismissed):

elichai (approved):
ACK

PR #1013, opened 2020-11-09
[NOD-1518] Fix genesis block insertion and integration tests

stasatdaglabs:


domain/consensus/datastructures/consensusstatestore/utxo.go
```diff
@@ -28,14 +28,14 @@ func (c *consensusStateStore) StageVirtualUTXODiff(virtualUTXODiff *model.UTXODi
...
 func (c *consensusStateStore) commitVirtualUTXODiff(dbTx model.DBTransaction) error {
-	if c.stagedVirtualUTXOSet != nil {
-		return errors.New("cannot commit virtual UTXO diff while virtual UTXO set is staged")
-	}
-
 	if c.stagedVirtualUTXODiff == nil {
 		return nil
 	}
 
+	if c.stagedVirtualUTXOSet != nil {
```
svarogg, line 35: I'd rather if this check happens in the beginning of Commit.
Right now we check this twice (and the second one is redundant, because once we reach commitVirtualUTXOSet the UTXODiff has already been commited).

Even better - disallow staging if the other side is staged. This way we'll catch the error as it happens.

domain/consensus/datastructures/utxodiffstore/utxodiffstore.go
```diff
@@ -64,12 +64,16 @@ func (uds *utxoDiffStore) Commit(dbTx model.DBTransaction) error {
...
+		if utxoDiffChild != nil {
+			var err error
+			utxoDiffChildBytes, err = uds.serializeUTXODiffChild(utxoDiffChild)
+			if err != nil {
+				return err
+			}
 		}
 
-		err = dbTx.Put(uds.utxoDiffHashAsKey(&hash), utxoDiffChildBytes)
+		err := dbTx.Put(uds.utxoDiffHashAsKey(&hash), utxoDiffChildBytes)
```
svarogg, line 76: I think this line should also inside the if. 
(And then you don't need all the shenanigans with pre-declaring utxoDiffChildBytes and everything.

domain/consensus/processes/consensusstatemanager/add_block_to_virtual.go
```diff
@@ -40,7 +27,36 @@ func (csm *consensusStateManager) AddBlockToVirtual(blockHash *externalapi.Domai
...
+	if !isNextVirtualSelectedParent {
+		return nil
+	}
+
+	blockStatus, err := csm.resolveBlockStatus(blockHash)
+	if err != nil {
+		return err
+	}
+
+	if blockStatus == externalapi.StatusValid {
```
svarogg, line 45: Not critical if you disagree, but I'd rather:
1. rename this method to maybeResolveBlockStatus
2. move the `if statusValid { checkFinalityViolation}` out of it

domain/consensus/processes/consensusstatemanager/verify_and_build_utxo.go
```diff
@@ -109,6 +105,10 @@ func calculateAcceptedIDMerkleRoot(multiblockAcceptanceData model.AcceptanceData
 			consensusserialization.TransactionID(acceptedTransactions[j]))
 	})
 
+	if len(acceptedTransactions) == 0 {
```
svarogg, line 108: Maybe we want this logic in merkle.merkleRoot?

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager.go
```diff
@@ -85,7 +85,7 @@ func (dtm *dagTraversalManager) HighestChainBlockBelowBlueScore(highHash *extern
 			return blockHash, nil
 		}
 		blockHash = chainBlock.SelectedParent
-		chainBlock, err = dtm.ghostdagDataStore.Get(dtm.databaseContext, highHash)
+		chainBlock, err = dtm.ghostdagDataStore.Get(dtm.databaseContext, blockHash)
```
svarogg, line 88: I'd rather if blockHash was called currentBlockHash or just current or something

svarogg (changes requested):

svarogg (changes requested):

domain/consensus/processes/syncmanager/syncinfo.go
```diff
@@ -54,11 +52,15 @@ func (sm *syncManager) resolveSyncState() (externalapi.SyncState, error) {
...
+	headerTipsPruningPoint, err := sm.consensusStateManager.HeaderTipsPruningPoint()
+	if err != nil {
+		return 0, err
+	}
+	headerTipsPruningPointStatus, err := sm.blockStatusStore.Get(sm.databaseContext, headerTipsPruningPoint)
 	if err != nil {
 		return 0, err
 	}
-	if headerVirtualSelectedParentBlockStatus != externalapi.StatusValid {
+	if headerTipsPruningPointStatus != externalapi.StatusValid {
```
svarogg, line 63: Please add comment explaining that this can happen only if we are still in headers first.

domain/miningmanager/mempool/mempool.go
```diff
@@ -90,7 +90,7 @@ type mempool struct {
 // transactions until they are mined into a block.
 func New(consensus consensusexternalapi.Consensus) miningmanagermodel.Mempool {
 	policy := policy{
-		MaxTxVersion:    0,
+		MaxTxVersion:    1,
```
svarogg, line 93: Please consider using constants.TransactionVersion

svarogg (changes requested):

app/appmessage/domainconverters.go
```diff
@@ -114,15 +114,26 @@ func MsgTxToDomainTransaction(msgTx *MsgTx) *externalapi.DomainTransaction {
 	for _, txOut := range msgTx.TxOut {
 		transactionOutputs = append(transactionOutputs, txOutToDomainTransactionOutput(txOut))
 	}
+
+	payload := make([]byte, 0)
```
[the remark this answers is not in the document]
stasatdaglabs: payload is a slice, so it can be nil. You're likely thinking about payloadHash.

svarogg (dismissed):

svarogg (approved):

PR #1017, opened 2020-11-09
[NOD-1525] Implement headers first ibd

someone235:


app/protocol/flows/ibd/ibd.go
```diff
@@ -55,6 +55,41 @@ func (flow *handleIBDFlow) runIBD() error {
 	flow.peer.WaitForIBDStart()
 	defer flow.FinishIBD()
 
+	syncInfo, err := flow.Domain().Consensus().GetSyncInfo()
```
svarogg, line 58: Shouldn't this be inside the loop?

app/protocol/flows/ibd/ibd.go
```diff
@@ -64,7 +99,104 @@ func (flow *handleIBDFlow) runIBD() error {
 
 	log.Debugf("Found highest shared chain block %s with peer %s", highestSharedBlockHash, flow.peer)
 
-	return flow.downloadBlocks(highestSharedBlockHash, peerSelectedTipHash)
+	return flow.downloadHeaders(highestSharedBlockHash, peerSelectedTipHash)
+}
+
+func (flow *handleIBDFlow) syncMissingBlockBodies() error {
+	hashes, err := flow.Domain().Consensus().GetMissingBlockBodyHashes(flow.peer.SelectedTipHash())
```
svarogg, line 106: Is there any chance we are going to have a billion hashes here, that we need to store for some time in memory, which will kill us?
I assume answer is "premature optimization", but please give that a thought as well.

svarogg (changes requested):

svarogg (dismissed):

svarogg (approved):

PR #1018, opened 2020-11-09
[NOD-1527] Allow to process headers while in missing utxo set sync state

someone235:


domain/consensus/processes/blockprocessor/validateandinsertblock.go
```diff
@@ -15,17 +15,22 @@ func (bp *blockProcessor) validateAndInsertBlock(block *externalapi.DomainBlock)
 
 	hash := consensusserialization.HeaderHash(block.Header)
 	if mode.State == externalapi.SyncStateMissingUTXOSet {
-		headerTipsPruningPoint, err := bp.consensusStateManager.HeaderTipsPruningPoint()
-		if err != nil {
-			return err
-		}
+		if len(block.Transactions) == 0 {
```
svarogg, line 18: I think we want function/boolean IsHeadersOnly

svarogg (changes requested):

svarogg (approved):

PR #1092, opened 2020-11-16
Add TestBigToCompact and TestCompactToBig

someone235:


domain/consensus/utils/math/math_test.go
```diff
@@ -33,3 +34,46 @@ func TestFastLog2Floor(t *testing.T) {
...
 }
+
+// TestBigToCompact ensures BigToCompact converts big integers to the expected
+// compact representation.
+func TestBigToCompact(t *testing.T) {
+	tests := []struct {
+		in  int64
+		out uint32
+	}{
+		{0, 0},
```
svarogg, line 45: Please add some more vectors in both tests.
Consult with Elichai if you need ideas to edge cases.

svarogg (changes requested):

domain/consensus/utils/math/math_test.go
elichai: There are ~20 tests here:
https://github.com/bitcoin/bitcoin/blob/master/src/test/arith_uint256_tests.cpp#L40

svarogg (approved):

PR #1103, opened 2020-11-17
Add blueScore to RPC GetBlock, and add more INFO logs

elichai:


app/protocol/flows/blockrelay/handle_relay_invs.go
```diff
@@ -249,6 +250,8 @@ func (flow *handleRelayInvsFlow) processAndRelayBlock(requestQueue *hashesQueueS
 		return err
 	}
 
+	log.Infof("Accepted domainBlock %s via p2p", blockHash)
```
stasatdaglabs, line 253: Instead, please write: "Accepted block %s via relay"
Also, in the log in submit_block.go, please replace "domainBlock" with just "block"

app/protocol/flows/blockrelay/handle_relay_invs.go
```diff
@@ -237,6 +237,7 @@ func (flow *handleRelayInvsFlow) processAndRelayBlock(requestQueue *hashesQueueS
 			for _, missingAncestor := range missingParentsError.MissingParentHashes {
 				requestQueue.enqueueIfNotExists(missingAncestor)
 			}
+			log.Infof("Received a block with unknown parents, adding to orphan pool: %s", blockHash)
```
stasatdaglabs, line 240: I'd prefer for this log to be inside AddOrphan
Also please write "missing parents" instead of "unknown parents"

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #1105, opened 2020-11-17
[NOD-1548] Remove PoW check from tests

someone235:


domain/dagconfig/params.go
```diff
@@ -147,6 +147,8 @@ type Params struct {
 
 	// DisableDifficultyAdjustment determine whether to use difficulty
 	DisableDifficultyAdjustment bool
+
+	SkipProofOfWork bool
```
svarogg, line 151: Please add a test that makes sure that SkipProofOfWork is false in {main,dev,test,sim}net, so that nobody flips it by mistake.

domain/dagconfig/params.go
```diff
@@ -32,17 +32,17 @@ var (
 	mainPowMax = new(big.Int).Sub(new(big.Int).Lsh(bigOne, 255), bigOne)
```
svarogg, line 32: Why not set mainnet to 239 as well?

svarogg (changes requested):

svarogg (dismissed):

svarogg (approved):

PR #1109, opened 2020-11-18
[NOD-1543] Optimize the performance of Count() in BlockHeaderStore and BlockStore

stasatdaglabs:


domain/consensus/datastructures/blockstore/blockstore.go
```diff
@@ -9,85 +9,119 @@ import (
...
+	err := blockStore.initializeCount(dbContext)
+	if err != nil {
+		return nil, err
+	}
+
+	return blockStore, nil
+}
+
+func (bs *blockStore) initializeCount(dbContext model.DBReader) error {
+	var count uint64
```
svarogg, line 37: I'd rather if you initialize count to be 0, and not rely on it being the default (same for blockHeaderStore)

svarogg (changes requested):

svarogg (approved):

PR #1129, opened 2020-11-19
Add TestFinality back

elichai:
This is based on the `TestFinality` from 0.7.3:
https://sourcegraph.com/github.com/kaspanet/kaspad@v0.7.3-dev/-/blob/domain/blockdag/finality_test.go

### DIFF
--- a/domain/consensus/finality_test.go
+++ b/domain/consensus/finality_test.go
@@ -1,0 +1,169 @@
+package consensus
+
+import (
+	"github.com/kaspanet/kaspad/domain/consensus/model/externalapi"
+	"github.com/kaspanet/kaspad/domain/consensus/utils/consensusserialization"
+	"github.com/kaspanet/kaspad/domain/consensus/utils/testutils"
+	"github.com/kaspanet/kaspad/domain/dagconfig"
+
+	"testing"
+)
+
+func TestFinality(t *testing.T) {
+	testutils.ForAllNets(t, true, func(t *testing.T, params *dagconfig.Params) {
+		// Set finalityInterval to 50 blocks, so that test runs quickly
+		params.FinalityDuration = 50 * params.TargetTimePerBlock
+
+		factory := NewFactory()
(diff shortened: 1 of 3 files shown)


domain/consensus/finality_test.go
```diff
@@ -0,0 +1,164 @@
...
+			mainChainTipHash = consensusserialization.BlockHash(mainChainTip)
+		}
+
+		dagFinality, err := consensus.DAGTraversalManager().BlockAtDepth(model.VirtualBlockHash, params.FinalityDepth())
+		if err != nil {
+			t.Fatalf("TestFinality: Failed getting the virtual's finality point: %v", err)
+		}
+
+		if *dagFinality == *params.GenesisHash {
+			t.Fatalf("virtual's finalityPoint is still genesis after adding finalityInterval + 1 blocks to the main chain")
```
stasatdaglabs, line 121: This doesn't translate very well from the old test.
I realize that you've copy+pasted csm.virtualFinalityPoint() here, but the casual reader probably would not.

I suggest you either write a comment explaining this or externalize virtualFinalityPoint in the CSM.

domain/consensus/finality_test.go
```diff
@@ -0,0 +1,164 @@
...
+		// though it is the block with the highest blue score.
+		for i := uint64(0); i < 2; i++ {
+			sideChainTip, err = buildAndInsertBlock([]*externalapi.DomainHash{sideChainTipHash})
+			if err != nil {
+				t.Fatalf("TestFinality: Failed to process sidechain Block #%d: %v", i, err)
+			}
+			sideChainTipHash = consensusserialization.BlockHash(sideChainTip)
+		}
+
+		// Check that sideChainTip is the bluest tip
```
stasatdaglabs, line 135: Please update this comment not to say "bluest tip"

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #1131, opened 2020-11-22
[RES-65] Add a test for BoundedMergeDepth - new

elichai:
* Add a test for BoundedMergeDepth
* Fix a bug in GetBlockInfo
* Add a test for reproducing and checking that bug

### DIFF
--- a/domain/consensus/consensus.go
+++ b/domain/consensus/consensus.go
@@ -124,6 +124,11 @@ func (s *consensus) GetBlockInfo(blockHash *externalapi.DomainHash) (*externalap
 	}
 	blockInfo.BlockStatus = blockStatus
 
+	// If the status is invalid, then we don't have the necessary reachability data to check if it's in PruningPoint.Future.
+	if blockStatus == externalapi.StatusInvalid {
+		return blockInfo, nil
+	}
+
 	isBlockInHeaderPruningPointFuture, err := s.syncManager.IsBlockInHeaderPruningPointFuture(blockHash)
 	if err != nil {
 		return nil, err
--- a/domain/consensus/finality_test.go
+++ b/domain/consensus/finality_test.go
@@ -1,11 +1,16 @@
 package consensus
 
 import (
+	"github.com/kaspanet/kaspad/domain/consensus/model"
 	"github.com/kaspanet/kaspad/domain/consensus/model/externalapi"
+	"github.com/kaspanet/kaspad/domain/consensus/model/testapi"
+	"github.com/kaspanet/kaspad/domain/consensus/ruleerrors"
 	"github.com/kaspanet/kaspad/domain/consensus/utils/consensusserialization"
 	"github.com/kaspanet/kaspad/domain/consensus/utils/testutils"
 	"github.com/kaspanet/kaspad/domain/dagconfig"
+	"github.com/pkg/errors"
 
+	"fmt"
 	"testing"
 )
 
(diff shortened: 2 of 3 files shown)


domain/consensus/finality_test.go
```diff
@@ -167,3 +172,248 @@ func TestFinality(t *testing.T) {
 		}
 	})
 }
+
+func TestBoundedMergeDepth(t *testing.T) {
+	testutils.ForAllNets(t, true, func(t *testing.T, params *dagconfig.Params) {
+		// Set finalityInterval to 50 blocks, so that test runs quickly
+		params.FinalityDuration = 50 * params.TargetTimePerBlock
```
stasatdaglabs, line 179: Please copy the params before modifying them to avoid possibly interfering with other tests.

domain/consensus/finality_test.go
```diff
@@ -167,3 +172,248 @@ func TestFinality(t *testing.T) {
...
+
+		if int(params.K) >= finalityInterval {
+			t.Fatal("K must be smaller than finality duration for this test to run")
+		}
+
+		checkViolatingMergeDepth := func(consensus testapi.TestConsensus, parents []*externalapi.DomainHash) (*externalapi.DomainBlock, bool) {
+			block, err := consensus.BuildBlockWithParents(parents, nil, nil)
+			if err != nil {
+				t.Fatalf("expected err: %v, found err: %v", ruleerrors.ErrViolatingBoundedMergeDepth, err)
+				select {} // fo some reason go doesn't recognize that t.Fatalf never returns
```
stasatdaglabs, line 190: Since it's not relevant to the test, simply `return` here to make it a little less complex for future readers.

domain/consensus/finality_test.go
```diff
@@ -167,3 +172,248 @@ func TestFinality(t *testing.T) {
...
+			}
+
+			err = consensus.ValidateAndInsertBlock(block)
+			if err == nil {
+				return block, false
+			} else if errors.Is(err, ruleerrors.ErrViolatingBoundedMergeDepth) {
+				return block, true
+			} else {
+				t.Fatalf("expected err: %v, found err: %v", ruleerrors.ErrViolatingBoundedMergeDepth, err)
+				select {} // fo some reason go doesn't recognize that t.Fatalf never returns
```
stasatdaglabs, line 200: Since it's not relevant to the test, simply return here to make it a little less complex for future readers.

domain/consensus/finality_test.go
```diff
@@ -167,3 +172,248 @@ func TestFinality(t *testing.T) {
...
+		finalityInterval := int(params.FinalityDepth())
+
+		if int(params.K) >= finalityInterval {
+			t.Fatal("K must be smaller than finality duration for this test to run")
+		}
+
+		checkViolatingMergeDepth := func(consensus testapi.TestConsensus, parents []*externalapi.DomainHash) (*externalapi.DomainBlock, bool) {
+			block, err := consensus.BuildBlockWithParents(parents, nil, nil)
+			if err != nil {
+				t.Fatalf("expected err: %v, found err: %v", ruleerrors.ErrViolatingBoundedMergeDepth, err)
```
stasatdaglabs, line 189: The error message here is incorrect.

domain/consensus/finality_test.go
```diff
@@ -167,3 +172,248 @@ func TestFinality(t *testing.T) {
...
+			processBlock(consensusReal, block, fmt.Sprintf("blocksChain2 block No %d", i))
+		}
+
+		// submit a block pointing at tip(chain1) and on block2
+		mergeDepthViolatingBlockBottom, isViolatingMergeDepth := checkViolatingMergeDepth(consensusReal, []*externalapi.DomainHash{consensusserialization.BlockHash(blocksChain2[0]), consensusserialization.BlockHash(selectedChain[len(selectedChain)-1])})
+		if !isViolatingMergeDepth {
+			t.Fatalf("TestBoundedMergeDepth: Expected mergeDepthViolatingBlockBottom to violate merge depth")
+		}
+
+		mergeDepthViolatingTop, isViolatingMergeDepth := checkViolatingMergeDepth(consensusReal, []*externalapi.DomainHash{consensusserialization.BlockHash(blocksChain2[len(blocksChain2)-1]), consensusserialization.BlockHash(selectedChain[len(selectedChain)-1])})
```
stasatdaglabs, line 292: Please add a comment explaining what is being checked here.

domain/consensus/consensus.go
```diff
@@ -124,6 +124,11 @@ func (s *consensus) GetBlockInfo(blockHash *externalapi.DomainHash) (*externalap
 	}
 	blockInfo.BlockStatus = blockStatus
 
+	// If the status is invalid, then we don't have the necessary reachability data to check if it's in P.Future.
```
stasatdaglabs, line 127: What's P? Please don't use contractions that never appear in the code.

stasatdaglabs (changes requested):

domain/consensus/finality_test.go
elichai: That sounds like the job of `testutils.ForAllNets` not each individual test

domain/consensus/finality_test.go
elichai: The reason I did this is because it's a closure so I can't just return, but I'll replace it with `return nil, false`

domain/consensus/finality_test.go
elichai: You assume I remember ;)

stasatdaglabs (approved):

PR #1134, opened 2020-11-22
Stop returning teardown from NewTestConsensus

elichai:
This uses the `testing.TB` interface to support both tests and benchmarks https://golang.org/pkg/testing/#TB

and uses the new `T.TempDir()` that was added in Go 1.15 (hence the docker bump).
also use the `T.Cleanup(func(){})` to run the teardown instead of manually doing that in every function


elichai:
We should instead use our own sub-interface only for cleanup and fatalf

PR #1140, opened 2020-11-23
[NOD-1560] Add TestValidateTransactionInIsolation

someone235:
Based on this test https://sourcegraph.com/github.com/kaspanet/kaspad@v0.7.3-dev/-/blob/domain/blockdag/validate_test.go#L563:6 but without the non native subnetwork transactions

### DIFF
--- a/domain/consensus/processes/transactionvalidator/transaction_in_isolation_test.go
+++ b/domain/consensus/processes/transactionvalidator/transaction_in_isolation_test.go
@@ -130,7 +130,6 @@ func TestValidateTransactionInIsolation(t *testing.T) {
 			err := tc.TransactionValidator().ValidateTransactionInIsolation(tx)
 			if !errors.Is(err, test.expectedErr) {
 				t.Errorf("TestValidateTransactionInIsolation: '%s': unexpected error %+v", test.name, err)
-				continue
 			}
 		}
 	})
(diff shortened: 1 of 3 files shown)


domain/consensus/processes/transactionvalidator/transaction_in_isolation_test.go
```diff
@@ -0,0 +1,166 @@
...
+type txSubnetworkData struct {
+	subnetworkID externalapi.DomainSubnetworkID
+	gas          uint64
+	payload      []byte
+}
+
+func TestValidateTransactionInIsolation(t *testing.T) {
+	testutils.ForAllNets(t, true, func(t *testing.T, params *dagconfig.Params) {
+		// Set finalityInterval to 50 blocks, so that test runs quickly
+		params.FinalityDuration = 50 * params.TargetTimePerBlock
```
stasatdaglabs, line 27: Please copy params here so that it doesn't interfere with other tests.

domain/consensus/processes/transactionvalidator/transaction_in_isolation_test.go
```diff
@@ -0,0 +1,166 @@
...
+			tx := createTxForTest(test.numInputs, test.numOutputs, test.outputValue, test.txSubnetworkData)
+
+			if test.extraModificationsFunc != nil {
+				test.extraModificationsFunc(tx)
+			}
+
+			err := tc.TransactionValidator().ValidateTransactionInIsolation(tx)
+			if !errors.Is(err, test.expectedErr) {
+				t.Errorf("TestValidateTransactionInIsolation: '%s': unexpected error %+v", test.name, err)
+				continue
```
stasatdaglabs, line 133: What's the point of this continue?

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #1130, opened 2020-11-20
add gh action for build and test

isabellaliu77:
added github action to build and test the project.  The build will run on three platforms ubuntu, mac, windows.

PR #1164, opened 2020-11-30
Add GitHub actions in preperation for deprecating Jenkins

elichai:
Added a new script `test.sh` that runs everything we want, so it can be easy to run locally.
you can pass `-v` to the script to get verbose outputs. (we might want to enable that in the CI?)
Added github actions that run this script on Ubuntu, OSX and windows.
This is a continuation of #1130

~The Github Action already uses go 1.15, but haven't changed the docker file yet~
It will keep using Go 1.14 for now


elichai:
So this is missing 1 thing:
The ability to re-run whenever the base branch changes, we can either achieve this by writing a custom action that re-triggers the PRs, or by writing an action that comments on all open PRs and then making that comment trigger it

.github/workflows/go.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Increase page size on windows
+        if: runner.os == 'Windows'
+        shell: powershell
+        run: powershell -command .\.github\workflows\SetPageFileSize.ps1
+
+
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
```
svarogg, line 37: Currently we are still at version 1.14
Upgrade to 1.15 should happen for everything at the same time.

test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh -ex
```
svarogg, line 1: This does much more then just test, so I'd rather it be called that accordingly.

svarogg (changes requested):

test.sh
elichai: I consider everything in it part of our "tests" but if you have a better name I'm all ears :)

.github/workflows/go.yml
elichai: Done

svarogg (approved):

PR #1172, opened 2020-12-01
Replace blue score with blue work in ghostdag

### LINKED ISSUE #1021: Replace blue score with blue work in the ghostdag ordering algorithm
The Problem
-----------

There’s an inherent problem with using blueScore to decide the selectedParentChain, and it is that if we assume a responsive DAA that actually gives us an average block rate of 1 block per second, then it’s easy enough for an attacker to create a competing DAG with fake timestamps making exactly 1 block per second such that the difficulty won’t change, and then he can make as big as DAG with the same blueScore (or even higher because he has no red blocks), but with minimal work.

This is the same problem as bitcoin’s longest chain vs heaviest chain, even though a lot is different (we include red blocks both in the DAG and the DAA, and we have a sliding window while bitcoin has a fixed size window)

The Solution
------------

Instead of picking the parent with the most blueScore we pick the parent with the most blueWork.

Definitions
-----------

*   Blue Score of a block - The amount of blue blocks in the DAG _block.Past_ (usually calculated as `Block.SelectedParent.BlueScore + |Block.Blues|`)
    
*   Work - Every block has a “work” associated with it, which is supposed to reflect the amount of hashrate that was spent finding the block. To calculate that we need to use the block’s target, the lower the target the more work you need to do, so “work” has to be an inverse of the target.  
    In practice it is calculated like this: 
    

*   Blue Work of a block - The sum of all the work done in all the blue blocks in the DAG _block.Past_ (calculated similarly to  )

Note
----

After this change, the blueScore of a block is no longer higher than its parents.  
A block could have lower blue score than some of its parents, since the selectedParent is no longer chosen by blueScore.  
It means that we can **only** assume that blueScore is monotonic in a selectedParentChain  
i.e. Every block in the chain has a higher blueScore than its selectedParent but this is **not** generally true in the DAG.

So it is important to make sure we do not rely on this elsewhere. As of commit: [https://github.com/kaspanet/kaspad/commit/3839767aede517622e15b4468189f84f3701cd01][1] I checked all the code and all usages of blueScore are correct and apply either in the context of transactions or for traversing chain blocks.

Implementation
--------------

### blockNode

1.  Add a `blueWork` field to `blockNode` with type `*big.Int` which will contain the blueWork of that block, just like `blueScore` has the blueScore of that block, initialize it to 0 for genesis.
    
2.  Add the field to the serialization logic
    
3.  In the function `func (node *blockNode) less(other *blockNode) bool` replace `blueScore` with `blueWork` like this:
    
    go

This function is called by ghostDAG to sort parents and choose the selectedParent

### ghostDAG

At the end of the function `func (dag *BlockDAG) ghostdag(newNode *blockNode)` assign the `blueWork` variable similar to how `blueScore` is assigned:

1.  Start with `block.blueWork = block.SelectedParent.blueWork`
    
2.  Iterate over the blue set of the block and add their _work_ (not blueWork, because that’s already a sum) to the block’s blueWork.
    

Exact code:

go

### Reachability

In reachability we use `blueScore` to check if a block could become the next selectedParent, instead we should use the `blockNode.less` function directly, this delegates it to use the same mechanism ghostDAG uses.

### blockIndexKey

blockIndexKey is the key in the `block-index` database, it is composed as `blueScore||hash` so that the cursor can then iterate and pull blocks in the topological order (ordered by blueScore, and then by hash), instead we need to change it to be `blueWork||hash`  
Exact code:
```go
func blockIndexKey(blockHash *daghash.Hash, blueWork *big.Int) []byte {
	indexKey := make([]byte, daghash.HashSize+32)
	bytes := blueWork.Bytes()
	copy(indexKey[32-len(bytes):], bytes)
	copy(indexKey[32:], blockHash[:])
	return indexKey
}
```
We also add a `BlueWorkByBlockHash`so that `BlockHashesFrom` can use the `blueWork` of a block when creating a cursor:
```go
// BlueWorkByBlockHash returns the blue work of a block with the given hash.
func (dag *BlockDAG) BlueWorkByBlockHash(hash *daghash.Hash) (*big.Int, error) {
	node, ok := dag.index.LookupNode(hash)
	if !ok {
		return nil, errors.Errorf("block %s is unknown", hash)
	}
	return node.blueWork, nil
}
```

### Full Implementation

[https://github.com/kaspanet/kaspad/tree/blueWork][2]

[1]: https://github.com/kaspanet/kaspad/commit/3839767aede517622e15b4468189f84f3701cd01
[2]: https://github.com/kaspanet/kaspad/tree/blueWork

elichai:
Resolves #1021
More details in the #1021 issue.

This still works like blueScore by: 
1. Inheriting the Blue Work of the SelectedParent.
2. Add to that the Sum of the *Work* of all the blue merge set.

Note that blue work and work are 2 separate things.
blueWork is the sum of all the work of all the blue blocks in the past of a block.
Work is just the work a single block did (`util.CalcWork(block.Bits)`)

Also note that just like with blueScore, the SelectedParent is part of the blue merge set, meaning that you also add the work of the selected parent when you sum the work of the blue merge set.


elichai:
@svarogg I want to note that this will probably affect performance, probably for the worse. but I hope not too much (especially now that the work will always be a single word anyway)

domain/consensus/database/serialization/dbobjects.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.25.0
-// 	protoc        v3.12.3
+// 	protoc        v3.12.4
```
someone235, line 4: Please don't change the protoc version

domain/consensus/processes/ghostdagmanager/ghostdag.go
```diff
@@ -77,9 +80,20 @@ func (gm *ghostdagManager) GHOSTDAG(blockHash *externalapi.DomainHash) error {
 			return err
 		}
 		newBlockData.BlueScore = selectedParentGHOSTDAGData.BlueScore + uint64(len(newBlockData.MergeSetBlues))
+		// We inherit the bluework from the selected parent
+		newBlockData.BlueWork.Set(selectedParentGHOSTDAGData.BlueWork)
+		// Then we add up all the *work*(not blueWork) that all our blues and selected parent did
```
someone235, line 85: all our blues -> all of new block merge set blues

someone235 (changes requested):

elichai:
Rebased

someone235 (dismissed):

someone235 (approved):

PR #1179, opened 2020-12-03
Replace BlueWindow implementation to accomadate a better DAA scheme

### LINKED ISSUE #1022: Adjust the difficulty window in DAA to order by Blue Accumulated Work
The difficulty window is currently defined as the last N blocks in the **blue past** with the highest **blue score** (N is the window size parameter defined here).

We need to change that to the last N **blocks**(not necessarily blue) in the entire past with the highest blue accumulated work (BAW).

More on Blue Accumulated Work in #1021

elichai:
This resolves #1022, basically until now we only used blue blocks in the DAA, instead we made 2 changes:
1. We pick all blocks for the window.
2. We order them by BlueWork, not by BFS or blueScore (this is implicitly changed by #1172).

2 important notes here:
1. For every `B` `B.SelectedParent` has the highest blueWork out of all of the blocks in `B.Past`, so if we go down the SelectedChain we always have the SP as the block with highest blueWork.
2. A block in the merge set can have lower blueWork than a block lower in the SelectedChain.

Because of 1 we can stop searching for higher blueWork if the next SP has lower blueWork than the minimal block in the window(and the window is full).
And because of 2 we need to keep going down the SelectedChain even if we already "filled" the window, until the sentence above is true.

Also, because the mergeSet is ordered we can stop adding to the window if the last block in the merge set failed adding.

Also note that we treat blue and red blocks equally here, resulting in 1 block per second, not 1 blue block per second.

### DIFF
--- a/domain/consensus/processes/dagtraversalmanager/block_heap.go
+++ b/domain/consensus/processes/dagtraversalmanager/block_heap.go
@@ -12,6 +12,10 @@ type blockHeapNode struct {
 	ghostdagData model.BlockGHOSTDAGData
 }
 
+func (left *blockHeapNode) less(right *blockHeapNode, gm model.GHOSTDAGManager) bool {
+	return gm.Less(left.hash, left.ghostdagData, right.hash, right.ghostdagData)
+}
+
 // baseHeap  is an implementation for heap.Interface that sorts blocks by their height
 type baseHeap struct {
 	slice           []*blockHeapNode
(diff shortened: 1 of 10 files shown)


domain/consensus/processes/dagtraversalmanager/block_heap.go
```diff
@@ -101,6 +110,56 @@ func (bh BlockHeap) Push(blockHash *externalapi.DomainHash) error {
 }
 
 // Len returns the length of this heap
-func (bh BlockHeap) Len() int {
+func (bh blockHeap) Len() int {
 	return bh.impl.Len()
 }
+
+// sizedUpBlockHeap represents a mutable heap of Blocks, sorted by their height, capped by a specific size.
```
someone235, line 117: Blocks -> blocks
height -> blue work and hash etc...

domain/consensus/processes/dagtraversalmanager/block_heap.go
```diff
@@ -101,6 +110,56 @@ func (bh BlockHeap) Push(blockHash *externalapi.DomainHash) error {
...
+func (sbh *sizedUpBlockHeap) len() int {
+	return sbh.impl.Len()
+}
+
+// pop removes the block with lowest height from this heap and returns it
+func (sbh *sizedUpBlockHeap) pop() *externalapi.DomainHash {
+	return heap.Pop(&sbh.impl).(*blockHeapNode).hash
+}
+
+// tryPush tries to pushe the block onto the heap, if the heap is full and it's less than the minimum it rejects it
```
someone235, line 145: pushe -> push

domain/consensus/processes/dagtraversalmanager/window.go
```diff
@@ -1,38 +1,80 @@
...
 		return nil, err
 	}
 
-	for uint64(len(window)) < windowSize && currentGHOSTDAGData.SelectedParent != nil {
-		for _, blue := range currentGHOSTDAGData.MergeSetBlues {
-			window = append(window, blue)
-			if uint64(len(window)) == windowSize {
+	windowHeap := dtm.newSizedUpHeap(windowSize)
+
+	for windowHeap.len() <= windowSize && currentGHOSTDAGData.SelectedParent != nil {
```
someone235, line 22: Why do you check the `.len()`? Isn't it enough to check `!added`?

domain/consensus/processes/dagtraversalmanager/window.go
```diff
@@ -1,38 +1,80 @@
...
+
+		// If the window is full and the selected parent is less than the minimum then we break
+		// because this means that there cannot be any more blocks in the past with higher blueWork
+		if !added {
+			break
+		}
+
+		// Now we go over the merge set.
+		// Remove the SP from the blue merge set because we already added it.
+		currentGHOSTDAGData.MergeSetBlues = currentGHOSTDAGData.MergeSetBlues[1:]
```
someone235, line 36: Please don't mutate ghostdag data

domain/consensus/processes/dagtraversalmanager/window.go
```diff
@@ -1,38 +1,80 @@
...
+			added, err := windowHeap.tryPush(currentGHOSTDAGData.MergeSetBlues[i])
+			if err != nil {
+				return nil, err
+			}
+			// If it's smaller than minimum then we won't be able to add the rest because they're even smaller.
+			if !added {
+				break
+			}
+		}
+		for i := len(currentGHOSTDAGData.MergeSetReds) - 1; i >= 0; i-- {
```
someone235, line 48: There can be a situation where a red has higher blue work than a blue in the merge set, but it won't be added because the window is full

domain/consensus/processes/dagtraversalmanager/block_heap.go
```diff
@@ -33,34 +37,39 @@ func (h *baseHeap) Pop() interface{} {
 	return popped
 }
 
+// peek returns the block with lowest height from this heap without removing it
+func (h *baseHeap) peek() *blockHeapNode {
+	return h.slice[0]
```
someone235, line 42: Is it guaranteed to be the first one?

someone235 (changes requested):

domain/consensus/processes/dagtraversalmanager/block_heap.go
elichai: Yes. https://golang.org/pkg/container/heap/
> The minimum element in the tree is the root, at index 0.

domain/consensus/processes/dagtraversalmanager/window.go
elichai: That can't happen because they are separate loops and the `break` only breaks the first loop

someone235 (dismissed):

someone235 (approved):

PR #1183, opened 2020-12-06
Adds "checkDelayedBlock" - checks if the block timeStamp is in the future.

### LINKED ISSUE #1056: Make sure delayed blocks are dropped and not marked as invalid
~If a block comes before its parent we need to make sure it's not marked as invalid, so that after the parent will arrive we can accept it again.~

talelbaz:
Resolves #1056

Adds "checkDelayedBlock" - check if the block timeStamp is in the future.
Adds 2 new fields to blockValidator struct.
Adds new Rule Error "ErrDelayedBlock" .


domain/consensus/ruleerrors/rule_error.go
```diff
@@ -236,6 +236,9 @@ var (
 
 	//ErrPruningPointViolation indicates that the pruning point isn't in the block past.
 	ErrPruningPointViolation = newRuleError("ErrPruningPointViolation")
+
+	//ErrDelayedBlock indicates that timeStamp is in the future.
```
someone235, line 240: Please fix comment to "ErrDelayedBlock indicates that the block timestamp is too much in the future."

someone235 (changes requested):

someone235 (dismissed):

someone235 (approved):

domain/consensus/processes/blockvalidator/block_header_in_isolation.go
```diff
@@ -35,3 +41,15 @@ func (v *blockValidator) checkParentsLimit(header *externalapi.DomainBlockHeader
 	}
 	return nil
 }
+
+func (v *blockValidator) checkBlockTimestampInIsolation(header *externalapi.DomainBlockHeader) error {
+
+	blockTimestamp := header.TimeInMilliseconds
+	now := mstime.Now().UnixMilliseconds()
+	maxCurrentTime := now + int64(v.timestampDeviationTolerance)*v.targetTimePerBlock.Milliseconds()
```
elichai, line 49: Is it `Milliseconds()` on purpose?

PR #1198, opened 2020-12-08
Abstract CheckProofOfWork as a public function and change PoW structure

elichai:
This lets everyone use `CheckProofOfWork` to make if a block has a valid PoW.
it also changes the PoW function to use a different structure that resembles bitcoin's headers.


domain/consensus/model/pow/pow.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/kaspanet/kaspad/domain/consensus/utils/serialization"
+	"github.com/kaspanet/kaspad/util"
+	"github.com/pkg/errors"
+	"math/big"
+)
+
+// CheckProofOfWork check's if the block has a valid PoW according to its Bits field
+// it does not check if the difficulty itself is valid or less than the maximum for the appropriate network
+func CheckProofOfWork(header *externalapi.DomainBlockHeader, optionalTarget *big.Int) bool {
+	if optionalTarget == nil {
```
svarogg, line 16: This is not very go-y to have optionalTarget as parameter.

Instead better do something like (Names were not thought through, feel free to alter):
```
CheckProofOfWorkWithTarget(header, target){
 // Do whatever
}

CheckProofOfWorkAccordingToBits(header){
    CheckProofOfWorkWithTarget(header, util.CompactToBig(header.Bits))
}

```

domain/consensus/model/pow/pow.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	// The block hash must be less than the claimed target unless the flag
+	// to avoid proof of work checks is set.
+	powNum := hashes.ToBig(calcPowValue(header))
+
+	// The block hash must be less or equal than the claimed target.
+	return powNum.Cmp(optionalTarget) <= 0
+}
+
+func calcPowValue(header *externalapi.DomainBlockHeader) *externalapi.DomainHash {
```
svarogg, line 27: Correct me if I misunderstood the code here... seems to me this means that the block hash returned by consensushashing (and therefore used in all APIs, etc.), and the block hash that is used for validating PoW are different.

Do we really want this?

I'd rather if hashserialization.BlockHash did all this as well.

domain/consensus/model/pow/pow.go
```diff
@@ -0,0 +1,55 @@
...
+
+// CheckProofOfWork check's if the block has a valid PoW according to its Bits field
+// it does not check if the difficulty itself is valid or less than the maximum for the appropriate network
+func CheckProofOfWork(header *externalapi.DomainBlockHeader, optionalTarget *big.Int) bool {
+	if optionalTarget == nil {
+		optionalTarget = util.CompactToBig(header.Bits)
+	}
+	// The block hash must be less than the claimed target unless the flag
+	// to avoid proof of work checks is set.
+	powNum := hashes.ToBig(calcPowValue(header))
```
svarogg, line 21: I'd expect clacPowValue to do the hashes.ToBig().
It's pretty weird to the outside onlooker that calcPowValue returns *externalapi.DomainHash

svarogg (changes requested):

domain/consensus/model/pow/pow.go
elichai: Yes, we explictly want to diffrenciate PoW from BlockHash.
They have different requirements, for example PoW doesn't even have to use a hash, just a memory-free hard to solve problem.
even SHA1 can work as PoW but not as a block hash (due to collisions)

domain/consensus/model/pow/pow.go
elichai: You're right. I did that because it's internal and I didn't care much, but I'll change it

svarogg (approved):

PR #1237, opened 2020-12-16
Use pointer receivers when needed

someone235:


domain/consensus/processes/dagtraversalmanager/block_heap.go
```diff
@@ -75,7 +75,7 @@ func (dtm dagTraversalManager) NewDownHeap() model.BlockHeap {
 		dbContext:     dtm.databaseContext,
```
svarogg, line 75: dtm.NewDownHeap should also get receiver by pointer (and maybe other dtm methods as well?)

svarogg (changes requested):

someone235:

svarogg (approved):

Issue #1204, opened 2020-12-09
Implement domain seperated hashing


elichai:
This is needed to prevent collisions between data types (that a block could be interpreted as a transaction etc.)

This can also Solve #1186

Resources:
https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-185.pdf
https://github.com/sipa/bips/blob/bip-schnorr/bip-schnorr.mediawiki 
https://twitter.com/Elichai2/status/1278329253990412288

Issue #1188, opened 2020-12-06
Replace Double-SHA256


elichai:
We currently use bitcoin's double SHA256, but there's no point in doing that.
we have a lot of options to choose from:
Single SHA256
SHA512-256
SHA3-256
Blake2b-256
Blake2s-256
and more.

Things we should consider:
1. Interoperability.
2. CPU performance.
3. Friendliness to binary/arithmetic circuts and R1CS (STARK/SNARK).

For interoperability, the only modern hash function bitcoin supports is SHA256.
Ethereum supports SHA256 and SHA3-256, and there's a [proposal](https://github.com/ethereum/EIPs/issues/152) for Blake2b.
ZCash uses Blake2b for most things, but blake2s for some other specific things (not sure why), [source](https://twitter.com/feministPLT/status/1135114502066884608)

For performance we can see the difference between them in the official Go implementations:
```
goos: linux
goarch: amd64
pkg: go_bench/go
BenchmarkSha256-16                547833              2188 ns/op               0 B/op          0 allocs/op
BenchmarkDoubleSha256-16          479936              2498 ns/op               0 B/op          0 allocs/op
BenchmarkSha512_256-16            697750              1699 ns/op               0 B/op          0 allocs/op
BenchmarkSha3_256-16              389749              2897 ns/op             960 B/op          4 allocs/op
BenchmarkBlake2b-16              1474974               817 ns/op               0 B/op          0 allocs/op
BenchmarkBlake2s-16              1096104              1099 ns/op               0 B/op          0 allocs/op
PASS
```

For SNARK/STARK friendliness, it's hard to know and requires more research but after talking with some people it seems that:
For [PLONK](https://eprint.iacr.org/2019/953) Blake2s is around ~2.5K gates per round, SHA256 and blake2b is around ~5K per round, and keccak/SHA3 is ~12K per round.

We should also think if we care that Schnorr(BIP-340) and ECMH use SHA256 internally

Issue #1186, opened 2020-12-06
Merkle Trees do not diffrenciate between leafs and nodes


elichai:
https://github.com/bitcoin/bitcoin/blob/master/src/consensus/merkle.cpp#L8  
https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2019-February/016697.html

PR #1245, opened 2020-12-17
Replace Double-SHA256 with blake2b and implement domain seperation

elichai:
This solves the following issues: #1204, #1188, #1186.

We chose Blake2b because of its speeds (See https://github.com/kaspanet/kaspad/issues/1188#issue-757981948) we use blake2b's MAC for domain separation because golang x/crypto currently doesn't expose the `personalization` feature (See https://github.com/golang/go/issues/32447).

We also fix the merkle tree bug by doing 2 things:
1. Domain seperation for txs and for merkle branches.
2. If there's an odd number of leafs/branches instead of duplicating we append zeros. 

There's now a new file called `domains.go` that contains all the constructors for the domain separated writers. which should be the only way to hash in kaspad.
if we ever need to hash something new we need to write a new constructor with the appropriate domain


domain/consensus/utils/merkle/merkle.go
```diff
@@ -90,12 +82,12 @@ func merkleRoot(hashes []*externalapi.DomainHash) *externalapi.DomainHash {
 			merkles[offset] = nil
 
 		// When there is no right child, the parent is generated by
-		// hashing the concatenation of the left child with itself.
+		// hashing the concatenation of the left child with zeros.
```
stasatdaglabs, line 85: What are the benefits of this over concatenating the child with itself?

domain/consensus/utils/txscript/opcode.go
```diff
@@ -1962,17 +1960,17 @@ func opcodeHash160(op *parsedOpcode, vm *Engine) error {
...
-func opcodeHash256(op *parsedOpcode, vm *Engine) error {
+// Stack transformation: [... x1] -> [... blake2b(x1)]
+func opcodeBlake2b(op *parsedOpcode, vm *Engine) error {
 	buf, err := vm.dstack.PopByteArray()
 	if err != nil {
 		return err
 	}
-
-	vm.dstack.PushByteArray(hashes.HashData(buf)[:])
+	hash := sha256.Sum256(buf)
```
stasatdaglabs, line 1972: Is this correct?

stasatdaglabs (changes requested):

domain/consensus/utils/merkle/merkle.go
elichai: This prevents having 2 separate inclusion proofs for the same thing in the same merkle tree

elichai: And it is also the standard for how merkle trees are built, see: https://tools.ietf.org/html/rfc6962#section-2.1

domain/consensus/utils/txscript/opcode.go
elichai: No :)

stasatdaglabs (approved):

PR #1250, opened 2020-12-20
Pruning calculation changes

someone235:
This PR deals with the performance issues of calculating the pruning point on two cases:
1) On IBD when it needs an indication if it needs to download the new pruning point UTXO set.
2) When it's up and running and needs to naturally update its pruning point.

The first case is handled by requesting the pruning point hash from the syncing peer, check if we're missing its data, and validate if it's really a pruning point. It solves the performance issue because it's much cheaper to validate if a pruning point is really a pruning point than to calculate it from scratch.

The second case handled by incrementally maintaining a pruning point candidate and update it after each F blocks.

In addition this PR also change selected child iterator to use reachability children because it's more performant.


domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -103,36 +108,68 @@ func (pm *pruningManager) UpdatePruningPointByVirtual() error {
...
 
-	if *newPruningPoint != *currentP {
-		err = pm.savePruningPoint(newPruningPoint)
-		if err != nil {
-			return err
-		}
-		return pm.deletePastBlocks(newPruningPoint)
+	// We move the pruning point every time the candidate's finality score is
+	// bigger than the current pruning point finality score.
+	if pm.finalityScore(newCandidateGHOSTDAGData.BlueScore()) <= pm.finalityScore(pruningPointGHOSTDAGData.BlueScore()) {
```
elichai, line 164: :)

domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -237,28 +274,64 @@ func (pm *pruningManager) deleteBlock(blockHash *externalapi.DomainHash) (alread
...
+		return false, err
+	}
+
+	// A pruning point has to be in the selected chain of the headers selected tip.
+	headersSelectedTipGHOSTDAGData, err := pm.ghostdagDataStore.Get(pm.databaseContext, headersSelectedTip)
+	if err != nil {
+		return false, err
+	}
+
+	isInSelectedParentChainOfHeadersSelectedTip, err := pm.dagTopologyManager.IsInSelectedParentChainOf(block,
```
elichai, line 289: is `pm.dagTopologyManager.IsInSelectedParentChainOf(a,a)==true`?

someone235: Yes

domain/consensus/processes/dagtraversalmanager/selected_child_iterator.go
```diff
@@ -9,21 +9,19 @@ import (
...
 	if err != nil {
 		panic(err)
 	}
 
-	for _, child := range children {
-		if *child == *model.VirtualBlockHash {
-			continue
-		}
-
+	for _, child := range data.TreeNode.Children {
```
michaelsutton, line 24: I think this loop can be replaced with a binary search - using the fact that tree children are ordered by interval 
(in previous versions of the codebase this logic was used in a method called findAncestorOfThisAmongChildrenOfOther - discuss with @svarogg @stasatdaglabs  )

elichai (dismissed):
This seems to work and is better than what we currently have.
we should really document the exact semantics of the pruning point though

elichai (approved):

domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -103,28 +108,71 @@ func (pm *pruningManager) UpdatePruningPointByVirtual() error {
...
+		}
+	}
+
+	if !newCandidate.Equal(currentCandidate) {
+		pm.pruningStore.StagePruningPointCandidate(newCandidate)
+	}
+
+	// We move the pruning point every time the candidate's finality score is
+	// bigger than the current pruning point finality score.
+	if pm.finalityScore(newCandidateGHOSTDAGData.BlueScore()) <= pm.finalityScore(currentPruningPointGHOSTDAGData.BlueScore()) {
```
michaelsutton, line 171: Is this line still necessary?

Issue #1260, opened 2020-12-22
Implement a simple CLI wallet


stasatdaglabs:
Currently, there's no simple (read: not programmatic) way to create wallets, view a wallet's balance, or to send transactions.

Create a simple, no-frills wallet to serve the above functions over RPC, using the new UTXO index.

PR #1261, opened 2020-12-22
Implement a simple CLI wallet

stasatdaglabs:
This implements https://github.com/kaspanet/kaspad/issues/1260
This CLI wallet is based on the example one in kasparov: https://github.com/kaspanet/kasparov/tree/master/examples/wallet

As the issue mentions, this wallet calls kaspad over RPC and capitalizes on its new UTXO index.


cmd/wallet/common.go
```diff
@@ -0,0 +1,10 @@
+package main
+
+import "github.com/kaspanet/kaspad/app/appmessage"
+
+const minConfirmations = 100
```
svarogg, line 5: Take BlockCoinbaseMaturity from DAGParams

cmd/wallet/common.go
```diff
@@ -0,0 +1,10 @@
...
+
+func isUTXOSpendable(entry *appmessage.UTXOsByAddressesEntry, virtualSelectedParentBlueScore uint64) bool {
+	blockBlueScore := entry.UTXOEntry.BlockBlueScore
+	return blockBlueScore+minConfirmations < virtualSelectedParentBlueScore
```
svarogg, line 9: Do this only for coinbase outputs. All other outputs are spendable with 0 confirmations.
Also we have locktimes etc. but I guess it's too much to ask of you rn.

cmd/wallet/create.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		return errors.Wrap(err, "Failed to generate public key")
+	}
+	publicKeySerialized, err := publicKey.Serialize()
+	if err != nil {
+		return errors.Wrap(err, "Failed to serialize public key")
+	}
+
+	for _, netParams := range []*dagconfig.Params{&dagconfig.MainnetParams, &dagconfig.TestnetParams, &dagconfig.DevnetParams} {
+		addr, err := util.NewAddressPubKeyHashFromPublicKey(publicKeySerialized[:], netParams.Prefix)
```
svarogg, line 32: If we think that this should be production-ready, then generating a wallet for all networks at the same time is a bad idea - add cli param of which network we are in instead.

svarogg (changes requested):
Please add a README with usage + explanation on how to use (Or maybe ask Sasha/whomever supervises him to write it? )

stasatdaglabs:

cmd/wallet/README.md
```diff
@@ -0,0 +1,19 @@
...
+`wallet` is a simple, no-frills wallet software operated via the command line.\
+It is capable of generating wallet key-pairs, printing a wallet's current balance, and sending simple transactions.
+
+
+Usage
+-----
+
+* Create a new wallet key-pair: `wallet create --testnet`
+* Print a wallet's current
+  balance: `wallet balance --testnet --address=kaspatest:000000000000000000000000000000000000000000`
```
svarogg, line 17: The formatting is a bit weird here and in the next bullet...
I'd expect the line break to be after `balance: `

svarogg (changes requested):

svarogg (approved):

PR #1266, opened 2020-12-22
Remove the sorting requirement from BlueWindow

elichai:
Right now we require `BlueWindow` to return a slice sorted by GHOSTDAG.
This sorting can take a few precentages of CPU time (on my machine 3s out of 30s profile)
The only component that used the ordering was the DAA for removing the "oldest" block, 
we can replace the definition of "oldest" there to oldest by timestamps, it's debatable which is more correct here but even if it's wrong it will only be by 1/windowSize, which is minor and shouldn't affect the DAA much.

### DIFF
--- a/domain/consensus/processes/dagtraversalmanager/window.go
+++ b/domain/consensus/processes/dagtraversalmanager/window.go
@@ -1,13 +1,11 @@
 package dagtraversalmanager
 
 import (
-	"sort"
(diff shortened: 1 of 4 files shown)


svarogg (approved):

svarogg:

svarogg:

PR #1268, opened 2020-12-23
Fix AddBlock not returning validation failure errors

stasatdaglabs:
Currently, AddBlock swallows RuleErrors and prints them to the log.
This causes SubmitBlock (the RPC method) to never report back to the client that their block was rejected.


elichai (approved):
Looks good

elichai (changes requested):
Can you also change the log in orphans.go to Warn?

elichai (approved):
ACK

PR #1272, opened 2020-12-23
Increase size of reachability and block-relations cache

svarogg:
This increases the size of the reachability and block-relations data-stores cache to 10,000 items, to accommodate csm.mergeSetIncrease


someone235 (approved):

svarogg:

Issue #1277, opened 2020-12-23
Transaction relay does not work


stasatdaglabs:
This is fairly simple to reproduce:
* Launch two kaspad instances connected to one another
* Send a valid transaction to kaspad A
* Observe that it never reaches kaspad B

stasatdaglabs:
Fixed by https://github.com/kaspanet/kaspad/pull/1279

PR #1279, opened 2020-12-23
Fix transaction relay not working

stasatdaglabs:
This PR fixes https://github.com/kaspanet/kaspad/issues/1277

The fix is twofold:
* TestTxRelay calls the getMempoolEntry RPC command to check whether a transaction has been added to the mempool of some node. GetMempoolEntry was unimplemented
* There was a faulty equality check in the transaction relay flow that rejected all relayed transactions


svarogg (approved):

stasatdaglabs:

PR #1280, opened 2020-12-23
Fix deletePastBlocks to check reachability using virtual parents

someone235:
There's a bug in `func (pm *pruningManager) deletePastBlocks` where it checks virtual reachability using `pm.dagTopologyManager.IsAncestorOf`, but it's impossible because virtual doesn't have reachability data. So instead we should check its reachability with `pm.dagTopologyManager.IsAncestorOfAny(tip, virtualParents)`


elichai (approved):
I wonder what's the perf of this whole operation

PR #1284, opened 2020-12-24
Implement a simple mechanism to stop a miner from mining while kaspad is not synced

stasatdaglabs:
This PR reintroduces the IsSynced field in GetBlockTemplateResponse.
Miners are to be encouraged to not mine when IsSynced is false.

Currently, IsSynced is false only when kaspad is in IBD.
This is quite easy to attack, so a more robust solution must be proposed and implemented before mainnet launch.


someone235 (approved):

svarogg:

PR #1286, opened 2020-12-24
Fix the stopOldTemplateSolving channel in the miner getting closed twice

stasatdaglabs:
This is an issue that was introduced in https://github.com/kaspanet/kaspad/pull/1284


svarogg (approved):

svarogg:

PR #1289, opened 2020-12-27
Fix a crash in GetMissingBlockBodyHashes

stasatdaglabs:
The cause of the crash is as follows:
* IBD starts
* The syncee syncs headers
* The syncee requests headers between the highest shared block and the syncer's reported tip
* The syncer, in order to build the response, calls GetHashesBetween. Turns out GetHashes between is limited to 2^17 hashes
* The syncee receives this response and assumes they received all the hashes between the above low and high hashes
* The syncee calls GetMissingBlockBodyHashes for those two hashes
* GetMissingBlockBodyHashes crashes because the high hash (and many headers before it) were never inserted

This PR does two things:
1. Removes the arbitrary limit in GetHashesBetween
2. Adds guards to `consensus` methods so that they'll crash sooner, with more descriptive error messages.

### DIFF
--- a/domain/consensus/processes/syncmanager/antipast.go
+++ b/domain/consensus/processes/syncmanager/antipast.go
@@ -67,14 +67,7 @@ func (sm *syncManager) antiPastHashesBetween(lowHash, highHash *externalapi.Doma
 		}
 	}
 
-	// Pop hashesUpHeap into a slice. Since hashesUpHeap is
-	// an up-heap, it's guaranteed to be ordered from low to high
-	hashesLength := hashesUpHeap.Len()
-	hashes := make([]*externalapi.DomainHash, hashesLength)
-	for i := 0; i < hashesLength; i++ {
-		hashes[i] = hashesUpHeap.Pop()
-	}
-	return hashes, nil
+	return hashesUpHeap.ToSlice(), nil
 }
 
 func (sm *syncManager) missingBlockBodyHashes(highHash *externalapi.DomainHash) ([]*externalapi.DomainHash, error) {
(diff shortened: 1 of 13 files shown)


domain/consensus/processes/syncmanager/antipast.go
```diff
@@ -88,20 +67,25 @@ func (sm *syncManager) antiPastHashesBetween(lowHash, highHash *externalapi.Doma
...
 	}
 
-	// Pop candidateHashes into a slice. Since candidateHashes is
+	// Pop hashesUpHeap into a slice. Since hashesUpHeap is
 	// an up-heap, it's guaranteed to be ordered from low to high
-	hashesLength := maxHashesInAntiPastHashesBetween
-	if candidateHashes.Len() < hashesLength {
-		hashesLength = candidateHashes.Len()
-	}
+	hashesLength := hashesUpHeap.Len()
```
svarogg, line 72: Please make function in heap ToSlice, and use it instead.

svarogg (changes requested):
A few questions:
1. Did you try to sync node after your fix? If not - please do
2. Did you make sure that the maxHashesInAntiPastHashesBetween didn't safeguard against requesting too much something somewhere else?
3. According to what did you decide which functions need the safeguard?

elichai:
We should also check in the syncee that we actually got what we wanted and if not then handle that.
We should never assume the other party gave us what we asked for

stasatdaglabs:
>     2. Did you make sure that the maxHashesInAntiPastHashesBetween didn't safeguard against requesting too much something somewhere else?

As far as I can tell, this was in place because BTCD decided that that was the max amount of hashes there could ever be in any message. This is no longer relevant, since the messages themselves are chunked. I did go through the code, though, and it doesn't look like there's anything that it would otherwise affect.

>     3. According to what did you decide which functions need the safeguard?

We spoke about this and decided to add guards to all consensus functions, in consensus.go.

svarogg (approved):

PR #1297, opened 2020-12-28
Fix a crash caused by orphans whose validation failed

stasatdaglabs:
When unorphaning of a block failed due to a validation failure we erroneously still added its (nil) insertion result to the "successfully unorphaned" slice.


svarogg (approved):

svarogg:

PR #1305, opened 2020-12-28
Add mutable and immutable header interfaces

someone235:
In order to save many copies, I added immutable interface to block header so block headers will be able to be moved safely.


domain/consensus/utils/blockheader/blockheader.go
```diff
@@ -0,0 +1,146 @@
...
+func NewImmutableBlockHeader(
+	version int32,
+	parentHashes []*externalapi.DomainHash,
+	hashMerkleRoot *externalapi.DomainHash,
+	acceptedIDMerkleRoot *externalapi.DomainHash,
+	utxoCommitment *externalapi.DomainHash,
+	timeInMilliseconds int64,
+	bits uint32,
+	nonce uint64,
+) externalapi.ImmutableBlockHeader {
```
stasatdaglabs, line 135: Consider renaming ImmutableBlockHeader to just BlockHeader.

stasatdaglabs (changes requested):

stasatdaglabs (dismissed):

stasatdaglabs (approved):

PR #1310, opened 2020-12-28
Remove from virtualDiffParents in pruning manager

elichai:
Also delete blocks from virtualDiffParents if they're behind P or in P.AC but not in P.Future.

### DIFF
--- a/domain/consensus/processes/pruningmanager/pruningmanager.go
+++ b/domain/consensus/processes/pruningmanager/pruningmanager.go
@@ -184,6 +184,27 @@ func (pm *pruningManager) UpdatePruningPointByVirtual() error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -261,6 +273,24 @@ func (pm *pruningManager) deletePastBlocks(pruningPoint *externalapi.DomainHash)
 			}
 		}
 	}
+
+	// Delete virtual diff parents that are in P.AC and not in V.Past
```
stasatdaglabs, line 277: Please don't abberviate P, AC, and V.

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #1311, opened 2020-12-28
Use sync rate for getBlockTemplate's isSynced

someone235:
In order to avoid mining red blocks from an non synced node, we want to give an indication in getBlockTemplate whether the node is synced.

This PR tries to define better whether the node is synced or not. We use the following indications:
1) If we're during IBD it means we're not synced, so we don't mine.
2) If our selected tip timestamp is not current, it means we'll probably enter IBD soon, so we don't mine.

But this two indications have some problems:
1) An attacker can mislead us to think we're behind by sending us an orphan blocks that can be very easily produced, because we can't validate its difficulty. This will prevent us from mining.
2) If for some reason the mining halts (because a bug in other miners or any other reason), the mining will be freezed forever because the selected tip timestamp will not be recent.

So to avoid those problems we add a third rule:
If the node recently added less blocks to the DAG than the expected with the network block rate, mine anyway.

### DIFF
--- a/app/protocol/flowcontext/sync_rate.go
+++ b/app/protocol/flowcontext/sync_rate.go
@@ -4,7 +4,7 @@ import "github.com/kaspanet/kaspad/util/mstime"
 
 const (
 	syncRateWindowInMilliSeconds                         = 60_000
-	syncRatMaxDeviation                                  = 0.05
+	syncRateMaxDeviation                                 = 0.05
 	maxSelectedParentTimeDiffToAllowMiningInMilliSeconds = 300_000
 )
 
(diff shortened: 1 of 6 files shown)


app/protocol/flowcontext/sync_rate.go
```diff
@@ -0,0 +1,78 @@
+package flowcontext
+
+import "github.com/kaspanet/kaspad/util/mstime"
+
+const (
+	syncRateWindowInMilliSeconds                         = 60_000
+	syncRatMaxDeviation                                  = 0.05
```
svarogg: Typo: Rate

svarogg (approved):

PR #1316, opened 2020-12-29
Make ReachabilityData a read-only interface with a writable variant, to prevent cloning

### LINKED ISSUE #1288: Make reachability read-only to prevent cloning
Reachability data for the same block is being used quite often, and every time it's retrieved from the datastore it's cloned, to prevent editing what's in cache/staging.

This prevents bugs, but the cloning is a huge hog on the CPU.

Instead - make reachability read-only, and only allow editing the reachability data after cloning.

svarogg:
This fixes #1288


domain/consensus/utils/reachabilitydata/reachability_data.go
```diff
@@ -0,0 +1,109 @@
...
+func (rd *reachabilityData) SetInterval(interval *model.ReachabilityInterval) {
+	rd.interval = interval
+}
+
+func (rd *reachabilityData) SetFutureCoveringSet(futureCoveringSet model.FutureCoveringTreeNodeSet) {
+	rd.futureCoveringSet = futureCoveringSet
+}
+
+// Equal returns whether rd equals to other
+func (rd *reachabilityData) Equal(other model.ReadOnlyReachabilityData) bool {
```
someone235, line 85: Is it really needed?

someone235 (changes requested):

someone235 (approved):

someone235 (changes requested):
Remove Equal

domain/consensus/utils/reachabilitydata/reachability_data.go
svarogg: It's only used by the test that makes sure that CloneMutable works correctly. Still important IMHO.

someone235 (approved):

PR #1318, opened 2020-12-29
Don't print the whole UTXODiff to log, it might be quite huge

svarogg:
We have a debug log every time a UTXO-Diff for a block is collected.
This have previously log the whole UTXO-Diff, which might be quite huge. Serializing it into a string and writing to the log file is quite a hog.

This PR changes the log to only state the sizes of the utxo-diff's toAdd and toRemove collections.


someone235 (approved):

elichai (approved):
LGTM

Issue #1306, opened 2020-12-28
Move the heavy lifting in BlockLocator from the syncer to the syncee


elichai:
Currently If a node gets an orphan block `B` where he doesn't know the 32 next blocks in  B.selectedChain he then asks for a block locator, let's call him the Syncee.
next the node that sent that block, lets call it the Syncer, returns a block locator from B to genesis until they find block C, the latest block in B.selectedChain that the Syncee knows, from there it sends all the blocks between B.Past and C.AntiPast. 

Instead we propose that the Syncee will send a BlockLocator on its Headers.SelectedTip.SelectedChain, and together they'll search over the Syncee SelectedChain until they find the highest block in there that has B in its past, and that will be block C, and from there the Syncer sends all the blocks between B.Past and C.AntiPast.

In some cases the syncee might get blocks they already knew about, but at the worse case the syncee is treated like a new node with no history.

elichai:
There are still a bunch of DOS attacks that this opens:
1. The syncer can send you the same block again and again.
2. The syncer can lie in the blockLocator and tell you that C=P and then slowly sync you from P.

We can solve 1 by:
A. Check that all blocks are in C.Future. 
B. Check that no blocks are in Virtual.SelectedChain (of the virtual when you started the sync).
C. Check that the blocks come in topological ordering (ie `ghostdag.Less`).

But we don't yet have a solution for 2.

PR #1324, opened 2020-12-30
Move the heavy lifting in BlockLocator from the syncer to the syncee

stasatdaglabs:
This implements https://github.com/kaspanet/kaspad/issues/1306


app/protocol/flows/blockrelay/ibd.go
```diff
@@ -118,40 +118,57 @@ func (flow *handleRelayInvsFlow) syncHeaders(highHash *externalapi.DomainHash) e
...
-		// getBlockLocator request and try again.
-		locatorHighHash := blockLocatorHashes[0]
-		locatorHighHashInfo, err := flow.Domain().Consensus().GetBlockInfo(locatorHighHash)
+		message, err := flow.dequeueIncomingMessageAndSkipInvs(common.DefaultTimeout)
 		if err != nil {
 			return nil, err
 		}
-		if locatorHighHashInfo.Exists {
-			return locatorHighHash, nil
+		ibdBlockLocatorHighestHashMessage := message.(*appmessage.MsgIBDBlockLocatorHighestHash)
```
svarogg, line 144: The incoming route for this flow registers to many message types. If the message received is not MsgIBDBlockLocatorHighestHash - we'll crash.
Instead, do a checked type assertion, and return a protocol error if it's not ok.

svarogg (changes requested):

svarogg (approved):

PR #1332, opened 2020-12-31
Fix a crash in the UTXO index

stasatdaglabs:
The crash was due to the UTXO index assuming Transaction.Inputs[*].UTXOEntry always exists and is not nil.
Turns out this field is never written to the database because it isn't a permanent part of Transaction.
Once the UTXO index attempts to pull this UTXOEntry from the database (as opposed to from the cache) the application runs into a nil dereference and crashes.

This PR adds a new field to TransactionAcceptanceIndex: TransactionInputUTXOEntries.
The UTXO index then refers to this field instead of relying on Transaction.


domain/consensus/model/acceptancedata_equal_clone_test.go
```diff
@@ -42,6 +42,7 @@ func initTestTransactionAcceptanceDataForClone() []*externalapi.TransactionAccep
 			},
 			1,
 			true,
+			[]externalapi.UTXOEntry{utxo.NewUTXOEntry(1, []byte{0, 1, 2, 3}, true, 2)},
```
someone235, line 45: Add a case where it fails on different UTXO entries

domain/consensus/utils/utxo/utxo_entry.go
```diff
@@ -74,3 +74,12 @@ func (u *utxoEntry) Equal(other externalapi.UTXOEntry) bool {
 
 	return true
 }
+
+func (u *utxoEntry) Clone() externalapi.UTXOEntry {
```
someone235, line 78: Get rid of clone

someone235 (changes requested):

someone235 (approved):

PR #1333, opened 2020-12-31
Add UTXO set cache

someone235:


domain/consensus/datastructures/consensusstatestore/utxo.go
```diff
@@ -79,12 +86,15 @@ func (css *consensusStateStore) commitVirtualUTXOSet(dbTx model.DBTransaction) e
 		return nil
 	}
 
+	css.virtualUTXOSetCache = utxolrucache.New(css.utxoSetCacheSize)
```
stasatdaglabs, line 89: Instead of creating a new cache, please add a clear() method to the cache instead.

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #1340, opened 2021-01-03
Improve the performance of downloading headers

stasatdaglabs:
This PR does two things:
* Creates a new message `BlockHeadersMessage` that contains many `MsgBlockHeader` messages. Sending this one message instead of `MsgBlockHeader`s individually saves a lot of chatter over the network, making the download quicker.
* Keeps a short queue (in the form of a channel) of `BlockHeadersMessages` so that there's never a moment where the node does nothing while waiting for the next headers message to arrive.


someone235 (changes requested):
Get rid of BlockHeaderMessage as an independent p2p message

someone235 (approved):

PR #1352, opened 2021-01-05
Separate reachability interval store field

someone235:


someone235:
It seems like it hurts the performance

PR #1363, opened 2021-01-06
Add sanity check that makes sure that the utxoset that we save fits the pruningPoints's commitment.

svarogg:


elichai (approved):
Looks correct

PR #1364, opened 2021-01-06
Update devnet genesis with older timestamp

elichai:
the current devnet genesis is too up to date, preventing us from creating old datadirs.


testing/integration/setup_test.go
```diff
@@ -110,6 +111,10 @@ func teardownHarness(t *testing.T, harness *appHarness) {
 	if err != nil {
 		t.Errorf("Error closing database context: %+v", err)
 	}
+	err = os.RemoveAll(harness.config.DataDir)
```
svarogg, line 114: Didn't we remove this thing some time ago to make sure we do have the datadir left in case of failures? (also in leveldb_test)

svarogg (changes requested):

testing/integration/setup_test.go
elichai: Removed. I do think we should delete the datadir if nothing fails, but whatever. (we already delete it for the rest of the tests)

svarogg (approved):

PR #1370, opened 2021-01-06
Add basic support for archival node

svarogg:
This adds basic support for running an archival node.
Only change is that blocks are not deleted from the database. They are still set to `StateHeaderOnly`, and therefore won't be returned from the various APIs.

Main reason for this, for now, is to enable debugging.


elichai (approved):

elichai (approved):
Looks good

PR #1386, opened 2021-01-10
Optimize block locator using finality store

elichai:
We store the finality point of every block, we can use that to make big jumps in the block locator.

### DIFF
--- a/domain/consensus/processes/syncmanager/blocklocator.go
+++ b/domain/consensus/processes/syncmanager/blocklocator.go
@@ -25,7 +25,8 @@ func (sm *syncManager) createBlockLocator(lowHash, highHash *externalapi.DomainH
 	currentHash := highHash
 	step := uint64(1)
 	locator := make(externalapi.BlockLocator, 0)
-	for currentHash != nil {
+	// The loop will break if we reached the limit or if we got to lowHash.
+	for {
 		locator = append(locator, currentHash)
 
 		// Stop if we've reached the limit (if it's set)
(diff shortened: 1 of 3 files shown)


domain/consensus/processes/syncmanager/blocklocator.go
```diff
@@ -57,6 +52,20 @@ func (sm *syncManager) createBlockLocator(lowHash, highHash *externalapi.DomainH
 		nextBlueScore := currentBlockBlueScore - step
 		if currentBlockBlueScore < step {
 			nextBlueScore = lowBlockGHOSTDAGData.BlueScore()
+		} else if nextBlueScore < lowBlockGHOSTDAGData.BlueScore() {
+			nextBlueScore = lowBlockGHOSTDAGData.BlueScore()
+		}
```
elichai, line 57: I added this if because now that we do it from PruningPoint and not genesis then we might have `nextBlueScore > lowBlockBlueScore && nextBlueScore > step` which will give us a one-off(because it will then break in the next loop iteration) block lower than lowBlueScore.

domain/consensus/processes/syncmanager/blocklocator.go
```diff
@@ -57,6 +52,20 @@ func (sm *syncManager) createBlockLocator(lowHash, highHash *externalapi.DomainH
...
+
+		for currentBlockBlueScore-nextBlueScore >= sm.finalityDepth {
```
svarogg, line 59: Please move this logic to `dagTraversalManager.LowestChainBlockAboveOrEqualToBlueScore` and add a comment explaining what you did here

svarogg (changes requested):

domain/consensus/processes/syncmanager/blocklocator.go
```diff
@@ -65,6 +60,7 @@ func (sm *syncManager) createBlockLocator(lowHash, highHash *externalapi.DomainH
 			return nil, err
 		}
 
+		locator = append(locator, currentHash)
```
svarogg, line 63: Moving the append to here means you are not including the highHash inside the locator.

svarogg (changes requested):

domain/consensus/processes/syncmanager/blocklocator.go
elichai: You're right. I thought more about the lowHash and not about the highHash. do we want the highHash inside the locator though?

svarogg (approved):

Issue #1373, opened 2021-01-07
Write unit tests for error cases in ValidateAndInsertBlock


elichai:
Right now most error cases in `validateAndInsertBlock` aren't covered, we should write more tests that insert invalid blocks in different ways to trigger different ruleErrors.

stasatdaglabs:
Fixed by the following, among others:
https://github.com/kaspanet/kaspad/pull/1397
https://github.com/kaspanet/kaspad/pull/1398
https://github.com/kaspanet/kaspad/pull/1405
https://github.com/kaspanet/kaspad/pull/1408
https://github.com/kaspanet/kaspad/pull/1415
https://github.com/kaspanet/kaspad/pull/1420
https://github.com/kaspanet/kaspad/pull/1422

PR #1398, opened 2021-01-11
Adding a test for the error cases on the function 'checkBlockStatus()'

talelbaz:
This pull request is part of the issue: #1373 

There are many error cases on the function 'ValidateAndInsertBlock()' that needs to be tested (need to cover more error cases also in functions that are called within this function), therefore, in each PR we will add more tests until we get fully covered.
By now, we are testing all the error cases in the function 'checkBlockStatus()' ( 'validateAndInsertBlock()' -> 'validateBlock()' -> 'checkBlockStatus()').


domain/consensus/processes/blockprocessor/validateandinsertblockErrorCases_test.go
```diff
@@ -0,0 +1,126 @@
...
+		if err == nil {
+			t.Fatalf("Test ValidateAndInsertBlock: Expected an error, because the block is invalid.")
+		}
+		_, err = tc.ValidateAndInsertBlock(blockWithStatusInvalid)
+		if err == nil {
+			t.Fatalf("Test ValidateAndInsertBlock: Expected an error, because the block is invalid.")
+		}
+		if !errors.Is(err, ruleerrors.ErrKnownInvalid) {
+			t.Fatalf("Expected block to be invalid with err: %v, instead found: %v", ruleerrors.ErrKnownInvalid, err)
+		}
```
elichai, line 44: Do you insert the same block twice on purpose or is that by mistake?
also, IMO there's no need to check both that the err isn't nil and that it's the correct error, either write `err == nil || !errors.Is(err, SomeErr)` or just write `!errors.Is(err, SomeErr)` directly

domain/consensus/processes/blockprocessor/validateandinsertblockErrorCases_test.go
```diff
@@ -0,0 +1,126 @@
...
+		errorStr := error.Error(err)
+		if !strings.Contains(errorStr, "is a known invalid block") {
+			t.Fatalf("Test ValidateAndInsertBlock: Expected an diff error, got: %+v.", err)
+		}
```
elichai, line 48: You can write `err.Error()`

domain/consensus/processes/blockprocessor/validateandinsertblockErrorCases_test.go
```diff
@@ -0,0 +1,126 @@
...
+			t.Fatalf("ValidateAndInsertBlock: %+v", err)
+		}
+		// resend the same block.
+		_, err = tc.ValidateAndInsertBlock(block)
+		if err == nil {
+			t.Fatalf("Test ValidateAndInsertBlock: Expected an error, because the block is invalid.")
+		}
+		if !errors.Is(err, ruleerrors.ErrDuplicateBlock) {
+			t.Fatalf("Expected block to be invalid with err: %v, instead found: %v", ruleerrors.ErrDuplicateBlock, err)
+		}
```
elichai, line 65: Same comment about checking both

domain/consensus/processes/blockprocessor/validateandinsertblockErrorCases_test.go
```diff
@@ -0,0 +1,126 @@
...
+		errorStr = error.Error(err)
+		if !strings.Contains(errorStr, " already exists") {
+			t.Fatalf("Test ValidateAndInsertBlock: Expected an diff error, got: %+v.", err)
+		}
```
elichai, line 69: Same comment about `err.Error()`

domain/consensus/processes/blockprocessor/validateandinsertblockErrorCases_test.go
```diff
@@ -0,0 +1,126 @@
...
+		if err == nil {
+			t.Fatalf("Test ValidateAndInsertBlock: Expected an error, because the block is invalid.")
+		}
+		errorStr = error.Error(err)
+		if !errors.Is(err, ruleerrors.ErrDuplicateBlock) {
+			t.Fatalf("Expected block to be invalid with err: %v, instead found: %v", ruleerrors.ErrDuplicateBlock, err)
+		}
+		if !strings.Contains(errorStr, "header already exists") {
+			t.Fatalf("Test ValidateAndInsertBlock: Expected an diff error, got: %+v.", err)
+		}
```
elichai, line 91: Same comments

domain/consensus/processes/blockprocessor/validateandinsertblockErrorCases_test.go
talelbaz: I inserted the same block twice because the error is triggered only if the block already exists.
I'll fix the second comment.

domain/consensus/processes/blockprocessor/validateandinsertblockErrorCases_test.go
talelbaz: fixed

domain/consensus/processes/blockprocessor/validateandinsertblockErrorCases_test.go
```diff
@@ -0,0 +1,114 @@
+package blockprocessor_test
```
someone235, line 1: Please call the file `validateandinsertblock_test.go`. This is more like the go convention, and we also don't use camel case for file names

someone235 (changes requested):

elichai (approved):

PR #1400, opened 2021-01-12
Change log sizes and add some new features to logger

elichai:
1. Increase the default log size to 100MB, and last 8 logs(after tar.gz it should be ~7MB each).
2. add an option to not write logs to stdout (by default it still does write to stdout).
3. allow logs in cwd.
4. Add more logs in pruning manager

### DIFF
--- a/infrastructure/logger/logs.go
+++ b/infrastructure/logger/logs.go
@@ -282,9 +282,11 @@ func (b *Backend) AddLogFile(logFile string, logLevel Level) error {
 // It'll create the file if it doesn't exist.
 func (b *Backend) AddLogFileWithCustomRotator(logFile string, logLevel Level, thresholdKB int64, maxRolls int) error {
 	logDir, _ := filepath.Split(logFile)
-	err := os.MkdirAll(logDir, 0700)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


infrastructure/logger/logs.go
```diff
@@ -281,9 +282,11 @@ func (b *Backend) AddLogFile(logFile string, logLevel Level) error {
 // It'll create the file if it doesn't exist.
 func (b *Backend) AddLogFileWithCustomRotator(logFile string, logLevel Level, thresholdKB int64, maxRolls int) error {
 	logDir, _ := filepath.Split(logFile)
-	err := os.MkdirAll(logDir, 0700)
-	if err != nil {
-		return errors.Errorf("failed to create log directory: %s", err)
+	if logDir != "" {
```
someone235, line 285: I think it's clearer if the condition is on the logFile

someone235 (changes requested):

infrastructure/logger/logs.go
elichai: What will that condition be? the logFile will be either `kaspad.log` or `./kaspad.log`

someone235 (approved):

PR #1402, opened 2021-01-12
Remove erroneous finality optimization from LowestChainBlockAboveOrEqualToBlueScore

someone235:
Using finality points to optimize LowestChainBlockAboveOrEqualToBlueScore is wrong because finality point is defined as highest chain block with blue score lower or equal to b.blueScore-finalityDepth, while LowestChainBlockAboveOrEqualToBlueScore is defined as lowest chain block with blue score above or equal to the provided one.


elichai (approved):
Arghh yeah, blockAtDepth is using finality semantics but `LowestChainBlockBelow` is not (not that it matters so much for block locator)

elichai (changes requested):
Please remove the finalityStore though

elichai (approved):

PR #1406, opened 2021-01-12
Log the networks hashrate 

elichai:
This will allow us/users/monitoring tools to know what's the hashrate of the entire network.
that together with counting the hashrate of each miner helps us also debug the responsiveness of the DAA.


domain/consensus/processes/blockprocessor/validateandinsertblock.go
```diff
@@ -116,7 +117,8 @@ func (bp *blockProcessor) validateAndInsertBlock(block *externalapi.DomainBlock,
 		return nil, err
 	}
 
-	log.Debugf("Block %s validated and inserted", blockHash)
+	hashrate := util.GetHashrateString(util.CompactToBig(block.Header.Bits()), bp.targetTimePerBlock)
```
stasatdaglabs, line 120: This is a bit of a heavy operation to do for every block. Please add it to the log closure below so that it doesn't run at higher log levels.

util/math.go
```diff
@@ -151,3 +152,46 @@ func CalcWork(bits uint32) *big.Int {
 	denominator := new(big.Int).Add(difficultyNum, bigOne)
 	return new(big.Int).Div(oneLsh256, denominator)
 }
+
+func getHashrate(target *big.Int, TargetTimePerBlock time.Duration) *big.Int {
```
stasatdaglabs, line 156: Please move this and GetHashrateString to a separate package, e.g. util/hashrate/hashrate.go

stasatdaglabs (changes requested):

domain/consensus/processes/blockprocessor/validateandinsertblock.go
elichai: I was trying to think how heavy it is, and I thought that if we think it is heavy then we can move it next to the PoW validation where we already have the target in `big.Int` but then it will be its own log.

I can convert this log also to a log closure, but the log closure below is about the virtual, not the block we just received (and I want to see this information also in headers sync)

stasatdaglabs: > I can convert this log also to a log closure
Sounds good, please do.

stasatdaglabs (approved):

PR #1409, opened 2021-01-13
Reject SubmitBlock if the node is in IBD

svarogg:


app/appmessage/rpc_submit_block.go
```diff
@@ -19,11 +19,32 @@ func NewSubmitBlockRequestMessage(block *MsgBlock) *SubmitBlockRequestMessage {
 	}
 }
 
+// RejectReason describes the reason why a block sent by SubmitBlock was rejected
+type RejectReason byte
+
+// RejectReason constants
+const (
+	RejectReasonNone RejectReason = iota
```
someone235, line 27: I think it's better not to use iota if we use the hardcoded numbers on rpc.proto

someone235 (changes requested):

someone235 (approved):

PR #1410, opened 2021-01-13
Prevent a race condition in findHighestSharedBlockHash where we get headersSelectedTip and then pass it as highHash to GetBlockLocator, without locking consensus

svarogg:


someone235 (changes requested):
This is not good enough because the selected chain can be changed after the next iteration

someone235 (approved):

PR #1412, opened 2021-01-13
Send the IBD root UTXO set in chunks instead of a massive monolythic message

stasatdaglabs:


app/protocol/flows/blockrelay/ibd.go
```diff
@@ -325,19 +335,58 @@ func (flow *handleRelayInvsFlow) receiveIBDRootUTXOSetAndBlock() ([]byte, *exter
...
+		message, err := flow.dequeueIncomingMessageAndSkipInvs(common.DefaultTimeout)
+		if err != nil {
+			return nil, nil, false, err
+		}
+
+		switch message := message.(type) {
+		case *appmessage.MsgIBDRootUTXOSetChunk:
+			serializedUTXOSet = append(serializedUTXOSet, message.Chunk...)
+		case *appmessage.MsgDoneIBDRootUTXOSetChunks:
+			receivedAllChunks = true
```
someone235, line 366: I would return from here instead, but it's not a must

someone235 (approved):

someone235 (approved):

someone235 (approved):

PR #1426, opened 2021-01-17
fix missing UTXO notifications

aspect:
The following change fixes missing UTXO notifications, where otherwise, clients are notified only of the last UTXO in the notification set.


elichai (approved):
Let's get these old commits in too

elichai (changes requested):
ops, there's a semicolon

svarogg:
Moved to new PR that is controlled by me, to fix formatting error: https://github.com/kaspanet/kaspad/pull/1428

PR #1428, opened 2021-01-18
Fix missing utxo notifications

svarogg:


elichai (approved):
Stas reviewed and said it's fine in #1426

PR #1429, opened 2021-01-18
Optimize IBD

elichai:
This is a draft for review/comments, it still has a lot of logs in it.


elichai:
This is no longer needed since IBD got a big overhaul

PR #1434, opened 2021-01-19
Remove docker test parallelism

someone235:
Test parallelism requires a lot of memory when building a docker


build_and_test.sh
```diff
@@ -19,4 +19,9 @@ go vet -composites=false $FLAGS ./...
 
 go build $FLAGS -o kaspad .
 
-go test $FLAGS ./...
\ No newline at end of file
+if [ -n "${NO_PARALLEL}" ]
+then
+  go test --parallel=1 $FLAGS ./...
```
elichai, line 24: I think it needs a single dash, not multiple?

PR #1436, opened 2021-01-19
Always request orphan roots, even when you get an inv of a known orphan

someone235:
This fixes a situation where you add a missing parent to the invs queue, and then ignore it because it's a known orphan


app/protocol/flows/blockrelay/handle_relay_invs.go
```diff
@@ -71,7 +72,7 @@ func (flow *handleRelayInvsFlow) start() error {
 		if err != nil {
 			return err
 		}
-		if blockInfo.Exists {
+		if blockInfo.Exists && blockInfo.BlockStatus != externalapi.StatusHeaderOnly {
```
svarogg, line 75: Not sure this is correct.
If we receive an inv of a block that is header-only, we are either in IBD (and then we are going to skip this inv in another branch of the code), or it's already pruned.

Why would we want to request it?

svarogg (changes requested):

app/protocol/flows/blockrelay/handle_relay_invs.go
someone235: The top blocks of the DAG can be header only if IBD didn't finish, and then you can get an orphan that will request these blocks.

svarogg (approved):

PR #1440, opened 2021-01-20
Don't ban peers that sent a requested duplicate block

### LINKED ISSUE #1439: Duplicate blocks should not warrant a ban
If ValidateAndInsertBlock returns ErrDuplicateBlock, as any other RuleError - the peer that sent said block is banned.

There's a separate question on why are DuplicateBlocks being requested - we know that IBD might create this situation, and it seemingly happens in block-relay as well. 

However, no matter what the reason for the request is - a peer that has sent a block that was requested should not be penalized for this, and therefore should not be banned.
And if the block was not requested - the peer is banned even before calling ValidateAndInsertBlock.

svarogg:
This fixes #1439


app/protocol/flows/blockrelay/handle_relay_invs.go
```diff
@@ -109,6 +109,9 @@ func (flow *handleRelayInvsFlow) start() error {
 		log.Debugf("Processing block %s", inv.Hash)
 		missingParents, blockInsertionResult, err := flow.processBlock(block)
 		if err != nil {
+			if errors.Is(err, ruleerrors.ErrDuplicateBlock) {
+				continue
```
someone235, line 113: Can you add log.Errorf?

app/protocol/flows/blockrelay/ibd.go
```diff
@@ -280,7 +281,9 @@ func (flow *handleRelayInvsFlow) processHeader(msgBlockHeader *appmessage.MsgBlo
 		}
 		log.Infof("Rejected block header %s from %s during IBD: %s", blockHash, flow.peer, err)
 
-		return protocolerrors.Wrapf(true, err, "got invalid block %s during IBD", blockHash)
+		if !errors.Is(err, ruleerrors.ErrDuplicateBlock) {
```
someone235, line 284: Can you add log.Errorf for the opposite condition?

app/protocol/flows/blockrelay/ibd.go
```diff
@@ -468,6 +471,10 @@ func (flow *handleRelayInvsFlow) syncMissingBlockBodies(highHash *externalapi.Do
 
 			blockInsertionResult, err := flow.Domain().Consensus().ValidateAndInsertBlock(block)
 			if err != nil {
+				if errors.Is(err, ruleerrors.ErrDuplicateBlock) {
+					log.Debugf("IBD Block %s is a duplicate", blockHash)
```
someone235, line 475: I think this should be Errorf

someone235 (changes requested):

someone235 (dismissed):

someone235 (approved):

someone235 (approved):

PR #1441, opened 2021-01-21
Remove accidental pointer indirection in dbKey

elichai:
`dbKey` is a by-value type that holds a pointer in it. so using it also as a pointer doesn't help anything.
This worked because in Go pointer values also have the method set of their by-value receiver methods (but not the other way around), from the spec:
> The method set of the corresponding pointer type *T is the set of all methods declared with receiver *T or T (that is, it also contains the method set of T).

This should:
1. Remove a little pressure from the GC by removing an unnecessary pointer indirection.
2. Remove all the allocations that happened because the downcast didn't work.

### DIFF
--- a/domain/consensus/database/key.go
+++ b/domain/consensus/database/key.go
@@ -9,6 +9,9 @@ func dbKeyToDatabaseKey(key model.DBKey) *database.Key {
 	if key, ok := key.(dbKey); ok {
 		return key.key
 	}
+	if key, ok := key.(*dbKey); ok {
+		return key.key
+	}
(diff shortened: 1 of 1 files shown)


domain/consensus/database/key.go
```diff
@@ -9,6 +9,9 @@ func dbKeyToDatabaseKey(key model.DBKey) *database.Key {
 	if key, ok := key.(dbKey); ok {
 		return key.key
 	}
+	if key, ok := key.(*dbKey); ok {
+		return key.key
+	}
```
elichai, line 14: I added this just in case we somehow end up with a pointer again

svarogg (approved):

PR #1444, opened 2021-01-24
When the pruning point moves, update its UTXO set outside of a database transaction

stasatdaglabs:
Currently, updating the pruning point's UTXO set takes an enormous amount of memory for batching along.
This PR implements a mechanism to alleviate that.

### DIFF
--- a/domain/consensus/datastructures/pruningstore/pruningstore.go
+++ b/domain/consensus/datastructures/pruningstore/pruningstore.go
@@ -11,7 +11,7 @@ import (
 var pruningBlockHashKey = database.MakeBucket(nil).Key([]byte("pruning-block-hash"))
 var candidatePruningPointHashKey = database.MakeBucket(nil).Key([]byte("candidate-pruning-point-hash"))
 var pruningPointUTXOSetBucket = database.MakeBucket([]byte("pruning-point-utxo-set"))
-var savingNewPruningPointUTXOSetKey = database.MakeBucket(nil).Key([]byte("saving-new-pruning-point-utxo-set"))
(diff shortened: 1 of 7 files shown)


infrastructure/logger/utils.go
```diff
@@ -15,3 +16,11 @@ func LogAndMeasureExecutionTime(log *Logger, functionName string) (onEnd func())
 		log.Debugf("%s end. Took: %s", functionName, time.Since(start))
 	}
 }
+
+// LogMemoryStats logs memory stats for `functionName`
+func LogMemoryStats(log *Logger, functionName string) {
+	stats := runtime.MemStats{}
+	runtime.ReadMemStats(&stats)
```
svarogg, line 23: I think we want to do this in a logClosure.
Reading the code of ReadMemStats it starts with the line
`stopTheWorld("read mem stats")`

So I think it's better not to run it if not necassary.

svarogg (changes requested):

svarogg (approved):

domain/consensus/datastructures/pruningstore/pruningstore.go
```diff
@@ -241,3 +249,15 @@ func (ps *pruningStore) PruningPointUTXOs(dbContext model.DBReader,
 	}
 	return outpointAndUTXOEntryPairs, nil
 }
+
+func (ps *pruningStore) StageStartUpdatingPruningPointUTXOSet() {
+	ps.startUpdatingPruningPointUTXOSetStaging = true
+}
+
+func (ps *pruningStore) HadStartedUpdatingPruningPointUTXOSet(dbContext model.DBWriter) (bool, error) {
```
elichai, line 257: I think you meant `Has`

elichai (approved):

PR #1447, opened 2021-01-25
Pass fromOutpoint into GetPruningPointUTXOs instead of offset, so it could Seek over the cursor

stasatdaglabs:


svarogg (approved):

elichai (approved):
Looks OK

PR #1452, opened 2021-01-26
Make kaspactl usable by human beings

svarogg:


cmd/kaspactl/main.go
```diff
@@ -23,28 +31,56 @@ func main() {
...
-		done <- struct{}{}
-	}()
+	if cfg.RequestJSON != "" {
+		go postJSON(cfg, client, responseChan)
+	} else {
+		go postCommand(cfg, client, responseChan)
+	}
 
-	timeout := time.Duration(cfg.Timeout) * time.Second
+	timeout := time.Duration(cfg.Timeout) * time.Second * 10000 // TODO: remove the * 10000
```
stasatdaglabs, line 42: Please resolve this TODO.

cmd/kaspactl/main.go
```diff
@@ -23,28 +31,56 @@ func main() {
...
+	requestDescs := commandDescriptions()
+	for _, requestDesc := range requestDescs {
+		fmt.Printf("\t%s\n", requestDesc.help())
+	}
+}
+
+func postCommand(cfg *configFlags, client *grpcclient.GRPCClient, responseChan chan string) {
+	message, err := parseCommand(cfg.CommandAndParameters, commandDescriptions())
+	if err != nil {
+		printErrorAndExit(fmt.Sprintf("error parsing command: %+v", err))
```
stasatdaglabs, line 61: I don't think you want to print the stack trace here.

cmd/kaspactl/reflection_helpers.go
```diff
@@ -0,0 +1,46 @@
...
+	return requestTypeWrapped.Field(0).Type.Elem()
+}
+
+// unwrapCommandValue convertes a reflect.Value of a pointer to a wrapped command into a concrete command
+func unwrapCommandValue(commandValueWrapped reflect.Value) reflect.Value {
+	return commandValueWrapped.Elem().Field(0)
+}
+
+// isFieldExported returns true if the given field is exported.
+// Currently the only way to check this is to check if the first run in the field's name is upper case.
```
stasatdaglabs: Type: run -> rune

cmd/kaspactl/command_parser.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	return generateKaspadMessage(commandValue, commandDesc)
+}
+
+func setField(commandValue reflect.Value, parameterValue reflect.Value, parameterDesc *parameterDescription) error {
+	parameterField := commandValue.Elem().FieldByName(parameterDesc.name)
+
+	parameterField.Set(parameterValue)
+	return nil
```
stasatdaglabs, line 51: setField appears to never return an error. Please remove it from the signature.

cmd/kaspactl/command_parser.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		valuePointedTo, err := stringToValue(dummyParameterDesc, valueStr)
+		if err != nil {
+			return reflect.Value{}, errors.WithStack(err)
+		}
+		pointer := pointerToValue(valuePointedTo)
+
+		value = pointer.Interface()
+
+	// Int and uint are not supported because their size is unexpected
```
stasatdaglabs, line 152: I'd write "platform dependent" rather than "unexpected"

stasatdaglabs (changes requested):

cmd/kaspactl/main.go
svarogg: I'm not sure about that... I printed the stack trace precisely because I needed it.

I think a real solution is something like RuleError that is returned when the parameters are wrong, and you print without stack trace, and any other error that you print wholesale.

But for now removed the %+v to be similar to the other errors in this area.

cmd/kaspactl/command_parser.go
```diff
@@ -0,0 +1,170 @@
...
+	case reflect.Invalid:
+	case reflect.Uintptr:
+	case reflect.Complex64:
+	case reflect.Complex128:
+	case reflect.Array:
+	case reflect.Chan:
+	default:
+		return reflect.Value{},
+			errors.Errorf("Unsupported type '%s' for parameter '%s'", parameterDesc.typeof.Kind(), parameterDesc.name)
+	}
```
elichai, line 167: Do these cases fall through into the default case?(ie error)

cmd/kaspactl/commands.go
```diff
@@ -0,0 +1,87 @@
...
+	reflect.TypeOf(protowire.KaspadMessage_GetUtxosByAddressesRequest{}),
+}
+
+type commandDescription struct {
+	name       string
+	parameters []*parameterDescription
+	typeof     reflect.Type
+}
+
+type parameterDescription struct {
```
elichai, line 43: Why not just use `StructField`?

stasatdaglabs (approved):

cmd/kaspactl/commands.go
svarogg: I'd rather not, StructField comes with a lot of strings attached and a lot of redundant stuff.

cmd/kaspactl/command_parser.go
svarogg: I was sure it does, but went to check, and found out it does not.
So fixed.

stasatdaglabs (approved):

PR #1457, opened 2021-01-27
Add more log and measure to pruning points

elichai:


domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -85,6 +85,8 @@ func New(
 // FindNextPruningPoint finds the next pruning point from the
 // given blockHash
 func (pm *pruningManager) UpdatePruningPointByVirtual() error {
+	onEnd := logger.LogAndMeasureExecutionTime(log, "pruningManager.UpdatePruningPointByVirtual")
+	defer onEnd()
```
elichai, line 89: Note that this will run on every block

someone235 (approved):

PR #1458, opened 2021-01-27
Fix duplicate IP addresses in AddressManager

elichai:
Before this we saw sometimes duplicate ips, like: 
```json
{
  "getPeerAddressesResponse": {
    "addresses": [
      {
        "Addr": "52.16.52.205:16211"
      },
      { 
        "Addr": "52.16.52.205:16211"
      }
    ]
  }
}
```

The reson was that one of them was save as ipv6 format + port `00000000000000000000ffff341034cd533f` and the other as ipv4 format + port `341034cd533f`.
Instead we now always convert to ipv6 format, and because it's fixed size we don't need to use the bytes->string trick

After this PR:
```json
{
  "getPeerAddressesResponse": { 
    "addresses": [
      { 
        "Addr": "52.16.52.205:16211"
      }
    ]
  }
}
```

### DIFF
--- a/infrastructure/network/addressmanager/addressmanager.go
+++ b/infrastructure/network/addressmanager/addressmanager.go
@@ -5,7 +5,7 @@
 package addressmanager
 
 import (
-	"encoding/binary"
+	"net"
 	"sync"
 
 	"github.com/kaspanet/kaspad/app/appmessage"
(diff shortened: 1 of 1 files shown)


infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -20,21 +20,21 @@ type AddressRandomizer interface {
 
 // AddressKey represents a "string" key of the ip addresses
 // for use as keys in maps.
```
stasatdaglabs, line 22: Please update this comment.

infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -20,21 +20,21 @@ type AddressRandomizer interface {
...
 func netAddressKey(netAddress *appmessage.NetAddress) AddressKey {
-	port := make([]byte, 2, 2)
-	binary.LittleEndian.PutUint16(port, netAddress.Port)
-
-	key := make([]byte, len(netAddress.IP), len(netAddress.IP)+len(port))
-	copy(key, netAddress.IP)
-
-	return AddressKey(append(key, port...))
+	key := AddressKey{port: netAddress.Port}
+	// all IpV4 can be represented as IPv6.
```
stasatdaglabs, line 35: Either "IpV" or "IPv" are fine, but please pick one of them for both 4 and 6.

stasatdaglabs (changes requested):

stasatdaglabs (approved):

Issue #1461, opened 2021-01-27
Cpu miner bug - testnet ALFA 


dimdumon:
After running cpu miner for couple of hours I got this:
```
2021-01-27 20:48:23.304 [INF] KSMN: Current hash rate is 401.52 Khash/s
2021-01-27 20:48:33.305 [INF] KSMN: Current hash rate is 382.46 Khash/s
2021-01-27 20:48:34.665 [INF] KSMN: Got timeout while requesting block template from localhost:16210
2021-01-27 20:48:34.665 [INF] KSMN: Found block 16693189990fd9c8a673864b803420efc5405eb0af0e83634aefac13e8ee0f82 with parents [1be4d4262e43588cacb40aae20ca745cf6e586431de890ebd0b72d8e5f4b3e7e 5bc598cef70e189fbc88fe769717ef1155d5c5a97a710ae556e922c1886d592a]. Submitting to localhost:16210
2021-01-27 20:48:43.304 [INF] KSMN: Current hash rate is 0.00 Khash/s
2021-01-27 20:48:44.665 [CRT] KSMN: Exiting: Fatal error in goroutine `mineLoop 3`: Error submitting block 16693189990fd9c8a673864b803420efc5405eb0af0e83634aefac13e8ee0f82 to localhost:16210: got timeout after 10s: timeout expired
main.handleFoundBlock
        C:/dev/kaspad/cmd/kaspaminer/mineloop.go:124
main.mineLoop.func1
        C:/dev/kaspad/cmd/kaspaminer/mineloop.go:53
github.com/kaspanet/kaspad/util/panics.handleSpawnedFunction
        C:/dev/kaspad/util/panics/panics.go:84
github.com/kaspanet/kaspad/util/panics.GoroutineWrapperFunc.func1.1
        C:/dev/kaspad/util/panics/panics.go:32
runtime.goexit
        c:/go/src/runtime/asm_amd64.s:1374
error in mine loop
main.main.func1
        C:/dev/kaspad/cmd/kaspaminer/main.go:53
github.com/kaspanet/kaspad/util/panics.handleSpawnedFunction
        C:/dev/kaspad/util/panics/panics.go:84
github.com/kaspanet/kaspad/util/panics.GoroutineWrapperFunc.func1.1
        C:/dev/kaspad/util/panics/panics.go:32
runtime.goexit
        c:/go/src/runtime/asm_amd64.s:1374
[...]

PR #1462, opened 2021-01-28
In kaspaminer, don't crash on submitBlock timeout

stasatdaglabs:
Solves #1461

### DIFF
--- a/cmd/kaspaminer/mineloop.go
+++ b/cmd/kaspaminer/mineloop.go
@@ -118,7 +118,7 @@ func handleFoundBlock(client *minerClient, block *externalapi.DomainBlock) error
 	rejectReason, err := client.SubmitBlock(block)
 	if err != nil {
 		if nativeerrors.Is(err, router.ErrTimeout) {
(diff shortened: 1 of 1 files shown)


cmd/kaspaminer/mineloop.go
```diff
@@ -117,6 +117,10 @@ func handleFoundBlock(client *minerClient, block *externalapi.DomainBlock) error
 
 	rejectReason, err := client.SubmitBlock(block)
 	if err != nil {
+		if nativeerrors.Is(err, router.ErrTimeout) {
+			log.Warnf("Got timeout while submitting block %s to %s: %s", blockHash, client.Address(), err)
+			return nil
```
elichai, line 122: The result of this is that the block will be discarded.
But there's no easy way to solve this right now.
I feel like we should re-write the cpu miner in an async way, that way a separate go-routine will loop trying to submit

elichai (approved):

elichai (approved):

PR #1463, opened 2021-01-28
Force stop gRPC servers after a short timeout

stasatdaglabs:
Currently, if e.g. a miner is connected to a kaspad via RPC, stopping kaspad will hang for a long time.
That is because kaspad attempts to wait for the miner to stop sending requests, which never happens.
This PR alleviates the issue by introducing a short timeout to the gRPCServer Stop() function.


infrastructure/network/netadapter/server/grpcserver/grpc_server.go
```diff
@@ -61,7 +62,20 @@ func (s *gRPCServer) listenOn(listenAddr string) error {
 }
 
 func (s *gRPCServer) Stop() error {
-	s.server.GracefulStop()
+	const stopTimeout = 2 * time.Second
+
+	stopChan := make(chan interface{})
+	go func() {
```
someone235, line 68: Use spawn

someone235 (changes requested):

someone235 (approved):

PR #1466, opened 2021-01-28
Use hard-coded sample config instead of assumed path

### LINKED ISSUE #1124: Stop assuming that `args[0]` is the path to the source code, and that the source code is available
In: https://github.com/kaspanet/kaspad/blob/v0.8.1-dev/infrastructure/config/config.go#L608

kaspad assumes that in `os.Args[0]` there's a sample config file.but this is the path to the **binary** not the source code.
Also, we can't assume that the source code itself is somehow available to us in production binaries.

Because of that, running:
```
rm -rf ~/.kaspad
go run . --devnet
```
results in:
`Error creating a default config file: open /tmp/go-build677716750/b001/exe/sample-kaspad.conf: no such file or directoryexit status 1`

svarogg:
This fixes #1124


elichai (approved):
LGTM

Issue #1053, opened 2020-11-12
The node is reconnecting to a banned peer


elichai:
You can see the node rejected a block, banned the peer, but then reconnects to it:
```
kaspad_1            | 2020-10-05 09:29:55.465 [INF] IBDS: Rejected block 00006bd1c4f068d7faf724c86cd89b3977a1ece9d09d0343e1cae7d9d472d8e5 from <4eaf32420dab27291f483e8781496df5: 10.0.25.82:46452> during IBD: already have block 00006bd1c4f068d7faf724c86cd89b3977a1ece9d09d0343e1cae7d9d472d8e5
kaspad_1            | 2020-10-05 09:29:55.465 [CRT] PROT: found IBD peer <4eaf32420dab27291f483e8781496df5: 10.0.25.82:46452>
kaspad_1            | 2020-10-05 09:29:55.465 [WRN] PROT: Banning <4eaf32420dab27291f483e8781496df5: 10.0.25.82:46452> (reason: got invalid block 00006bd1c4f068d7faf724c86cd89b3977a1ece9d09d0343e1cae7d9d472d8e5 during IBD: already have block 00006bd1c4f068d7faf724c86cd89b3977a1ece9d09d0343e1cae7d9d472d8e5)
kaspad_1            | 2020-10-05 09:29:55.469 [DBG] TXMP: outgoing 'Reject [code 21]' message to 10.0.25.82:46452
kaspad_1            | 2020-10-05 09:29:55.489 [DBG] TXMP: Disconnecting from 10.0.25.82:46452
kaspad_1            | 2020-10-05 09:30:23.570 [DBG] CMGR: Connecting to 10.0.25.82:16611 because we have 5 outgoing connections and the target is 8
kaspad_1            | 2020-10-05 09:30:23.570 [INF] CMGR: Connecting to 10.0.25.82:16611
kaspad_1            | 2020-10-05 09:30:23.570 [INF] TXMP: Dialing to 10.0.25.82:16611
kaspad_1            | 2020-10-05 09:30:23.572 [INF] TXMP: Connected to 10.0.25.82:16611
```

PR #1471, opened 2021-01-31
Ban by IP

### LINKED ISSUE #1380: Bans should be IP based
Right now bans are ip:port based, making them not very effective (since attacker will connect from a different port every time).

This carries a downside - if running a few nodes on the same machine/network and one of said nodes gets banned, all other nodes get banned as well. Nevertheless, it's the best of two evils.

someone235:
Fixes #1380

### DIFF
--- a/domain/consensus/processes/blockvalidator/block_header_in_context.go
+++ b/domain/consensus/processes/blockvalidator/block_header_in_context.go
@@ -2,6 +2,7 @@ package blockvalidator
 
 import (
 	"fmt"
+	"github.com/kaspanet/kaspad/domain/consensus/model"
 	"github.com/kaspanet/kaspad/domain/consensus/model/externalapi"
 	"github.com/kaspanet/kaspad/domain/consensus/ruleerrors"
 	"github.com/kaspanet/kaspad/domain/consensus/utils/consensushashing"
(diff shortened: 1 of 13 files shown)


someone235:
Fixes #1053

domain/consensus/processes/blockvalidator/block_header_in_context.go
```diff
@@ -29,6 +30,15 @@ func (v *blockValidator) ValidateHeaderInContext(blockHash *externalapi.DomainHa
 		if err != nil {
 			return err
 		}
+
+		log.Debug(logger.NewLogClosure(func() string {
+			ghostdagData, err := v.ghostdagDataStore.Get(v.databaseContext, blockHash)
+			if err != nil {
+				panic(err)
```
stasatdaglabs, line 37: Instead of panicking, populate an error variable outside of the closure and check it afterwards.

infrastructure/network/connmanager/outgoing_connections.go
```diff
@@ -42,7 +42,6 @@ func (c *ConnectionManager) checkOutgoingConnections(connSet connectionSet) {
 		err := c.initiateConnection(addressString)
 		if err != nil {
 			log.Infof("Couldn't connect to %s: %s", addressString, err)
-			c.addressManager.RemoveAddress(netAddress)
```
stasatdaglabs, line 45: If you don't remove the address here, doesn't it mean that we keep all addresses indefinitely?

infrastructure/network/addressmanager/addressmanager.go
```diff
@@ -111,7 +117,7 @@ func (am *AddressManager) RemoveAddress(address *appmessage.NetAddress) {
 
 	key := netAddressKey(address)
 	delete(am.addresses, key)
-	delete(am.bannedAddresses, key)
+	delete(am.bannedAddresses, key.address)
```
stasatdaglabs, line 120: Banned addresses should remain banned until they are unbanned. Please remove this line.

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #1473, opened 2021-02-01
Add defaultTargetBlocksPerSecond

someone235:


cmd/kaspaminer/config.go
```diff
@@ -42,7 +43,8 @@ type configFlags struct {
 
 func parseConfig() (*configFlags, error) {
 	cfg := &configFlags{
-		RPCServer: defaultRPCServer,
+		RPCServer:             defaultRPCServer,
+		TargetBlocksPerSecond: defaultTargetBlocksPerSecond,
```
elichai, line 47: Can we instead use `2 * cfg.ActiveNetParams.TargetTimePerBlock`?
(should also probably convert `TargetBlocksPerSecond` to a pointer so we can differentiate not specified from 0, unless we want to say that 0=default and if you want to disable just pass a very big number)

elichai (approved):

PR #1474, opened 2021-02-01
Fix data races in ConnectionManager and flow tests

elichai:


infrastructure/network/connmanager/connmanager.go
```diff
@@ -147,13 +147,14 @@ func (c *ConnectionManager) IsBanned(netConnection *netadapter.NetConnection) (b
...
 	case <-c.resetLoopChan:
-		c.loopTicker.Stop()
-		c.loopTicker = time.NewTicker(connectionsLoopInterval)
+		c.loopTicker.Reset(connectionsLoopInterval)
 	case <-c.loopTicker.C:
 	}
 }
 
 func (c *ConnectionManager) connectionExists(addressString string) bool {
+	c.connectionRequestsLock.Lock()
```
someone235, line 156: I think it's better to just add this in the exported function that calls it (to avoid unexpected deadlocks)

infrastructure/network/connmanager/connmanager.go
```diff
@@ -170,6 +171,8 @@ func (c *ConnectionManager) connectionExists(addressString string) bool {
 }
 
 func (c *ConnectionManager) isPermanent(addressString string) bool {
+	c.connectionRequestsLock.Lock()
```
someone235, line 174: I think it's better to just add this in the exported function that calls it (to avoid unexpected deadlocks)

someone235 (changes requested):

infrastructure/network/connmanager/connmanager.go
elichai: Deleted the function

someone235 (approved):

PR #1481, opened 2021-02-02
Refactor miner and mine when waiting for block to validate

someone235:


cmd/kaspaminer/mineloop.go
```diff
@@ -132,48 +125,65 @@ func handleFoundBlock(client *minerClient, block *externalapi.DomainBlock) error
...
-			headerForMining.SetNonce(i)
-			atomic.AddUint64(&hashesTried, 1)
-			if pow.CheckProofOfWorkWithTarget(headerForMining, targetDifficulty) {
-				block.Header = headerForMining.ToImmutable()
-				foundBlock <- block
-				return
-			}
+		block := getBlockForMining(mineWhenNotSynced)
+		targetDifficulty := difficulty.CompactToBig(block.Header.Bits())
+		headerForMining := block.Header.ToMutable()
```
stasatdaglabs, line 133: I think the above three lines should be outside of the loop.

cmd/kaspaminer/mineloop.go
```diff
@@ -132,48 +125,65 @@ func handleFoundBlock(client *minerClient, block *externalapi.DomainBlock) error
...
+		headerForMining.SetNonce(i)
+		atomic.AddUint64(&hashesTried, 1)
+		if pow.CheckProofOfWorkWithTarget(headerForMining, targetDifficulty) {
+			block.Header = headerForMining.ToImmutable()
+			return block
 		}
 	}
+	panic("Exhausted the nonce space")
```
stasatdaglabs, line 141: This could happen. Please return an error and handle it by retrying with the next template.

cmd/kaspaminer/mineloop.go
```diff
@@ -132,48 +125,65 @@ func handleFoundBlock(client *minerClient, block *externalapi.DomainBlock) error
...
 }
 
-func templatesLoop(client *minerClient, miningAddr util.Address,
-	newTemplateChan chan *appmessage.GetBlockTemplateResponseMessage, errChan chan error, stopChan chan struct{}) {
+func getBlockForMining(mineWhenNotSynced bool) *externalapi.DomainBlock {
+	for i := 0; ; i++ {
```
stasatdaglabs, line 145: Please don't make confusing declarations like that. Instead:
```
tryCounter := 0
for {
  tryCounter++
}
```

cmd/kaspaminer/templatemanager/templatemanager.go
```diff
@@ -0,0 +1,23 @@
...
+var protectedTemplate *appmessage.GetBlockTemplateResponseMessage
+
+// Get returns the template to work on
+func Get() *appmessage.GetBlockTemplateResponseMessage {
+	return (*appmessage.GetBlockTemplateResponseMessage)(
+		atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&protectedTemplate))))
+}
+
+// Set sets the current template to work on
+func Set(template *appmessage.GetBlockTemplateResponseMessage) {
```
stasatdaglabs, line 18: Is there any reason as to why this is an atomic rather than a mutex?

cmd/kaspaminer/mineloop.go
```diff
@@ -45,15 +48,7 @@ func mineLoop(client *minerClient, numberOfBlocks uint64, targetBlocksPerSecond
...
-
-			foundBlock := make(chan *externalapi.DomainBlock)
-			mineNextBlock(client, miningAddr, foundBlock, mineWhenNotSynced, templateStopChan, errChan)
-			block := <-foundBlock
-			templateStopChan <- struct{}{}
-			err := handleFoundBlock(client, block)
-			if err != nil {
-				errChan <- err
-			}
+			foundBlockChan <- mineNextBlock(mineWhenNotSynced)
```
stasatdaglabs, line 51: This implicitly blocks if 100 blocks were solved but submitBlock takes a long time. Is this intentional?

stasatdaglabs (changes requested):

cmd/kaspaminer/mineloop.go
someone235: Yes

cmd/kaspaminer/mineloop.go
someone235: No, because you can get different block for each nonce. This reduces the hash rate by ~20%, but it's just a CPU example miner, so I think it's not a big deal.

cmd/kaspaminer/templatemanager/templatemanager.go
```diff
@@ -0,0 +1,23 @@
+package templatemanager
+
+import (
+	"github.com/kaspanet/kaspad/app/appmessage"
+	"sync"
+)
+
+var currentTemplate *appmessage.GetBlockTemplateResponseMessage
+var lock *sync.Mutex
```
stasatdaglabs, line 9: You never seem to initialize this variable.

docker/Dockerfile
```diff
@@ -27,7 +27,6 @@ WORKDIR /app
 RUN apk add --no-cache ca-certificates tini
 
 COPY --from=build /go/src/github.com/kaspanet/kaspad/kaspad /app/
-COPY --from=build /go/src/github.com/kaspanet/kaspad/sample-kaspad.conf /app/
```
stasatdaglabs, line 30: Why did you remove this?

cmd/kaspaminer/mineloop.go
stasatdaglabs: In that case, please:
Rename initialNonce to nonce
Unfancify the `for` declaration (it's just an infinite loop)
Increment the nonce in the loop's body 
Add a comment explaining what's going on here

stasatdaglabs (changes requested):

cmd/kaspaminer/mineloop.go
```diff
@@ -132,48 +131,67 @@ func handleFoundBlock(client *minerClient, block *externalapi.DomainBlock) error
...
+			block.Header = headerForMining.ToImmutable()
+			log.Infof("Found block %s with parents %s", consensushashing.BlockHash(block), block.Header.ParentHashes())
+			return block
 		}
 	}
 }
 
-func templatesLoop(client *minerClient, miningAddr util.Address,
-	newTemplateChan chan *appmessage.GetBlockTemplateResponseMessage, errChan chan error, stopChan chan struct{}) {
+func getBlockForMining(mineWhenNotSynced bool) *externalapi.DomainBlock {
```
elichai, line 150: @someone235 I must say that all this complication leads me to believe that a single channel with a for+switch on it would be simpler

stasatdaglabs (approved):

PR #1486, opened 2021-02-04
Move testGHOSTDagSorter to testutils, and build a boilerplate for overriding specific managers

elichai:
Every once in a while I find myself wanting to override a manager/store, but I never saw a good way to do it. 
I believe using the Factory in a builder pattern matter and providing constructors is a good way to do this, so even though I don't use it in this PR I decided it's worth to have the skeleton and a sample(ghostdag manager) so that in the future we can extend on that.

I also used the same pattern for the datadir, decreasing the amount of boilerplate.

### DIFF
--- a/domain/consensus/constructors.go
+++ b/domain/consensus/constructors.go
@@ -1,0 +1,6 @@
+package consensus
+
+import "github.com/kaspanet/kaspad/domain/consensus/model"
+
+// GHOSTDAGManagerConstructor is the function signature for a constructor of a type implementing model.GHOSTDAGManager
+type GHOSTDAGManagerConstructor func(model.DBReader, model.DAGTopologyManager, model.GHOSTDAGDataStore, model.BlockHeaderStore, model.KType) model.GHOSTDAGManager
--- a/domain/consensus/factory.go
+++ b/domain/consensus/factory.go
@@ -51,15 +51,21 @@ type Factory interface {
 		externalapi.Consensus, error)
 	NewTestConsensus(dagParams *dagconfig.Params, isArchivalNode bool, testName string) (
 		tc testapi.TestConsensus, teardown func(keepDataDir bool), err error)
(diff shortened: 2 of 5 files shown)


domain/consensus/constructors.go
```diff
@@ -0,0 +1,6 @@
+package consensus
+
+import "github.com/kaspanet/kaspad/domain/consensus/model"
+
+// GHOSTDAGConstructor is the function signature for a constructor of a type implementing model.GHOSTDAGManager
+type GHOSTDAGConstructor func(model.DBReader, model.DAGTopologyManager, model.GHOSTDAGDataStore, model.BlockHeaderStore, model.KType) model.GHOSTDAGManager
```
stasatdaglabs, line 6: Please rename this (and instance variables) to GHOSTDAGManagerConstructor, since we also have GHOSTDAGDataStore.

domain/consensus/factory.go
```diff
@@ -51,15 +51,21 @@ type Factory interface {
 		externalapi.Consensus, error)
 	NewTestConsensus(dagParams *dagconfig.Params, isArchivalNode bool, testName string) (
 		tc testapi.TestConsensus, teardown func(keepDataDir bool), err error)
-	NewTestConsensusWithDataDir(dagParams *dagconfig.Params, dataDir string, isArchivalNode bool) (
-		tc testapi.TestConsensus, teardown func(keepDataDir bool), err error)
+
+	OverrideDataDir(dataDir string)
+	OverrideGHOSTDAGManager(ghostdagConstructor GHOSTDAGConstructor)
```
stasatdaglabs, line 56: These should probably be called something like SetTestDataDir and SetTestGHOSTDAGManager, since they apply only to the test consensus.

stasatdaglabs (changes requested):

domain/consensus/factory.go
elichai: Yeah I thought about Set vs Override, I went with override but Set also works.

stasatdaglabs (approved):

PR #1488, opened 2021-02-04
Add TestBlueWork

someone235:


domain/consensus/processes/ghostdagmanager/ghostdag_test.go
```diff
@@ -177,6 +178,91 @@ func TestGHOSTDAG(t *testing.T) {
 	})
 }
 
+func TestBlueWork(t *testing.T) {
```
stasatdaglabs, line 181: Please add a rudimentary drawing of the DAG you're building in this test, as well as an explanation as to what this test tests.

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #1492, opened 2021-02-04
Remove IsPushOnlyScript from mempool validation

someone235:
This check is redundant because it's validate here as well: https://github.com/kaspanet/kaspad/blob/v0.9.0-dev/domain/consensus/utils/txscript/engine.go#L115


elichai (dismissed):
Looks right

elichai:
I think you meant to link this? https://github.com/kaspanet/kaspad/blob/v0.9.0-dev/domain/consensus/utils/txscript/engine.go#L467

elichai (approved):

someone235:
> I think you meant to link this? https://github.com/kaspanet/kaspad/blob/v0.9.0-dev/domain/consensus/utils/txscript/engine.go#L467

Yeah right

Issue #1314, opened 2020-12-29
Fix and merge old unit test PRs


stasatdaglabs:
The following PRs were closed:
https://github.com/kaspanet/kaspad/pull/960
https://github.com/kaspanet/kaspad/pull/1008
https://github.com/kaspanet/kaspad/pull/1016
https://github.com/kaspanet/kaspad/pull/1043
https://github.com/kaspanet/kaspad/pull/1078
https://github.com/kaspanet/kaspad/pull/1148
https://github.com/kaspanet/kaspad/pull/1212
https://github.com/kaspanet/kaspad/pull/1215
https://github.com/kaspanet/kaspad/pull/1216
https://github.com/kaspanet/kaspad/pull/1218
https://github.com/kaspanet/kaspad/pull/1224
https://github.com/kaspanet/kaspad/pull/1254

PR #960, opened 2020-10-19
[DO NOT MERGE] [NOD-1430] DAGTraversal: add general unit tests.

sept-en:

PR #1500, opened 2021-02-07
Adds new tests for "BlockAtDepth" function and validate the old tests on DAGTraversal.

talelbaz:
#1314 
The first test file on the list - #960 
1. Adds new tests for the function "BlockAtDepth" according to the pr on Jira -https://daglabs.atlassian.net/browse/NOD-1431
2. Validate the correctness of the old tests of dagTraversal (TestLowestChainBlockAboveOrEqualToBlueScore) - no changes.


domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -3,11 +3,109 @@ package dagtraversalmanager_test
...
+	testutils.ForAllNets(t, true, func(t *testing.T, params *dagconfig.Params) {
+		factory := consensus.NewFactory()
+		tc, tearDown, err := factory.NewTestConsensus(params, false,
+			"TestBlockAtDepth")
+		if err != nil {
+			t.Fatalf("NewTestConsensus: %s", err)
+		}
+		defer tearDown(false)
+
+		// "checkExpectedBlockChainDag" - compare BlockAtDepth to a loop on the SelectedChainIterator on a single chain DAG.
```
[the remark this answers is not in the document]
svarogg: I'd also write "compares **the result** of BlockAtDepth to the result of looping over the SelectedChainIterator..."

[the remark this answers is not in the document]
svarogg: Also, you are not using a SelectedChainIterator inside the test, you are "manually" iterating over them using .SelectedParent()

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -3,11 +3,109 @@ package dagtraversalmanager_test
...
+		checkExpectedBlockChainDag := func(highHash *externalapi.DomainHash, depth uint64) {
+			currentBlock := highHash
+			currentBlockData, err := tc.GHOSTDAGDataStore().Get(tc.DatabaseContext(), currentBlock)
+			if err != nil {
+				t.Fatalf("GHOSTDAGDataStore().Get: %+v", err)
```
svarogg, line 27: I know other devs don't do this, but I love it when the t.Fatalf error message has something meaningful inside it e.g. "Failed getting GhostdagData for block with hash %s: %+v"

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -3,11 +3,109 @@ package dagtraversalmanager_test
...
+				if currentBlockData.SelectedParent() == nil {
+					break
+				}
+				currentBlock = currentBlockData.SelectedParent()
+				currentBlockData, err = tc.GHOSTDAGDataStore().Get(tc.DatabaseContext(), currentBlock)
+				if err != nil {
+					t.Fatalf("GHOSTDAGDataStore().Get: %+v", err)
+				}
+			}
+			expected := currentBlock
```
svarogg, line 39: It could be a bit easier to understand if it's called `expectedBlockHash` and `actualBlockHash`

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -3,11 +3,109 @@ package dagtraversalmanager_test
...
+			if err != nil {
+				t.Fatalf("BlockAtDepth: %+v", err)
+			}
+
+			if !blockHash.Equal(expected) {
+				t.Fatalf("Expected block %s but got %s", expected, blockHash)
+			}
+		}
+
+		// "checkExpectedBlockTwoChildrenSameSelectedParent" checks blockAtDepth on 2 children with the same selectedParent.
```
svarogg, line 50: 2 children of whom?
I'd simply write "2 blocks with the same selectedParent"

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -3,11 +3,109 @@ package dagtraversalmanager_test
...
+			expectedSameHash, err := tc.DAGTraversalManager().BlockAtDepth(secondChild, depth)
+			if err != nil {
+				t.Fatalf("BlockAtDepth: %+v", err)
+			}
+			if !blockHash.Equal(expectedSameHash) {
+				t.Fatalf("Expected block %s but got %s", expectedSameHash, blockHash)
+			}
+		}
+
+		// "checkExpectedBlockTwoChildrenSameSelectedParent" checks on different chains, on the same DAG, that they merge at the correct point.
```
svarogg, line 65: This comment is broken I think... can't understand what it means and the name of the function is wrong

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -3,11 +3,109 @@ package dagtraversalmanager_test
...
+
+		}
+
+		highHash, err := createAChainDag(params.GenesisHash, tc)
+		if err != nil {
+			t.Fatalf("createAChainDagInBlockAtDepth: %+v", err)
+		}
+		checkExpectedBlockChainDag(highHash, 2)
+
+		firstChild, secondChild, err := expendDagToTwoChildrenSameSelectedParent(highHash, tc)
```
svarogg, line 94: Another reason to split this to 3 different tests - you don't really want to "Expend" on the DAG from one test-case to the next, as those should be kept separate.
There's some probability that a change in one test will inadvertently effect some other test.

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -3,11 +3,109 @@ package dagtraversalmanager_test
...
+			t.Fatalf("createADagInBlockAtDepth: %+v", err)
+		}
+		checkExpectedBlockTwoChildrenSameSelectedParent(firstChild, secondChild, 2)
+
+		firstChild, secondChild, err = expendDagToDiffChains(firstChild, secondChild, tc)
+		if err != nil {
+			t.Fatalf("createADagInBlockAtDepth: %+v", err)
+		}
+		checkExpectedBlockOnTwoDiffChain(firstChild, secondChild, 3, 2)
+
```
svarogg, line 105: Redundant empty line

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -111,3 +208,52 @@ func TestLowestChainBlockAboveOrEqualToBlueScore(t *testing.T) {
...
+	block := genesisHash
+	var err error
+	for i := 0; i < 5; i++ {
+		block, _, err = tc.AddBlock([]*externalapi.DomainHash{block}, nil, nil)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return block, nil
+
```
svarogg, line 222: This empty line is redundant, but you should add an empty line after the }

I'm not commenting about any more empty lines, just go over the code and clear all of them.

There's a use for empty lines - to separate two distinct blocks of code. These should stay. But just empty lines before or after a { or } shouldn't be there.

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
svarogg: Also, We try as hard as possible to never use abbreviations, because those are very ambiguous.

Specifically here, "Diff" is only used for "UTXODiffs", as in Difference. Using Diff for "Different" is confusing - so better just use the full word

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -3,11 +3,109 @@ package dagtraversalmanager_test
...
+		}
+		checkExpectedBlockChainDag(highHash, 2)
+
+		firstChild, secondChild, err := expendDagToTwoChildrenSameSelectedParent(highHash, tc)
+		if err != nil {
+			t.Fatalf("createADagInBlockAtDepth: %+v", err)
+		}
+		checkExpectedBlockTwoChildrenSameSelectedParent(firstChild, secondChild, 2)
+
+		firstChild, secondChild, err = expendDagToDiffChains(firstChild, secondChild, tc)
```
svarogg, line 100: Dag (as any other ראשי תיבות) should be written with all words in capital, a.k.a. DAG.
This applies to any other place you wrote Dag in this or any other PR.

Note: This convention is somewhat unique to go. In most other languages Dag would have been the correct way to write.

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -111,3 +208,52 @@ func TestLowestChainBlockAboveOrEqualToBlueScore(t *testing.T) {
...
+		return nil, nil, err
+	}
+	return firstChild, secondChild, nil
+
+}
+
+func expendDagToDiffChains(firstChainHash, secondChainHash *externalapi.DomainHash, tc testapi.TestConsensus) (*externalapi.DomainHash, *externalapi.DomainHash, error) {
+	block := firstChainHash
+	var err error
+	for i := 0; i < 3; i++ {
```
svarogg, line 241: Note that here you hard-coded 3 and 2, while `checkExpectedBlockOnTwoDiffChain` receives those.
I'd suggest defining constants for these.

Similar problem you have at `createAChainDag`

svarogg (commented):
I'm not sure why do you need all those checks in separate functions, and also why do you need these functions as closures?

Why not just write them as separate tests, or as separate sections in a single test?

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
talelbaz: Changed, thanks.

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
talelbaz: Changed, thanks.

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
talelbaz: Changed, thanks.

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
talelbaz: Changed, thanks.

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
talelbaz: Changed, thanks.

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
talelbaz: Changed, thanks.

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
talelbaz: Changed, thanks.

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -3,11 +3,113 @@ package dagtraversalmanager_test
 import (
 	"github.com/kaspanet/kaspad/domain/consensus"
 	"github.com/kaspanet/kaspad/domain/consensus/model/externalapi"
+	"github.com/kaspanet/kaspad/domain/consensus/model/testapi"
 	"github.com/kaspanet/kaspad/domain/consensus/utils/testutils"
 	"github.com/kaspanet/kaspad/domain/dagconfig"
 	"testing"
 )
 
+const sizeOfTheChainThatCommonToAllGraphs int = 5
```
svarogg, line 12: This could be shortened to something like "commonChainSize".
Also: you don't have to specify int (it's implied, and the go compiler can understand it by it's own by looking at the value)

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -111,3 +212,75 @@ func TestLowestChainBlockAboveOrEqualToBlueScore(t *testing.T) {
...
+	const sizeOfTheSecondChildSubChainDAG int = 2
+	block := genesisHash
+	var err error
+	for i := 0; i < sizeOfTheChainThatCommonToAllGraphs; i++ {
+		block, _, err = tc.AddBlock([]*externalapi.DomainHash{block}, nil, nil)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	firstChainHash, _, err := tc.AddBlock([]*externalapi.DomainHash{block}, nil, nil)
```
svarogg, line 259: These are not hashes of the chain - they are hashes of the chain tips.
It could be nice that this distinction is reflected in the name (a.k.a. rename to firstChainTipHash, secondChainTipHash)

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -111,3 +212,75 @@ func TestLowestChainBlockAboveOrEqualToBlueScore(t *testing.T) {
...
+	if err != nil {
+		return nil, nil, err
+	}
+	secondChainHash, _, err := tc.AddBlock([]*externalapi.DomainHash{block}, nil, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	block = firstChainHash
```
svarogg, line 268: The block variable is kinda redundant, you could have just done
```
	for i := 0; i < sizeOfTheFirstChildSubChainDAG; i++ {
		firstChainHash, _, err = tc.AddBlock([]*externalapi.DomainHash{firstChainHash}, nil, nil)
		if err != nil {
			return nil, nil, err
		}
	}
```
And when you just initialize firstChainHash
```
firstChainHash, _, err := tc.AddBlock([]*externalapi.DomainHash{genesisHash}, nil, nil)
```

svarogg (changes requested):

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -3,11 +3,137 @@ package dagtraversalmanager_test
...
+	testutils.ForAllNets(t, true, func(t *testing.T, params *dagconfig.Params) {
+		factory := consensus.NewFactory()
+		tc, tearDown, err := factory.NewTestConsensus(params, false,
+			"TestBlockAtDepthOnChainDag")
+		if err != nil {
+			t.Fatalf("Failed creating a NewTestConsensus: %s", err)
+		}
+		defer tearDown(false)
+
+		// This test compares the result of BlockAtDepth to the result of looping over the SelectedChain on a single chain DAG.
```
svarogg, line 25: This comment should be on top of the function declaration, in the format of `// TestBlockAtDepthOnChainDag compares the result...`

(The same for all other comments that describe what a test does)

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -111,3 +236,69 @@ func TestLowestChainBlockAboveOrEqualToBlueScore(t *testing.T) {
...
+		return nil, nil, err
+	}
+	secondChild, _, err := tc.AddBlock([]*externalapi.DomainHash{block}, nil, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	return firstChild, secondChild, nil
+}
+
+func createADAGWithTwoDifferentChains(genesisHash *externalapi.DomainHash, tc testapi.TestConsensus, sizeOfTheFirstChildSubChainDAG int, sizeOfTheSecondChildSubChainDAG int) (*externalapi.DomainHash, *externalapi.DomainHash, error) {
```
svarogg, line 272: Please break this line into two lines (so that it fits in the screen)

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -111,3 +236,69 @@ func TestLowestChainBlockAboveOrEqualToBlueScore(t *testing.T) {
...
+	for i := 0; i < commonChainSize; i++ {
+		block, _, err = tc.AddBlock([]*externalapi.DomainHash{block}, nil, nil)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return block, nil
+}
+
+func createADAGTwoChildrenWithSameSelectedParent(genesisHash *externalapi.DomainHash, tc testapi.TestConsensus) (*externalapi.DomainHash, *externalapi.DomainHash, error) {
```
svarogg, line 252: Please break this line into two lines (so that it fits in the screen)

domain/consensus/processes/dagtraversalmanager/dagtraversalmanager_test.go
```diff
@@ -111,3 +236,69 @@ func TestLowestChainBlockAboveOrEqualToBlueScore(t *testing.T) {
 		checkExpectedBlock(tipHash, 10, blueScore11BlockHash)
 	})
 }
+
+func createAChainDAG(genesisHash *externalapi.DomainHash, tc testapi.TestConsensus) (*externalapi.DomainHash, error) {
```
svarogg, line 240: Please move the helper functions closer to the test functions.
a.k.a.
```
TestA
HelperForTestA
TestB
HelpersForTestB
etc.
```

svarogg (changes requested):

svarogg (changes requested):
You still have a few places that say DAG rather then Dag

svarogg (approved):

PR #1509, opened 2021-02-09
Prune blocks between old and imported pruning points when importing pruning point, to assure that no unrpuned blocks remain

### LINKED ISSUE #1416: Full blocks below or in the anticone of the pruning point that existed before IBD, are never pruned
Consider a network where the pruning point is not genesis.
A node that was once synced did not work for a few days is turned back on.
The node will download headers, move it's pruning point to the network's pruning point, download and apply it's UTXO-set.
However, the unpruned blocks it had prior to turning back on have header-only blocks on top of them, and therefore will never be pruned.

Thoughts about design:
One of two approaches is possible:
1. Pre-application of pruning point - Start from tips, BFS towards genesis or till you reach genesis or headers-only blocks.
2. Start from previous pruningPoint, BFS towards new pruning point, till you reach header-only blocks.

svarogg:
This fixes #1416 

Things to be cautions of while reviewing:
* Is `PruneBlocksBelowImportedPruningPoint` called in the correct point?
* Does `PruneBlocksBelowImportedPruningPoint` miss any blocks that should be pruned / prune any blocks that shouldn't?


svarogg:
We have figured this method would be too pricey in some cases.
Closing PR to re-implement using another TBD method.

PR #1511, opened 2021-02-09
Change some logs from trace to debug

elichai:
These are accidentally trace (their start log is debug and end log is trace)

### DIFF
--- a/domain/consensus/processes/consensusstatemanager/resolve_block_status.go
+++ b/domain/consensus/processes/consensusstatemanager/resolve_block_status.go
@@ -56,7 +56,7 @@ func (csm *consensusStateManager) resolveBlockStatus(blockHash *externalapi.Doma
 
 		csm.blockStatusStore.Stage(unverifiedBlockHash, blockStatus)
 		selectedParentStatus = blockStatus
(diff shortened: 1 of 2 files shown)


domain/consensus/processes/consensusstatemanager/resolve_block_status.go
```diff
@@ -56,7 +56,7 @@ func (csm *consensusStateManager) resolveBlockStatus(blockHash *externalapi.Doma
 
 		csm.blockStatusStore.Stage(unverifiedBlockHash, blockStatus)
 		selectedParentStatus = blockStatus
-		log.Debugf("Block %s status resolved to `%s`", unverifiedBlockHash, blockStatus)
+		log.Debugf("Block %s status resolved to `%s`, finished %d/%d of unverified blocks", unverifiedBlockHash, blockStatus, len(unverifiedBlocks)-i, len(unverifiedBlocks))
```
someone235, line 59: I think you should replace `len(unverifiedBlocks)-i` with `i+1`

someone235 (changes requested):

someone235 (approved):

PR #1513, opened 2021-02-09
Prune blocks below pruning point when moving pruning point during IBD

svarogg:


domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -221,47 +224,47 @@ func (pm *pruningManager) deletePastBlocks(pruningPoint *externalapi.DomainHash)
...
+
+	err = pm.pruneVirtualDiffParents(pruningPoint, virtualParents)
+	if err != nil {
+		return err
 	}
 
+	return nil
+}
+
+func (pm *pruningManager) deleteBlocksDownward(queue model.BlockHeap) error {
```
elichai, line 267: Can you repalce the parents loop in there(line 286-287) with `queue.PushSlice(parents)`?

domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -551,3 +584,33 @@ func (pm *pruningManager) updatePruningPointUTXOSet() error {
...
+	if err != nil {
+		return err
+	}
+
+	for ok := iterator.First(); ok; ok = iterator.Next() {
+		blockHash, err := iterator.Get()
+		if err != nil {
+			return err
+		}
+		isInPastOfPruningPoint, err := pm.dagTopologyManager.IsDescendantOf(blockHash, pruningPointHash)
```
elichai, line 602: nit, I'd prefer if we just called `IsAncestorOf(pruningPointHash, blockHash)` and completely remove `IsDescendantOf`

svarogg: Fun fact: I was the only user of IsDescenadntOf, so removed it.

elichai (approved):

PR #1524, opened 2021-02-14
Cleanup the logger and make it asynchronous

### LINKED ISSUE #1035: Look into replacing the logger
Keep same API of loggerSee following medium article: https://medium.com/syrinsecurity/golang-logging-fast-and-concurrently-3780d98fdad8

The following library is a good candidate, but give a deeper look before picking: https://github.com/uber-go/zap

Make sure to preserve the sub-system functionality, but try to make it with less boilerplate

elichai:
* Clean up and remove unused utilities/features
* Make the logger abstract over writers, and have a special go-routine for writing the logs to the writers.
* The formatting of the logs happen in the goroutine that wrote the log, but then it's sent to the backend goroutine.

Note that now we have to run the backend before changing the log level+writing to the log.
also you can only add writers *before* you run the backend.

Closes #1035


app/app.go
```diff
@@ -149,6 +150,7 @@ func (app *kaspadApp) main(startedChan chan<- struct{}) error {
 			log.Criticalf("Graceful shutdown timed out %s. Terminating...", shutdownTimeout)
 		}
 		log.Infof("Kaspad shutdown complete")
+		logger.BackendLog.Close()
```
elichai, line 153: It's apparently not quite clear where is the last exit route of the app. 
I thought this ism but it looks like some goroutines are still exiting and writing logs after this is called?

infrastructure/logger/backend.go
```diff
@@ -0,0 +1,187 @@
...
+	})
+	return nil
+}
+
+// Run launches the logger backend in a separate go-routine. should only be called once.
+func (b *Backend) Run() error {
+	if !atomic.CompareAndSwapUint32(&b.isRunning, 0, 1) {
+		return errors.New("The logger is already running")
+	}
+	go func() {
```
svarogg, line 144: This is the main loop of the logger, I'd give it the honor of it's own function.
(Or maybe not this function, but the `for log := range b.writeChan` loop itself, to keep all the boilerplate out of it)

infrastructure/logger/level.go
```diff
@@ -0,0 +1,73 @@
...
+	LevelDebug
+	LevelInfo
+	LevelWarn
+	LevelError
+	LevelCritical
+	LevelOff
+)
+
+// levelStrs defines the human-readable names for each logging level.
+var levelStrs = [...]string{"TRC", "DBG", "INF", "WRN", "ERR", "CRT", "OFF"}
```
svarogg, line 21: What does the `[...]` mean?

infrastructure/logger/level.go
```diff
@@ -0,0 +1,73 @@
...
+// the level will not produce any log output.
+func (l Level) String() string {
+	if l >= LevelOff {
+		return "OFF"
+	}
+	return levelStrs[l]
+}
+
+// validLogLevel returns whether or not logLevel is a valid debug log level.
+func validLogLevel(logLevel string) bool {
```
svarogg, line 57: It seems like this function is redundant, since it's always called a few moments before SetLogLevel is called, which re-validates the log level string.

However, if you think it's still needed, why not `_, ok := LevelFromString; return ok`?
This will re-use the hard-coded strings in LevelFromString.

Alternatively, create a map between strings to levels, and then use then in both functions.

Or even better - enjoy both worlds - create a map between string to levels just to avoid hard-coded strings, and then delete validLogLevel/re-use LevelFromString because reasons from above.

infrastructure/logger/logger.go
```diff
@@ -26,199 +27,103 @@ var (
...
-	SubsystemTags.P2PS: p2psLog,
-	SubsystemTags.NTAR: ntarLog,
-	SubsystemTags.DNSS: dnssLog,
-	SubsystemTags.SNVR: snvrLog,
-	SubsystemTags.WSVC: wsvcLog,
-	SubsystemTags.REAC: reacLog,
-	SubsystemTags.PRNM: prnmLog,
-	SubsystemTags.BLVL: blvlLog,
+// InitLogStdoutOnly attaches stdout to the backend log and starts the logger.
+func InitLogStdoutOnly(logLevel Level) {
```
svarogg, line 49: The word "Only" is somewhat misleading here... 
It's `InitLogStdoutPlusAllOtherBackendsThatIHaveAlreadyInited` (As can be seen when this is called from InitLog).

But seriously - I'd rename to `InitLogStdout`

svarogg (changes requested):

cmd/kaspaminer/log.go
```diff
@@ -24,4 +25,15 @@ func initLog(logFile, errLogFile string) {
 		fmt.Fprintf(os.Stderr, "Error adding log file %s as log rotator for level %s: %s", errLogFile, logger.LevelWarn, err)
 		os.Exit(1)
 	}
+	err = backendLog.AddLogWriter(os.Stdout, logger.LevelInfo)
```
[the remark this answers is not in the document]
elichai: If we want to log to stdout then yeah.

infrastructure/logger/backend.go
```diff
@@ -0,0 +1,187 @@
...
+// New logger backends can override these default flags using WithFlags.
+// We're using this instead of `init()` function because variables are initialized before init functions,
+// and this variable is used inside other variable intializations, so runs before them
+var defaultFlags = getDefaultFlags()
+
+// Flags to modify Backend's behavior.
+const (
+	// Llongfile modifies the logger output to include full path and line number
+	// of the logging callsite, e.g. /a/b/c/main.go:123.
+	Llongfile uint32 = 1 << iota
```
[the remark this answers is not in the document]
elichai: I think Log?
that was the original name

infrastructure/logger/backend.go
```diff
@@ -0,0 +1,187 @@
...
+					_, _ = writer.Write(log.log)
+				}
+			}
+		}
+	}()
+	return nil
+}
+
+// IsRunning returns true if backend.Run() has been called and false if it hasn't.
+func (b *Backend) IsRunning() bool {
```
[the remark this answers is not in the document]
elichai: `Logger.printf` uses it to panic.
but actually they're in the same package so I can make it private

infrastructure/logger/level.go
elichai: But with a map I need to care about concurrency (or assume no one will ever modify that map)

infrastructure/logger/backend.go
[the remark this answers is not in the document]
elichai: We return an error in `backend.Run()` is called twice, so I think it makes sense to let someone check if it's running?
(it also helps me avoid fixing the name collision with the `isRunning` field lol)

svarogg (approved):

PR #1529, opened 2021-02-15
Don't mark block that got rejected because of ruleerrors.ErrPrunedBlock as invalid

someone235:
If someone sends you a pruned block you reject the block, but the header is still valid


domain/consensus/processes/blockprocessor/validateblock.go
```diff
@@ -54,11 +54,14 @@ func (bp *blockProcessor) validateBlock(block *externalapi.DomainBlock, isPrunin
 	if err != nil {
 		if errors.As(err, &ruleerrors.RuleError{}) {
 			// If we got ErrMissingParents the block shouldn't be considered as invalid
-			// because it could be added later on when its parents are present, and if
```
svarogg, line 57: I'd reformat this comment as something like:
```
We don't mark the following errors as invalid for the following reasons:
ErrMissingParents - explanation
ErrBlabla - explanation
etc.
```

svarogg (changes requested):

svarogg (approved):

PR #1534, opened 2021-02-16
Nightly run for go test race detector

elichai:
This will run every night at 00:00 UTC `go test -race ./...` on the latest version branch + on master.
we can also trigger it manually in `Actions` and choose which branch we want it to run on.

Example of manually choosing: https://github.com/elichai/kaspad/runs/1912918114

(note that even when manually choosing the branch it will show it as master+latest, but both will just run on that branch )

### DIFF
--- a/.github/workflows/go-race.yml
+++ b/.github/workflows/go-race.yml
@@ -1,0 +1,48 @@
+name: Go-Race
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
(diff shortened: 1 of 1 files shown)


elichai:
The scheduled run worked: https://github.com/elichai/kaspad/actions/runs/573292293
(didn't pass but that's because my latest is an old branch)

someone235 (approved):

PR #1536, opened 2021-02-17
Make antiPastHashesBetween return blocks sorted in ghostdag-order

svarogg:
This PR fixes a performance bug in IBD:

Currently IBD synced block bodies in BFS order, validating against their UTXO-Set blocks that would later not be in the selected-parent-chain. Those blocks are then left as virtualDiffParents, forcing us to update diffs between them and virtual until those are merged.

This PR re-orders the list of blocks returned by antiPastHashesBetween to be in an order where the selectedParent is always first, thus he is the only block in it's merge set that will be validated against UTXO-Set.

### DIFF
--- a/domain/consensus/processes/syncmanager/antipast.go
+++ b/domain/consensus/processes/syncmanager/antipast.go
@@ -137,7 +137,7 @@ func (sm *syncManager) findHighHashAccordingToMaxBlueScoreDifference(lowHash *ex
 	highHash *externalapi.DomainHash, maxBlueScoreDifference uint64, highBlockGHOSTDAGData *model.BlockGHOSTDAGData,
 	lowBlockGHOSTDAGData *model.BlockGHOSTDAGData) (*externalapi.DomainHash, error) {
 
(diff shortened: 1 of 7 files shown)


domain/consensus/processes/syncmanager/antipast.go
```diff
@@ -55,78 +47,137 @@ func (sm *syncManager) antiPastHashesBetween(lowHash, highHash *externalapi.Doma
...
+		highBlockGHOSTDAGData, err = sm.ghostdagDataStore.Get(sm.databaseContext, highHash)
 		if err != nil {
 			return nil, err
 		}
-		if !isInPastOfOriginalLowHash {
-			err = hashesUpHeap.Push(current)
-			if err != nil {
-				return nil, err
-			}
+		if highBlockGHOSTDAGData.BlueScore()-lowBlockGHOSTDAGData.BlueScore()+1 > maxBlueScoreDifference {
```
someone235, line 157: You should update the high hash after this if, and not before

someone235 (changes requested):

someone235 (approved):

PR #1541, opened 2021-02-18
Make the CI more verbose and cache kaspad dependencies in the dockerfile

elichai:
This might help us solve the nondeterministic failures that only happen in the CI

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -48,7 +48,7 @@ jobs:
 
       - name: Test
         shell: bash
-        run: ./build_and_test.sh
+        run: ./build_and_test.sh -v
 
   coverage:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -48,7 +48,7 @@ jobs:
 
       - name: Test
         shell: bash
-        run: ./build_and_test.sh
+        run: ./build_and_test.sh -v
```
svarogg, line 51: What's the point of this? 
build_and_test.sh doesn't take `-v` AFAICT.

svarogg (changes requested):

.github/workflows/go.yml
elichai: it does take `-v` :)
it'll pass it into the go commands, so it will print the name of the test it's running, which will help us figure out exactly which test is spuriously failing at `leveldb closed`

svarogg (approved):

PR #1542, opened 2021-02-18
Close iterators

elichai:
We use goleveldb for our database, and it explictly says that all iterators must be released:

> The iterator must be released after use, by calling Release method. 

https://pkg.go.dev/github.com/syndtr/goleveldb/leveldb#Transaction.NewIterator

I'm not sure if in practice this means we're leaking memory, because goleveldb uses `runtime.SetFinalizer` to call the Release, but we don't want to trust it works correctly, and also goleveldb never promised it's doing that, so they can always stop doing that.

I did not explictly invalidated the iterators, so some iterators could still be "used" after closing, and closing won't free their memory (assigning nil to all fields), but I'm open to discuss that with or without an explicit `isClosed` flag in all iterators.


svarogg (commented):
I think we want to make sure nobody uses iterators after Close, at the very least for the sanity of the thing, and later down the road to maybe release some resources.

Even just setting iterator.isClosed = true, and then erroring out on any function that returns an error can be a good start.

elichai:
> I think we want to make sure nobody uses iterators after Close, at the very least for the sanity of the thing, and later down the road to maybe release some resources.
> 
> Even just setting iterator.isClosed = true, and then erroring out on any function that returns an error can be a good start.

So do we want to add an `isClosed` boolean to all the iterators? (note that if you then call `Next()` we can only return false/true not an error)

svarogg:
> So do we want to add an `isClosed` boolean to all the iterators? 
I believe we do. In addition we might release any resources held, but less critical at this point IMHO.

> (note that if you then call `Next()` we can only return false/true not an error)
Yeah, I thought about that and didn't really come up with a solution I'm 100% satisfied with.
I guess we should just return `false` when calling Next() on a closed iterator.

elichai:
> I guess we should just return `false` when calling Next() on a closed iterator.

If the `Get()` of that iterator returns an error we can return true and then set the error (We already do that in different cases when there's an error inside `Next()`)

svarogg:
> If the `Get()` of that iterator returns an error we can return true and then set the error (We already do that in different cases when there's an error inside `Next()`)

Yeah, I though about this, it indeed will achieve the desired behavior, at the price of correctness (Next returning true while there's no actual Next).
I'm starting to see the merits of this solution in a broader light, so I think I'm convinced. Let's do this as you suggested:
`.Next()` on closed iterator always returns `true`, then caller will call `.Get()`, that always returns an error.

svarogg (approved):

PR #1548, opened 2021-02-22
Fix UTXO index

### LINKED ISSUE #1543: Data loss in UTXO index
There are two scenarios in which data loss in the UTXO index could occur:

* If kaspad crashes/closes after inserting a block but before it was added to the UTXO index, the index would never learn of this UTXO. Meaning that it's possible to lose UTXOs between kaspad runs.
* If kaspad enters IBD and is required to download the pruning point's UTXO set, the UTXO index is never updated. This essentially corrupts the index.

someone235:
Fixes #1543


domain/consensus/processes/blockprocessor/validateandinsertblock.go
```diff
@@ -145,8 +146,15 @@ func (bp *blockProcessor) validateAndInsertBlock(block *externalapi.DomainBlock,
 
 	bp.blockLogger.LogBlock(block)
 
+	virtualParents, err := bp.dagTopologyManager.Parents(model.VirtualBlockHash)
+	if err != nil {
+		return nil, err
+	}
+
```
elichai, line 153: I think we want `LogBlock` to be the last logic thing in ValidateAndInsertBlock

infrastructure/network/netadapter/server/grpcserver/protowire/rpc.proto
```diff
@@ -546,6 +546,40 @@ message VirtualSelectedParentBlueScoreChangedNotificationMessage {
...
+// StopNotifyPruningPointUTXOSetOverrideRequestMessage unregisters this connection for
+// pruning point UTXO set override notifications.
+//
+// This call is only available when this kaspad was started with `--utxoindex`
+//
+// See: PruningPointUTXOSetOverrideNotificationMessage
+message StopNotifyPruningPointUTXOSetOverrideRequestMessage {
+}
+
+message StopNotifyPruningPointUTXOSetOverrideResponseMessage {
```
elichai, line 579: In previous places we've called it `StopNotifying`

infrastructure/network/rpcclient/rpc_on_pruning_point_utxo_set_override.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return err
+	}
+	notifyPruningPointUTXOSetOverrideResponse := response.(*appmessage.NotifyPruningPointUTXOSetOverrideResponseMessage)
+	if notifyPruningPointUTXOSetOverrideResponse.Error != nil {
+		return c.convertRPCError(notifyPruningPointUTXOSetOverrideResponse.Error)
+	}
+	spawn("RegisterPruningPointUTXOSetNotifications", func() {
+		for {
+			_, err := c.route(appmessage.CmdPruningPointUTXOSetOverrideNotificationMessage).Dequeue()
```
elichai, line 27: Should we downcast the message to `PruningPointUTXOSetOverrideNotificationMessage` as a sanity check?

infrastructure/network/rpcclient/rpc_on_pruning_point_utxo_set_override.go
```diff
@@ -0,0 +1,57 @@
...
+	response, err := c.route(appmessage.CmdStopNotifyPruningPointUTXOSetOverrideResponseMessage).DequeueWithTimeout(c.timeout)
+	if err != nil {
+		return err
+	}
+	stopNotifyPruningPointUTXOSetOverrideResponse := response.(*appmessage.StopNotifyPruningPointUTXOSetOverrideResponseMessage)
+	if stopNotifyPruningPointUTXOSetOverrideResponse.Error != nil {
+		return c.convertRPCError(stopNotifyPruningPointUTXOSetOverrideResponse.Error)
+	}
+	return nil
+}
```
elichai, line 57: Thanks! I see that in the other notifications we don't provide this nice helper

testing/integration/utxo_index_test.go
```diff
@@ -70,21 +70,11 @@ func TestUTXOIndex(t *testing.T) {
...
-	mineNextBlock(t, kaspad)
-	notification := <-onUTXOsChangedChan
-	if len(notification.Removed) > 0 {
-		t.Fatalf("Unexpectedly received that a UTXO has been removed")
-	}
-	for _, added := range notification.Added {
-		notificationEntries = append(notificationEntries, added)
-	}
-
-	// Mine another block to accept the above block
```
elichai, line 83: Why did you remove this?

infrastructure/network/netadapter/server/grpcserver/protowire/messages_grpc.pb.go
```diff
@@ -11,7 +11,7 @@ import (
 
 // This is a compile-time assertion to ensure that this generated file
 // is compatible with the grpc package it is being compiled against.
-const _ = grpc.SupportPackageIsVersion7
+const _ = grpc.SupportPackageIsVersion6
```
elichai, line 14: Why did this change?

domain/utxoindex/utxoindex.go
```diff
@@ -20,143 +17,143 @@ type UTXOIndex struct {
 }
 
 // New creates a new UTXO index
-func New(consensus externalapi.Consensus, database database.Database) *UTXOIndex {
-	store := newUTXOIndexStore(database)
-	return &UTXOIndex{
+func New(consensus externalapi.Consensus, database database.Database) (*UTXOIndex, error) {
```
elichai, line 20: We should document that this should run *before* the flows. otherwise the virtual can be changed between `isSynced()` and registering

domain/utxoindex/utxoindex.go
```diff
@@ -20,143 +17,143 @@ type UTXOIndex struct {
...
+		return err
+	}
+
+	var fromOutpoint *externalapi.DomainOutpoint
+	for {
+		const step = 1000
+		virtualUTXOs, err := ui.consensus.GetVirtualUTXOs(virtualInfo.ParentHashes, fromOutpoint, step)
 		if err != nil {
-			return nil, err
+			return err
```
elichai, line 58: Do we want to check the error and restart the loop if the parents changed? or do we assert here that the flows are off and if they're not then we want to crash here?

domain/utxoindex/serialization.go
```diff
@@ -0,0 +1,74 @@
+package utxoindex
```
elichai, line 1: nit, I prefer that these functions will be free(no receiver)

testing/integration/utxo_index_test.go
someone235: Because the notification are now based on the virtual and not on the selected tip, so you don't need to mine extra block to get the expected notification

domain/utxoindex/utxoindex.go
someone235: Yeah, I think I prefer to assert

elichai (approved):

Issue #1417, opened 2021-01-14
Remove deep virtual diff parents


someone235:
In some situations (for example, when you accidentally mine on top of genesis), a block can be in the selected chain and get removed later while still remaining a tip, which means it'll remain a virtual diff parent for a very long time (especially if pointing at it violates finality).
In such situation every block will trigger an update to that virtual diff parent, and every block will increase the diff size. This severely harms the performance of IBD, and makes it very slow.
In order to solve it I suggest removing from the virtual diff parents blocks that have blue score < `virtual.blueScore - 100`, and set some flag that says that the diff child of that block is just a dead end.  And if in the future we'll have a reorg that will need to use the diff of that node, we'll just need to remove its UTXO diffs and resolveBlockStatus while relying on the latest block that has a valid diff-child path.

svarogg:
I'll start working on this.
The solution is very different from what was initially proposed, so lining it up here:

The gist: There’s only 1 virtualDiffParent at any time: the selectedTip.

Practically:
Whenever newBlock is being UTXO-Verified, when setting it’s diffChild:
If newBlock is going to be the new selectedTip:
Set oldSelectedTip.DiffChild = newBlock
newBlock.DiffChild = virtual (as has been done previously)
Else:
Set newBlock.DiffChild = selectedTip

This allows us to get rid of the whole virtualDiffParent mechanisms, simplifying the code tremendously.

There is one downside to this:
In re-orgs, all the blocks in the new selectedChain (excluding the tip) are going to have the oldSelectedTip as their DiffChild, which might incur a (very) large diff.
We agreed that for now this situation is acceptable. We have proposed a few solutions, to be implemented later down the line.

elichai:
This isn't relevant anymore since we have only a single virtual diff parent
https://github.com/kaspanet/kaspad/pull/1550

PR #1550, opened 2021-02-23
Remove virtual diff parents

svarogg:
This fixes https://github.com/kaspanet/kaspad/issues/1417

The gist: There’s only 1 virtualDiffParent at any time: the selectedTip.

Practically:
1. Whenever newBlock is being UTXO-Verified, when setting it’s diffChild:
2. If newBlock is going to be the new selectedTip:
2.1. Set oldSelectedTip.DiffChild = newBlock
2.2. newBlock.DiffChild = virtual (as has been done previously)
3. Else:
3.1. Set newBlock.DiffChild = selectedTip

This allows us to get rid of the whole virtualDiffParent mechanisms, simplifying the code tremendously.

There is one downside to this:
In re-orgs, all the blocks in the new selectedChain (excluding the tip) are going to have the oldSelectedTip as their DiffChild, which might incur a (very) large diff.
For now this situation is acceptable, and much better then previous situation in the average case. There are a few proposed solutions, to be decided upon and implemented later down the line.


domain/consensus/processes/consensusstatemanager/multisets.go
```diff
@@ -38,8 +38,7 @@ func (csm *consensusStateManager) calculateMultiset(
 			isCoinbase := i == 0
 			log.Tracef("Is transaction %s a coinbase transaction: %t", transactionID, isCoinbase)
 
-			var err error
-			err = addTransactionToMultiset(ms, transaction, blockGHOSTDAGData.BlueScore(), isCoinbase)
+			err := addTransactionToMultiset(ms, transaction, blockGHOSTDAGData.BlueScore(), isCoinbase)
```
elichai, line 41: Great Fix! ;)

domain/consensus/processes/consensusstatemanager/resolve_block_status.go
```diff
@@ -152,72 +156,63 @@ func (csm *consensusStateManager) resolveSingleBlockStatus(blockHash *externalap
 	log.Tracef("Staging the multiset of block %s", blockHash)
 	csm.multisetStore.Stage(blockHash, multiset)
 
-	log.Tracef("Staging the utxoDiff of block %s", blockHash)
-	err = csm.stageDiff(blockHash, pastUTXODiff, nil)
-	if err != nil {
-		return 0, err
+	if csm.genesisHash.Equal(blockHash) {
+		log.Tracef("Staging the utxoDiff of block %s", blockHash)
```
elichai, line 160: maybe write "for genesis" instead?

elichai (dismissed):

elichai (approved):

stasatdaglabs (approved):

Issue #1527, opened 2021-02-15
mempool.RemoveTransactions doesn't work properly


elichai:
It calls `removeDoubleSpends` which makes sure it is a double spend (ie has a different TransactionID than what's in the mempool).
that's not the behavior we want, we want it to unconditionally remove that outpoint and its ancestors.

PR #1551, opened 2021-02-23
Fix mempool.RemoveTransactions

svarogg:
This fixes https://github.com/kaspanet/kaspad/issues/1527


domain/miningmanager/mempool/mempool.go
```diff
@@ -986,16 +986,10 @@ func (mp *mempool) HandleNewBlockTransactions(txs []*consensusexternalapi.Domain
...
 	defer mp.mtx.Unlock()
 
-	for _, tx := range txs {
-		err := mp.removeDoubleSpends(tx)
-		if err != nil {
-			log.Infof("Failed removing tx from mempool: %s, '%s'", consensushashing.TransactionID(tx), err)
-		}
-		mp.removeOrphan(tx, true)
-	}
+	return mp.removeBlockTransactionsPool(txs)
```
elichai, line 994: I don't think this should return an error.
it looks like the error comes from `cleanTransactionFromSets->removeTx` which returns an error on missing outpoints.

but if the mempool changed between BuildBlock and here then we shouldn't error the GetBlockTemplate on this.

but I'm not sure on which level we should remove/ignore the error (in blockTemplateBuilder, in here or deeper)

svarogg: I've looked into this a bit deeper and came to the conclusion that current behavior is probably correct.

removeBlockTransactionsFromPool, for each transaction first fetches it's TxDesc, if it's not found - it `continue`s.
The error comes only when fetchTxDesc returned true (a.k.a. `mp.pool` has said transaction), but one of the other sets does not - meaning there's some internal structure failure in mempool - something that should be reported, and on a higher level should probably crash the server.

elichai (approved):

PR #1557, opened 2021-02-25
Fix subnetworks FromString and disable reversal

elichai:
Removed DomainSubnetworkID bytes reversal.
And fixed FromString. (Before it switched each ascii letter in the hex, now it uses FromBytes)

### DIFF
--- a/domain/consensus/utils/subnetworks/from_string.go
+++ b/domain/consensus/utils/subnetworks/from_string.go
@@ -8,11 +8,7 @@ import (
 
 // FromString creates a DomainSubnetworkID from the given byte slice
 func FromString(str string) (*externalapi.DomainSubnetworkID, error) {
-	runes := []rune(str)
-	for i := 0; i < externalapi.DomainSubnetworkIDSize*2; i++ {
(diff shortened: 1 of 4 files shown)


domain/consensus/utils/subnetworks/from_string.go
```diff
@@ -8,13 +8,18 @@ import (
 
 // FromString creates a DomainSubnetworkID from the given byte slice
 func FromString(str string) (*externalapi.DomainSubnetworkID, error) {
-	runes := []rune(str)
-	for i := 0; i < externalapi.DomainSubnetworkIDSize*2; i++ {
-		runes[i], runes[externalapi.DomainSubnetworkIDSize-1-i] = runes[externalapi.DomainSubnetworkIDSize-1-i], runes[i]
+	subnetworkIDBytes, err := hex.DecodeString(str)
+	if err != nil {
+		return nil, err
 	}
```
someone235, line 14: I would first reverse the bytes, and only then call `FromBytes`

someone235 (changes requested):

domain/consensus/utils/subnetworks/from_string.go
elichai: But then I need to check the size and handle it if it's not even.

someone235 (approved):

PR #1592, opened 2021-03-10
Adjust the difficulty in the first difficultyAdjustmentWindowSize blocks

### LINKED ISSUE #1546: RequiredDifficulty wrong if there are less then difficultyAdjustmentWindowSize blue + some red blocks
```
	// Not enough blocks for building a difficulty window.
	if bluestGhostDAG.BlueScore() < uint64(dm.difficultyAdjustmentWindowSize)+1 {
		return dm.genesisBits()
	}
```
^ This was correct back when BlockWindow was only blue blocks.
But now that it also might contain red blocks, this should be revised.

elichai:
Closes #1546

The solution is this: https://github.com/kaspanet/kaspad/issues/1546#issuecomment-793774683
>1. Exclude genesis from the window.
>2. Don't pad the window with genesis.
>3. Divide the target by the current window size (which will be smaller than 2640 for the first 2640 blocks).
>4. We might need to clamp the timestamp difference(max-min) to make it no lower than 1 second(targetTime.MilliSeconds()), depending on how PastMedianTime works for those blocks.


domain/consensus/processes/pastmediantimemanager/pastmediantimemanager.go
```diff
@@ -46,6 +50,13 @@ func (pmtm *pastMedianTimeManager) PastMedianTime(blockHash *externalapi.DomainH
 	if err != nil {
 		return 0, err
 	}
+	if len(window) == 0 {
+		header, err := pmtm.blockHeaderStore.BlockHeader(pmtm.databaseContext, pmtm.genesisHash)
```
someone235, line 54: It's very weird to check the window size and then give the timestamp of a hard coded hash. I would expect to choose between:
a. if window size is 0 -> return blockHash's timetstamp
b. if blockHash == genesis -> return genesis' timetstamp

domain/consensus/processes/blockvalidator/block_body_in_context_test.go
```diff
@@ -154,42 +153,70 @@ func TestIsFinalizedTransaction(t *testing.T) {
...
 			}
 
 			tx.LockTime = lockTime
 			tx.Inputs[0].Sequence = sequence
 
-			_, _, err = tc.AddBlock([]*externalapi.DomainHash{block1Hash}, nil, []*externalapi.DomainTransaction{tx})
+			_, _, err = tc.AddBlock(blockParents, nil, []*externalapi.DomainTransaction{tx})
 			if (shouldPass && err != nil) || (!shouldPass && !errors.Is(err, ruleerrors.ErrUnfinalizedTx)) {
-				t.Fatalf("Unexpected error: %+v", err)
+				t.Fatalf("ShouldPass: %t Unexpected error: %+v", shouldPass, err)
```
someone235, line 206: Why capital S?

someone235 (changes requested):

PR #1598, opened 2021-03-11
Implement BIP-143-like sighash

### LINKED ISSUE #1341: Replace sighash algorithm with BIP143 sighash
BIP143 sighash has a few advantages:
1. Isn't vulnerable to quadratic hashing (not a big deal for small blocks).
2. Uses the amount, allowing cold/hw wallets to easily know and present the correct amount to the user, removing another attack vector against cold/hw wallets.

We should do this only after #1030, because it will also affect sighashes. (need to make sure we include the script version).

Complete spec:
https://github.com/kaspanet/docs/blob/main/Specs/BIP143-like%20SigHashes.md

svarogg:
This PR resolves #1341

Complete spec: https://github.com/kaspanet/docs/blob/main/Specs/BIP143-like%20SigHashes.md


domain/consensus/utils/txscript/sign_test.go
```diff
@@ -391,12 +368,11 @@ func TestSignTxOutput(t *testing.T) {
...
-				mkGetKey(map[string]*secp256k1.SchnorrKeyPair{
-					address.EncodeAddress(): key,
-				}), mkGetScript(map[string][]byte{
-					scriptAddr.EncodeAddress(): scriptPubKey.Script,
-				}), nil); err != nil {
+			if err := signAndCheck(msg, tx, i, scriptScriptPubKey, hashType, mkGetKey(map[string]*secp256k1.SchnorrKeyPair{
+				address.EncodeAddress(): key,
+			}), mkGetScript(map[string][]byte{
+				scriptAddr.EncodeAddress(): scriptPubKey.Script,
+			})); err != nil {
```
stasatdaglabs, line 375: Please reformat the above five lines for readability's sake.

domain/consensus/utils/consensushashing/calculate_signature_hash.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+// SigHashType represents hash type bits at the end of a signature.
+type SigHashType uint32
+
+// Hash type bits from the end of a signature.
+const (
+	SigHashAll          SigHashType = 0x1
+	SigHashNone         SigHashType = 0x2
+	SigHashSingle       SigHashType = 0x3
```
stasatdaglabs, line 19: I think it'll be a tiny bit more clear that this is a bit field if you write it as 0x01.

domain/consensus/utils/consensushashing/calculate_signature_hash.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	// SigHashSingle: If the relevant output exists - return it's hash, otherwise return zero-hash
+	if hashType.isSigHashSingle() {
+		if inputIndex >= len(tx.Outputs) {
+			return externalapi.NewZeroHash()
+		}
+		hashWriter := hashes.NewTransactionSigningHashWriter()
+		hashTxOut(hashWriter, tx.Outputs[inputIndex])
+		return hashWriter.Finalize()
```
stasatdaglabs, line 149: I think you should update reusedValues.outputsHash here and only return later.

stasatdaglabs (changes requested):
.

domain/consensus/utils/consensushashing/calculate_signature_hash.go
```diff
@@ -0,0 +1,196 @@
...
+	return sht&SigHashMask == SigHashAll
+}
+func (sht SigHashType) isSigHashNone() bool {
+	return sht&SigHashMask == SigHashNone
+}
+func (sht SigHashType) isSigHashSingle() bool {
+	return sht&SigHashMask == SigHashSingle
+}
+func (sht SigHashType) isSigHashAnyOneCanPay() bool {
+	return sht&SigHashAnyOneCanPay == SigHashAnyOneCanPay
```
elichai, line 37: I'm not sure this is correct, I think the correct line is `sht&SigHashAnyOneCanPay != 0`
(This seems to be how bitcoin core does it: https://github.com/bitcoin/bitcoin/blob/8e6532053f580003869346510fc1dc20c46c8067/src/script/interpreter.cpp#L1587 and how we did it before: https://github.com/kaspanet/kaspad/blob/master/domain/consensus/utils/txscript/script.go#L358)

domain/consensus/utils/consensushashing/calculate_signature_hash.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	return reusedValues.sequencesHash
+}
+
+func getOutputsHash(tx *externalapi.DomainTransaction, inputIndex int, hashType SigHashType, reusedValues *SighashReusedValues) *externalapi.DomainHash {
+	// SigHashNone: return zero-hash
+	if hashType.isSigHashNone() {
+		return externalapi.NewZeroHash()
+	}
```
elichai, line 140: I think there's a meaningful difference between this and the bitcoin approach.
the difference is how do we handle unknown sighashes, in bitcoin unknown hash types will be all zeros here, and in this code `SigHashAll` will be the default.
I'm not sure which is better and what kinds of validation happen to the sighash before this function (maybe only valid known sighashes can enter this function)

elichai: Same for `getSequencesHash`

domain/consensus/utils/txscript/engine.go
```diff
@@ -358,9 +361,9 @@ func (vm *Engine) currentScript() []parsedOpcode {
 
 // checkHashTypeEncoding returns whether or not the passed hashtype adheres to
 // the strict encoding requirements if enabled.
-func (vm *Engine) checkHashTypeEncoding(hashType SigHashType) error {
-	sigHashType := hashType & ^SigHashAnyOneCanPay
-	if sigHashType < SigHashAll || sigHashType > SigHashSingle {
+func (vm *Engine) checkHashTypeEncoding(hashType consensushashing.SigHashType) error {
+	sigHashType := hashType & ^consensushashing.SigHashAnyOneCanPay
+	if sigHashType < consensushashing.SigHashAll || sigHashType > consensushashing.SigHashSingle {
```
elichai, line 366: Looks correct according to https://github.com/bitcoin/bitcoin/blob/master/src/script/interpreter.cpp#L193-L194

domain/consensus/utils/txscript/engine.go
```diff
@@ -358,9 +361,9 @@ func (vm *Engine) currentScript() []parsedOpcode {
 
 // checkHashTypeEncoding returns whether or not the passed hashtype adheres to
 // the strict encoding requirements if enabled.
-func (vm *Engine) checkHashTypeEncoding(hashType SigHashType) error {
-	sigHashType := hashType & ^SigHashAnyOneCanPay
-	if sigHashType < SigHashAll || sigHashType > SigHashSingle {
+func (vm *Engine) checkHashTypeEncoding(hashType consensushashing.SigHashType) error {
+	sigHashType := hashType & ^consensushashing.SigHashAnyOneCanPay
```
elichai, line 365: (FYI &^ is a single operation in golang, but I think I like the space between them (altough `^` by itself is xor))

domain/consensus/utils/consensushashing/calculate_signature_hash.go
elichai: we could even go full literal: `0b00000001` which makes it easier to see the bits and masks with the naked eye (ie `SigHashMask = 0x1f` becomes: `0b00011111`)

domain/consensus/utils/consensushashing/calculate_signature_hash.go
elichai: this outputsHash isn't cached, because if it's single then each input will have its own outputsHash, you can't cache them between inputs

domain/consensus/utils/consensushashing/calculate_signature_hash_test.go
```diff
@@ -276,3 +279,113 @@ func generateTxs() (nativeTx, subnetworkTx *externalapi.DomainTransaction, err e
...
+		consensushashing.SigHashNone | consensushashing.SigHashAnyOneCanPay,
+		consensushashing.SigHashSingle | consensushashing.SigHashAnyOneCanPay}
+
+	for _, size := range []int{10, 100, 1000} {
+		tx := generateTransaction(b, sigHashTypes, size)
+
+		b.Run(fmt.Sprintf("%d-inputs-and-outputs", size), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				reusedValues := &consensushashing.SighashReusedValues{}
+				sigHashType := sigHashTypes[i%len(sigHashTypes)]
```
elichai, line 298: It's best to not use the benchmark index in the benchmark, because it scales it up without taking divisable by `len(sigHashTypes)` into account so the result will be inconsistent between scales

domain/consensus/utils/consensushashing/calculate_signature_hash.go
svarogg: Both are correct and widely used.
I prefer the more explicit form of `field&flag == flag`.

domain/consensus/utils/consensushashing/calculate_signature_hash.go
svarogg: I've added a validation in the beggining of CalculateSignatureHash that makes sure that the SigHashType passed is a standard one, so we are not going to have a default (it will error out if you pass 0)

domain/consensus/utils/consensushashing/calculate_signature_hash_test.go
svarogg: That's actually a bug, I planned to use inputIndex, not i

domain/consensus/utils/txscript/sign_test.go
```diff
@@ -248,10 +226,9 @@ func TestSignTxOutput(t *testing.T) {
 				t.Errorf("failed to make scriptPubKey "+
 					"for %s: %v", msg, err)
 			}
-			if err := signAndCheck(msg, tx, i, scriptPubKey, hashType,
-				mkGetKey(map[string]*secp256k1.SchnorrKeyPair{
-					address.EncodeAddress(): key,
-				}), mkGetScript(nil), nil); err != nil {
+			if err := signAndCheck(msg, tx, i, scriptPubKey, hashType, mkGetKey(map[string]*secp256k1.SchnorrKeyPair{
+				address.EncodeAddress(): key,
+			}), mkGetScript(nil)); err != nil {
```
stasatdaglabs, line 231: Please reformat these three lines as well.

stasatdaglabs (changes requested):

stasatdaglabs (approved):

domain/consensus/utils/consensushashing/calculate_signature_hash.go
```diff
@@ -14,16 +14,26 @@ type SigHashType uint32
 
 // Hash type bits from the end of a signature.
 const (
-	SigHashAll          SigHashType = 0x1
-	SigHashNone         SigHashType = 0x2
-	SigHashSingle       SigHashType = 0x3
-	SigHashAnyOneCanPay SigHashType = 0x80
+	SigHashAll          SigHashType = 0b00000001
+	SigHashNone         SigHashType = 0b00000010
+	SigHashSingle       SigHashType = 0b00000100
```
elichai, line 19: That's 4, it should be 3

domain/consensus/utils/consensushashing/calculate_signature_hash.go
```diff
@@ -14,16 +14,26 @@ type SigHashType uint32
...
+	SigHashAnyOneCanPay SigHashType = 0b10000000
 
 	// SigHashMask defines the number of bits of the hash type which is used
 	// to identify which outputs are signed.
-	SigHashMask = 0x1f
+	SigHashMask = 0b00000111
```
elichai, line 24: That's 0x07, it should be `0b00011111`(0x1f)

domain/consensus/utils/consensushashing/calculate_signature_hash.go
```diff
@@ -14,16 +14,26 @@ type SigHashType uint32
...
+func (sht SigHashType) isStandardSigHashType() bool {
+	switch sht {
+	case SigHashAll, SigHashNone, SigHashSingle,
+		SigHashAll | SigHashAnyOneCanPay, SigHashNone | SigHashAnyOneCanPay, SigHashSingle | SigHashAnyOneCanPay:
+		return true
+	default:
+		return false
+	}
+}
+
```
elichai, line 36: Are we sure this is good and correct?

domain/consensus/utils/txscript/sign_test.go
```diff
@@ -736,12 +736,12 @@ var sigScriptTests = []tstSigScript{
...
 					ScriptPublicKey: p2pkhScriptPubKey,
 				},
-				sigscriptGenerates: true,
+				sigscriptGenerates: false,
 				inputValidates:     false,
 				indexOutOfRange:    false,
 			},
 		},
-		hashType:           0x04,
+		hashType:           0b00000011,
```
elichai: you meant `0b00000100`?

domain/consensus/utils/serialization/common.go
```diff
@@ -19,7 +19,11 @@ func WriteElement(w io.Writer, element interface{}) error {
 	// type assertions first.
 	switch e := element.(type) {
 	case []byte:
-		_, err := w.Write(e)
+		err := WriteElement(w, len(e))
```
elichai: We don't have support for `int`, so this won't work, can you cast to `uint64`? (or just write it as uint64 here)

domain/consensus/utils/consensushashing/calculate_signature_hash.go
elichai: Hmm you're right haven't thought it through

domain/consensus/utils/consensushashing/calculate_signature_hash.go
svarogg: This is deliberate.
@stasatdaglabs asked me to do this so that SigHashType is a true bitfield.

LMK if you disagree

domain/consensus/utils/consensushashing/calculate_signature_hash.go
svarogg: As sure as one can be.
Many things inside the code and tests of CalculateSignatureHash assume the above is correct. 
If it is not - we should fix both it and the code.

elichai (approved):

stasatdaglabs (approved):

PR #1618, opened 2021-03-17
Use BLAKE2B instead of HASH160, and get rid of any usage of RIPEMD160 and SHA1

### LINKED ISSUE #1023: Remove RIPEMD-160 from any of the standard script
We shouldn't use RIPEMD-160 in any of the standard scripts,
we should also consider completely removing RIPEMD-160 and SHA1 from the script.

someone235:
Closes #1023


util/hash160.go
```diff
@@ -5,19 +5,11 @@
 package util
```
stasatdaglabs, line 5: I think you might want to rename this file.

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #1627, opened 2021-03-24
Add mass limit to mempool

someone235:


domain/miningmanager/factory.go
```diff
@@ -2,20 +2,21 @@ package miningmanager
...
+	NewMiningManager(consensus externalapi.Consensus, blockMaxMass uint64, acceptNonStd bool, params *dagconfig.Params) MiningManager
 }
 
 type factory struct{}
 
 // NewMiningManager instantiate a new mining manager
-func (f *factory) NewMiningManager(consensus externalapi.Consensus, blockMaxMass uint64, acceptNonStd bool) MiningManager {
-	mempool := mempoolpkg.New(consensus, acceptNonStd)
+func (f *factory) NewMiningManager(consensus externalapi.Consensus, blockMaxMass uint64, acceptNonStd bool, params *dagconfig.Params) MiningManager {
+	mempool := mempoolpkg.New(consensus, acceptNonStd, params)
```
stasatdaglabs, line 19: It appears that the only reason why you're passing the entire `params` into the mempool is for MaxMassAcceptedByBlock.
You've already passed MaxMassAcceptedByBlock into NewMiningManager, (it's `blockMaxMass`) so why not just pass that into the mempool instead?

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #1630, opened 2021-03-25
Fixes to stability tests: Move orphans test to simnet + Change fakePublicKeyHash size to 32 bytes

svarogg:


stability-tests/netsync/node.go
```diff
@@ -177,7 +179,7 @@ func useDirOrCreateTemp(dataDir, tempName string) (string, error) {
 }
 
 func mineOnTips(client *rpc.Client) (appmessage.RejectReason, error) {
-	fakePublicKeyHash := make([]byte, 20)
+	fakePublicKeyHash := make([]byte, secp256k1.HashSize)
```
elichai, line 182: I think the correct constant is `blake2b.Size256`, not `secp256k1.HashSize`

elichai: (but just doing `32` is also fine IMHO)

elichai (approved):

PR #1636, opened 2021-03-29
Improve wallet functionality

### LINKED ISSUE #1262: Move the CLI wallet from an example to production ready code
This requires going over the code, fixing edge cases, auditing the code etc.
should be done after #1265 #1264 #1263

### LINKED ISSUE #1263: Add multisig support for the CLI wallet
We want to be able to create K-of-N wallets in the CLI wallet.

### LINKED ISSUE #1264: Add dry sending suport to the CLI wallet
We want to be able to create unsigned transactions via address/public keys only.

### LINKED ISSUE #1265: Add offline signing support to the CLI wallet
We want to enable users to create transactions via public keys only with kaspanet/kaspad#1264 and then sign them in a cold wallet via offline signing

someone235:
Closes #1262
Closes #1263 
Closes #1264 
Closes #1265


cmd/kaspawallet/keys/keys.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+
+	decoder := json.NewDecoder(file)
+	decoder.DisallowUnknownFields()
+	decodedFile := &keysFileJSON{}
```
elichai, line 112: Maybe also make `MinimumSignatures` a string? I'm a little scared about JSON numbers

cmd/kaspawallet/send.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return err
+	}
+
+	tx, err := libkaspawallet.ExtractTransaction(psTx)
+	if err != nil {
+		return err
+	}
+
+	transactionID, err := sendTransaction(client, tx)
```
elichai, line 68: Isn't this an unsigned transactions?
I think you missed the `sign` step?

cmd/kaspawallet/keys/keys.go
elichai: or if you try to encode a big integer, but that's probably only a risk with uint64

cmd/kaspawallet/balance.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"github.com/kaspanet/kaspad/cmd/kaspawallet/keys"
+	"github.com/kaspanet/kaspad/cmd/kaspawallet/libkaspawallet"
+	"github.com/kaspanet/kaspad/infrastructure/network/rpcclient"
+
+	"github.com/kaspanet/kaspad/util"
+)
+
+func balance(conf *balanceConfig) error {
+	client, err := rpcclient.NewRPCClient(conf.RPCServer)
```
elichai, line 13: You should add this everywhere you use the RPCServer, this allows not passing `--rpcserver` to the CLI
```go
	rpcAddress, err := conf.NetParams().NormalizeRPCServerAddress(conf.RPCServer)
	if err != nil {
		return err
	}
```

cmd/kaspawallet/keys/keys.go
```diff
@@ -0,0 +1,245 @@
...
+	exists, err := pathExists(dir)
+	if err != nil {
+		return err
+	}
+
+	if exists {
+		return nil
+	}
+
+	return os.MkdirAll(dir, 0700)
```
elichai, line 156: why 0700? I think 0600 makes more sense

elichai (approved):

elichai (approved):

Issue #1619, opened 2021-03-18
Investigate mempool limits


svarogg:
Mempool should be limited to 1mil transactions, which translates to about 300 megabytes.
However, we saw a node in which (seemingly) mempool transactions took up to 2.3 gigs.

Write a stability-test that:

    Fills the mempool up to the limit
    Keeps adding more transactions, making sure eviction works
    Starts mining blocks to make sure this cleans the mempool eventually.

stasatdaglabs:
Implemented in https://github.com/kaspanet/kaspad/pull/1647

PR #1647, opened 2021-04-04
Implement a stability test for mempool limits

stasatdaglabs:
This implements https://github.com/kaspanet/kaspad/issues/1619


stability-tests/mempool-limits/main.go
```diff
@@ -0,0 +1,110 @@
...
+	common.UseLogger(backendLog, log.Level())
+
+	cfg := activeConfig()
+	if cfg.Profile != "" {
+		profiling.Start(cfg.Profile, log)
+	}
+
+	defer func() {
+		if err := recover(); err != nil {
+			log.Criticalf("mempool-limits failed: %s", err)
```
svarogg, line 30: Instead of "failed" better write "paniced". Basically the same, but conveys info much more accurately.

stability-tests/mempool-limits/transactions.go
```diff
@@ -0,0 +1,186 @@
...
+var (
+	payAddressKeyPair     = decodePayAddressKeyPair()
+	payToPayAddressScript = buildPayToPayAddressScript()
+
+	fundingCoinbaseTransactions = make([]*externalapi.DomainTransaction, fundingCoinbaseTransactionAmount)
+)
+
+func generateFundingCoinbaseTransactions(rpcClient *rpcclient.RPCClient) {
+	// Generate one coinbase transaction for its side effect:
+	// the block containing it accepts the empty genesis coinbase
```
svarogg, line 38: I think the comment is not exactly correct, you are generating 2 coinbases because that's the minimum you need to get a spendable output (ignoring coinbase maturity).
The fact that the first one accepts the genesis coinbase is (1) not precisely correct and (2) not interesting.

svarogg (changes requested):
Please differentiate errors between whether something unexpected happened, or whatever the test is testing failed.

Currently you are panicking in both cases. 
The panicking itself is very annoying to start with, but I'll let it pass since this is a stability test, but I won't accept it that genuine "the ran correctly and found the bug it was designed to find" occasion as a panic.

One way to achieve this is to convert the test to use go's testing framework, and then use `t.Fatalf()`. 
There's an example on how to do this here:
https://github.com/kaspanet/automation/pull/23

stability-tests/mempool-limits/transactions.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func submitAnAmountOfTransactionsToTheMempool(
+	rpcClient *rpcclient.RPCClient, amountToSubmit int, ignoreOrphanRejects bool) {
+
+	log.Infof("Generating %d transactions", amountToSubmit)
+	transactions := make([]*externalapi.DomainTransaction, 0)
+	for len(transactions) < amountToSubmit {
+		var coinbaseTransaction *externalapi.DomainTransaction
+		coinbaseTransaction, fundingCoinbaseTransactions = fundingCoinbaseTransactions[0], fundingCoinbaseTransactions[1:]
```
svarogg, line 59: fundingCoinbaseTransactions should be passed around, not a global.

Otherwise it's really hard to understand where is it coming from, and in case it's not pre-filled correctly, very hard to find what happened.

stability-tests/mempool-limits/transactions.go
```diff
@@ -0,0 +1,186 @@
...
+			if ignoreOrphanRejects && strings.Contains(err.Error(), "orphan") {
+				continue
+			}
+			panic(err)
+		}
+		log.Infof("Submitted %d transactions", i+1)
+	}
+}
+
+func generateCoinbaseTransaction(rpcClient *rpcclient.RPCClient) *externalapi.DomainTransaction {
```
svarogg, line 88: I'd rename to `mineBlockAndGetCoinbaseTransaction` or something.
`generateCoinbaseTransaction` is only 10% of what this function actually does.

stability-tests/mempool-limits/transactions.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	mine.SolveBlock(templateBlock)
+	_, err = rpcClient.SubmitBlock(templateBlock)
+	if err != nil {
+		panic(err)
+	}
+	return templateBlock.Transactions[0]
+}
+
+func generateTransactionsWithLotsOfOutputs(fundingTransaction *externalapi.DomainTransaction) []*externalapi.DomainTransaction {
```
svarogg, line 105: I'm not sure 3 qualify as "lots"

svarogg (changes requested):

stability-tests/mempool-limits/log.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/kaspanet/kaspad/util/panics"
+)
+
+var (
+	backendLog = logger.NewBackend()
+	log        = backendLog.Logger("MPLM")
+	spawn      = panics.GoroutineWrapperFunc(log)
+)
+
+func initLog(logFile, errLogFile string) {
```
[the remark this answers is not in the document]
stasatdaglabs: It does print the log to stdout, yeah.

stability-tests/mempool-limits/transactions.go
```diff
@@ -0,0 +1,193 @@
...
+	templateBlock, err := appmessage.RPCBlockToDomainBlock(getBlockTemplateResponse.Block)
+	if err != nil {
+		t.Fatalf("RPCBlockToDomainBlock: %s", err)
+	}
+	mine.SolveBlock(templateBlock)
+	_, err = rpcClient.SubmitBlock(templateBlock)
+	if err != nil {
+		t.Fatalf("SubmitBlock: %s", err)
+	}
+	return templateBlock.Transactions[0]
```
svarogg, line 106: better use `transactionhelper.CoinbaseTransactionIndex`

svarogg (changes requested):

stability-tests/mempool-limits/transactions.go
```diff
@@ -0,0 +1,193 @@
...
+const (
+	payAddress                       = "kaspasim:qr79e37hxdgkn4xjjmfxvqvayc5gsmsql2660d08u9ej9vnc8lzcywr265u64"
+	payAddressPrivateKey             = "0ec5d7308f65717f3f0c3e4d962d73056c1c255a16593b3989589281b51ad5bc"
+	fundingCoinbaseTransactionAmount = 1000
+	outputsPerTransaction            = 3
+	transactionFee                   = 1000
+	coinbaseMaturity                 = 100
+)
+
+type fundingCoinbaseTransactions struct {
```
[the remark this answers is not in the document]
stasatdaglabs: Added an explanation as to why this is a struct.

svarogg (approved):

PR #1653, opened 2021-04-05
Add a show-address subcommand to kaspawallet

stasatdaglabs:


cmd/kaspawallet/config.go
```diff
@@ -63,6 +64,11 @@ type broadcastConfig struct {
 	config.NetworkFlags
 }
 
+type showAddressConfig struct {
+	KeysFile string `long:"keys-file" short:"f" description:"Keys file location (only if different from default)"`
```
svarogg, line 68: Instead of "(only if different from default)" better to write something like "(default: ~/.kaspawallet/blabla.json)".

svarogg (changes requested):

svarogg (approved):

PR #1673, opened 2021-04-11
Use a separate folder for every network under ~/.kaspawallet

### LINKED ISSUE #1669: Kaspawallet should keep a separate keys.json file per network
Currently, no matter how one runs kaspawallet, the default keys file location is `~/.kaspawallet/keys.json`.
If a user wants to keep separate wallets for e.g. testnet and devnet, he would have to type a non-default path for one of the key files, otherwise the file would be overriden.

This is pretty dangerous, and could easily lead to loss of coins.

I suggest keeping a separate file per network, using the same scheme kaspad uses (e.g. `~/.kaspawallet/kaspa-devnet/keys.json`, `~/.kaspawallet/kaspa-testnet/keys.json`, etc.).

svarogg:
This fixes #1669

### DIFF
--- a/cmd/kaspawallet/keys/keys.go
+++ b/cmd/kaspawallet/keys/keys.go
@@ -6,19 +6,24 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+
+	"github.com/kaspanet/kaspad/domain/dagconfig"
 	"github.com/kaspanet/kaspad/util"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/argon2"
 	"golang.org/x/crypto/chacha20poly1305"
-	"os"
-	"path/filepath"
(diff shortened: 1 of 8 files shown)


cmd/kaspawallet/keys/keys.go
```diff
@@ -157,7 +162,7 @@ func createFileDirectoryIfDoesntExist(path string) error {
 		return nil
 	}
 
-	return os.MkdirAll(dir, 0600)
+	return os.MkdirAll(dir, 0700)
```
elichai, line 165: Why this change?

someone235 (approved):

cmd/kaspawallet/keys/keys.go
elichai: Manually checked and the permissions end up correct:
1. directories need "execution" permissions (to enter).
2. files are still 0600.

elichai (approved):

PR #1674, opened 2021-04-12
Add windows to the CI + fix errors when testing on Windows

svarogg:


infrastructure/network/addressmanager/addressmanager_test.go
```diff
@@ -274,6 +274,7 @@ func TestRestoreAddressManager(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Could not create a database: %s", err)
 	}
+	defer database.Close()
```
elichai, line 277: How did you find this?

elichai (approved):

infrastructure/network/addressmanager/addressmanager_test.go
elichai: Also, is this correct? we return the `addressManager` from this function, and with this defer it will return it with a closed underlying database

svarogg: t.Tempdir() deletes the temp directory after the test is over.
So the test always failed, because it was unable to delete the file, because it was still under use.

I'm not sure why did this not fail under *nix systems. 
My guess is that this is somehow related to the GC. It probably sprang into action very quickly after the function returned, causing the automatic close of the database when the descriptors are collected, while in Windows it took some more time to react, not closing the DB by the time the framework tried to delete the files.

PR #1678, opened 2021-04-13
Add version file to database

### LINKED ISSUE #1347: Add DB version file to the datadir
We should have a version file in the datadir, that we can use to upgrade datadirs in case of:
1. Changing data structures.
2. Changing the serialization.
3. Changing the actual database.

Because of 3 the version shouldn't be saved in the DB.

svarogg:
This PR closes #1347

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -169,15 +169,23 @@ func removeDatabase(cfg *config.Config) error {
 
 func openDB(cfg *config.Config) (database.Database, error) {
 	dbPath := databasePath(cfg)
+
+	doesVersionFileExist, err := checkDatabaseVersion(dbPath)
+	if err != nil {
+		return nil, err
+	}
+
 	log.Infof("Loading database from '%s'", dbPath)
 	db, err := ldb.NewLevelDB(dbPath, leveldbCacheSizeMiB)
 	if err != nil {
 		return nil, err
 	}
 
-	err = checkDatabaseVersion(dbPath)
-	if err != nil {
-		return nil, err
+	if !doesVersionFileExist {
+		err := createDatabaseVersionFile(dbPath)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	return db, nil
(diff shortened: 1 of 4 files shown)


app/app.go
```diff
@@ -182,5 +170,15 @@ func removeDatabase(cfg *config.Config) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = checkDatabaseVersion(dbPath)
+	if err != nil {
+		return nil, err
+	}
+
+	return db, nil
```
elichai, line 183: Why after opening the db?
I think this should be done before (because theoretically we might support multiple different dbs and the error should be more correct)

app/db_version.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+		return err
+	}
+
+	databaseVersion, err := strconv.Atoi(string(versionBytes))
+	if err != nil {
+		return err
+	}
+
+	if databaseVersion != currentDatabaseVersion {
```
elichai, line 28: Maybe change to if `databaseVersion > currentDatabaseVersion`? it's just easier IMHO

app/app.go
svarogg: Hmmmm... it was to avoid checking if the folder exists when creating the file etc. 
But yeah, you are right, we might fail to open because of bad version.

We should do the check before open, and the write after open.

app/db_version.go
svarogg: I disagree, we are going to support some upgrade paths, but not necessarily all of them.
Threrefore, I'd rather be as prohibitive as possible, and then whomever writes the upgrade logic will be able to permit the specific things he wishes to permit.

app/app.go
```diff
@@ -181,6 +169,24 @@ func removeDatabase(cfg *config.Config) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if !doesVersionFileExist {
+		err := createDatabaseVersionFile(dbPath)
+		if err != nil {
+			return nil, err
+		}
+	}
```
elichai, line 189: Is this also to avoid creating the directory?
as it could also crash between `ldb.NewLevelDB` and here, and have a db without a db version file

elichai (approved):

PR #1679, opened 2021-04-13
Disable validate commitment unless explictly requested

elichai:
This check isn't really needed and it's just for sanity (it's also unrecoverable).
so only do it if the user explictly requested, this configuration is hidden because normal users should not use this.

### DIFF
--- a/infrastructure/config/config.go
+++ b/infrastructure/config/config.go
@@ -121,7 +121,7 @@ type Flags struct {
 	MaxUTXOCacheSize                uint64        `long:"maxutxocachesize" description:"Max size of loaded UTXO into ram from the disk in bytes"`
 	UTXOIndex                       bool          `long:"utxoindex" description:"Enable the UTXO index"`
 	IsArchivalNode                  bool          `long:"archival" description:"Run as an archival node: don't delete old block data when moving the pruning point (Warning: heavy disk usage)'"`
-	EnableSanityCheckPruningUTXOSet bool          `long:"enable-sanity-check-pruning-utxo" hidden:"true" description:"Check that the utxo set of the pruning point matches the utxo commitment of that block"`
+	EnableSanityCheckPruningUTXOSet bool          `long:"enable-sanity-check-pruning-utxo" hidden:"true" description:"When moving the pruning point - check that the utxo set matches the utxo commitment"`
 	NetworkFlags
 	ServiceOptions *ServiceOptions
 }
--- a/infrastructure/config/config.go
+++ b/infrastructure/config/config.go
@@ -76,51 +76,52 @@ var RunServiceCommand func(string) error
 //
 // See loadConfig for details on the configuration load process.
 type Flags struct {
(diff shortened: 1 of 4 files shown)


infrastructure/config/config.go
```diff
@@ -76,51 +76,52 @@ var RunServiceCommand func(string) error
...
+	BlocksOnly                      bool          `long:"blocksonly" description:"Do not accept transactions from remote peers."`
+	RelayNonStd                     bool          `long:"relaynonstd" description:"Relay non-standard transactions regardless of the default settings for the active network."`
+	RejectNonStd                    bool          `long:"rejectnonstd" description:"Reject non-standard transactions regardless of the default settings for the active network."`
+	ResetDatabase                   bool          `long:"reset-db" description:"Reset database before starting node. It's needed when switching between subnetworks."`
+	MaxUTXOCacheSize                uint64        `long:"maxutxocachesize" description:"Max size of loaded UTXO into ram from the disk in bytes"`
+	UTXOIndex                       bool          `long:"utxoindex" description:"Enable the UTXO index"`
+	IsArchivalNode                  bool          `long:"archival" description:"Run as an archival node: don't delete old block data when moving the pruning point (Warning: heavy disk usage)'"`
+	EnableSanityCheckPruningUTXOSet bool          `long:"enable-sanity-check-pruning-utxo" hidden:"true"`
```
svarogg, line 124: Please add a description even though this is never shown in help, for the sake of whomever is reading this file

infrastructure/config/config.go
```diff
@@ -577,6 +578,8 @@ func LoadConfig() (*Config, error) {
 		log.Warnf("%s", configFileError)
 	}
 
+	cfg.ActiveNetParams.EnableSanityCheckPruningUTXOSet = cfg.EnableSanityCheckPruningUTXOSet
```
svarogg, line 581: This isn't really a NetParam.
I assume you did this so that it's easier to pass the thing into consensus, however, this should be treated the same way we treat isArchivalNode.

Seeing the slippery slope here, I suggest passing all the config to consensus, and let it pick out of it the parts it needs.

svarogg (changes requested):

infrastructure/config/config.go
elichai: Theoretically it does "change" consensus behavior, but I get what you're saying.
Assuming that `ActiveNetParams` means network specific configurations, and that `config.Config` or `config.Flags` contains a lot more stuff than consensus will ever need,
maybe we should add another struct in between? `ConsensusConfig` or something. that will contain `ActiveNetParams` and only flags that consensus actually care about.
That way it will also be easier in tests when passing a config to consensus to know that changing something in the config will actually affect consensus and not just be something that seems related to consensus but is actually totally unrelated

svarogg: What you suggest sounds like a good idea, but I suggest we do this in a separate PR.

Meanwhile, treat EnableSanityCheckPruningUTXOSet same way we treat isArchivalNode and pass it separately.

svarogg (approved):

Issue #1670, opened 2021-04-11
Add a stability-test that checks what happens when there are 10,000 tips


svarogg:
Write a stability-test that:

1. Mines a chain of 1k blocks
2. Mines on top of said chain 10k tips
3. Starts mining as BuildBlock will suggest
4. Checks how much time it takes to reach a state where we have a single tip

Fix any issues that arise from said stability-test

PR #1694, opened 2021-05-02
New stability test - many tips

talelbaz:
#1670 
Add a stability test that checks what happens when there are 10,000 tips.
Test result: the number of tips only increases when mining( running on devnet -> there is proof of work).
The scenario of the test:
1. Mine block that its timestamp is one second after the genesis timestamp.
2. Mine block at the current time (get a long time between both blocks to get low difficulty.)
3. Mine on top of it 10k tips.
4. Mines until the DAG will have only one tip.
* This issue can be caused after pruning and IBD.
* Since it's currently unfixed, this test is not part of "run-fast" / "run-slow" scripts, and needs to be run manually.


stasatdaglabs (changes requested):

stability-tests/many-tips/main.go
```diff
@@ -271,13 +255,23 @@ func mineLoopUntilHavingOnlyOneTipInDAG(rpcAddress string, miningAddress util.Ad
...
 		}
 	}
+	spawn("ChecksIfTimeIsUp", func() {
+		timer := time.NewTimer(20 * time.Minute)
+		<-timer.C
+		hasTimedOut = true
+	})
+
+	if hasTimedOut {
+		return errors.Errorf("Out of time - the graph still has more than one tip.")
```
stasatdaglabs, line 273: This and the check in line 203 seem to do the same thing. Please remove the latter.

stability-tests/many-tips/main.go
```diff
@@ -271,13 +255,23 @@ func mineLoopUntilHavingOnlyOneTipInDAG(rpcAddress string, miningAddress util.Ad
...
 			return errors.Wrapf(err, "Error in getCurrentTipsLength")
 		}
 	}
+	spawn("ChecksIfTimeIsUp", func() {
+		timer := time.NewTimer(20 * time.Minute)
+		<-timer.C
+		hasTimedOut = true
+	})
+
+	if hasTimedOut {
```
stasatdaglabs, line 272: hasTimedOut will never actually be true, since the routine is not spawned until _after_ the for loop.

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #1716, opened 2021-05-13
Fix getBlock and getBlocks RPC commands to return blocks and transactions properly

someone235:


domain/consensus/consensus.go
```diff
@@ -112,6 +112,30 @@ func (s *consensus) GetBlock(blockHash *externalapi.DomainHash) (*externalapi.Do
...
+	defer s.lock.Unlock()
+
+	stagingArea := model.NewStagingArea()
+
+	block, err := s.blockStore.Block(s.databaseContext, stagingArea, blockHash)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			header, err := s.blockHeaderStore.BlockHeader(s.databaseContext, stagingArea, blockHash)
+			if err != nil {
+				if errors.Is(err, database.ErrNotFound) {
```
stasatdaglabs, line 126: Instead of building a pyramid, please exit out of the function when `err` is not ErrNotFound.

stasatdaglabs (changes requested):

stasatdaglabs (approved):

PR #1717, opened 2021-05-13
Rename IncludeTransactionVerboseData flag to IncludeTransactions

someone235:
This depends on #1716


stasatdaglabs (changes requested):
Please regenerate rpc.md as well.

stasatdaglabs (approved):

PR #1738, opened 2021-06-03
Bad PR - Get connections from Seeder when no connections left

### LINKED ISSUE #1588: If kaspad runs out of addresses - no new ones will be ever added
If, for any reason, AddressManager ran out of addresses, then ConnectionManager.checkOutgoingConnections would get 0 addresses from addressManager.RandomAddresses, and will simply do nothing.

Currently we request addresses from DNSSeeder only on start, so it is no help either.

Things we can do at such situations:
1. Send GetAddresses to all peers (if any such exist)
2. Request addresses from DNSSeeder.

constantineatdaglabs:


infrastructure/network/connmanager/connmanager.go
```diff
@@ -119,6 +121,23 @@ func (c *ConnectionManager) connectionsLoop() {
 
 		c.checkIncomingConnections(connSet)
 
+		if len(connSet) == 0 && !cfg.DisableDNSSeed {
```
stasatdaglabs, line 124: Checking `connSet` (or `connections`) is a bit problematic, since that includes connections that are currently being initiated (and may fail without ever successfully connecting). Also note that DNS-seeding is not concerned with incoming connections at all, and `connSet` does contain those.
It's better to check whether `ConnectionManager.activeOutgoing` is empty instead.

infrastructure/network/connmanager/connmanager.go
```diff
@@ -119,6 +121,23 @@ func (c *ConnectionManager) connectionsLoop() {
...
+					// source. So we'll take first returned address as source.
+					_ = c.addressManager.AddAddresses(addresses...)
+				})
+
+			dnsseed.SeedFromGRPC(cfg.NetParams(), cfg.GRPCSeed, false, nil,
+				func(addresses []*appmessage.NetAddress) {
+					_ = c.addressManager.AddAddresses(addresses...)
+				})
+		}
+
```
stasatdaglabs, line 140: Please extract the above to a separate method.

stasatdaglabs (changes requested):
Please write a more descriptive PR title. It should at the very least explain what you have done

constantineatdaglabs:
Replaced by https://github.com/kaspanet/kaspad/pull/1742

PR #1740, opened 2021-06-07
Added comment to database open error

constantineatdaglabs:


app/app.go
```diff
@@ -101,7 +101,7 @@ func (app *kaspadApp) main(startedChan chan<- struct{}) error {
 	// Open the database
 	databaseContext, err := openDB(app.cfg)
 	if err != nil {
-		log.Error(err)
+		log.Error("Loading database failed:", err)
```
svarogg, line 104: Please use `log.Errorf` with `%+v` as the format instruction (this will print the error including the stack trace)

svarogg (changes requested):

constantineatdaglabs:
Replaced by https://github.com/kaspanet/kaspad/pull/1746

Issue #1589, opened 2021-03-09
Starting with wrong genesis creates weird errors


svarogg:
### To replicate:
1. Start and sync a node with --testnet
2. Stop node
3. Copy ~/.kaspad/data/kaspa-testnet-2 (or whatever directory testnet currently is in) to ~/.kaspad/data/kaspa-devnet
4. Start node with --devnet

### What happens:
```
2021-03-09 09:31:27.246 [ERR] KASD: Unable to start kaspad: ErrPruningPointViolation
expected pruning point 15fed03a1658cff532648ed9f4b395098d1fca7d57b563dee23c75755bac5726 to be in block 3066aa51c01f8a8d6e09657e243b9eb00d89a7216467bd4d767467e4df169a93 past.
github.com/kaspanet/kaspad/domain/consensus/processes/blockvalidator.(*blockValidator).checkPruningPointViolation
        /home/mike/dev/go/src/github.com/kaspanet/kaspad/domain/consensus/processes/blockvalidator/pruning_violation_proof_of_work_and_difficulty.go:158
github.com/kaspanet/kaspad/domain/consensus/processes/blockvalidator.(*blockValidator).ValidatePruningPointViolationAndProofOfWorkAndDifficulty
        /home/mike/dev/go/src/github.com/kaspanet/kaspad/domain/consensus/processes/blockvalidator/pruning_violation_proof_of_work_and_difficulty.go:33
github.com/kaspanet/kaspad/domain/consensus/processes/blockprocessor.(*blockProcessor).validateBlock
        /home/mike/dev/go/src/github.com/kaspanet/kaspad/domain/consensus/processes/blockprocessor/validateblock.go:45
github.com/kaspanet/kaspad/domain/consensus/processes/blockprocessor.(*blockProcessor).validateAndInsertBlock
        /home/mike/dev/go/src/github.com/kaspanet/kaspad/domain/consensus/processes/blockprocessor/validateandinsertblock.go:59
github.com/kaspanet/kaspad/domain/consensus/processes/blockprocessor.(*blockProcessor).ValidateAndInsertBlock
        /home/mike/dev/go/src/github.com/kaspanet/kaspad/domain/consensus/processes/blockprocessor/blockprocessor.go:141
github.com/kaspanet/kaspad/domain/consensus.(*consensus).ValidateAndInsertBlock
        /home/mike/dev/go/src/github.com/kaspanet/kaspad/domain/consensus/consensus.go:68
[...]

PR #1745, opened 2021-06-09
Added a friendly error message when genesis hash is not relevant to database

constantineatdaglabs:
Fixes https://github.com/kaspanet/kaspad/issues/1589


domain/consensus/factory.go
```diff
@@ -418,6 +420,9 @@ func (f *factory) NewConsensus(config *Config, db infrastructuredatabase.Databas
 	}
 
 	if !genesisInfo.Exists {
+		if c.blockStore.Count(model.NewStagingArea()) > 0 {
+			return nil, errors.Errorf("Genesis hash is not relevant to database: wrong config or appdir?")
```
svarogg, line 424: I'd be more implicit with the problem... something like:
"Expected genesis block {genesis hash} for {Params.name} is not found in the DAG, while the DAG is not empty".

Also note: you are using errors.Errorf(), with no formatting token in it. In such cases better use errors.New() (but since what I said earlier, there's going to be format tokens now.)

svarogg (changes requested):

domain/consensus/factory.go
```diff
@@ -418,6 +420,9 @@ func (f *factory) NewConsensus(config *Config, db infrastructuredatabase.Databas
 	}
 
 	if !genesisInfo.Exists {
+		if c.blockStore.Count(model.NewStagingArea()) > 0 {
+			return nil, errors.Errorf("expected genesis block %s is not found in the non-empty DAG \"%s\": wrong config or appdir?", genesisHash, config.Params.Name)
```
svarogg, line 424: Please break down line to multiple lines, so that it fits in the screen.

svarogg (changes requested):

svarogg (approved):

Issue #1734, opened 2021-05-30
Use a protocol that skips downloading headers and starts by syncing from the pruning point


someone235:
The node should just ask its peers what is the pruning point with the most blue work, and download its antipast.
As a first step the node should not validate that the pruning point actually has the most blue work. The validation part will be implemented after https://github.com/kaspanet/research/issues/3 is specked

someone235:
Beside the pruning point antipast, the syncer should find the intersection point `I` between `P` and `P.Anticone`, and send the following data:

1. The block header for `I` and each block in its DAA block window.
2. `I.DAAScore`
3. `I.ghostdagData` and ghostdagData for the last `k` blocks in `I.selectedChain`
4. `I.future` headers

The syncer should manually set 1-3 in its data stores, and then insert `I.future` headers like any other headers. We can optimize it later by skipping header verification for blocks inside `P.Past ∪ P.Anticone.Past` and making the syncer provide 1-3 to `{P} ∪ P.Anticone`.

someone235:
One problem that we have when pruning headers is that when we're using time-based relative time locks we can't determine the timestamp of the block where the UTXO was created if we prune its header. I suggest solving it by cancelling the functionality of time-based relative time locks, and leave only DAA score relative time locks

svarogg:
> One problem that we have when pruning headers is that when we're using time-based relative time locks we can't determine the timestamp of the block where the UTXO was created if we prune its header. I suggest solving it by cancelling the functionality of time-based relative time locks, and leave only DAA score relative time locks

I think this is somewhat presumptuous to remove such a basic feature.
Instead, I suggest including the timestamp inside the UTXOEntry. 
It will be somewhat complicated, since all utxo-diff comparisons would have to take both DAAScore and timestamp, but it's not going to be the end of the world IMHO.

someone235:
> I think this is somewhat presumptuous to remove such a basic feature.
> Instead, I suggest including the timestamp inside the UTXOEntry.
> It will be somewhat complicated, since all utxo-diff comparisons would have to take both DAAScore and timestamp, but it's not going to be the end of the world IMHO.

It's a good idea, but we just need to remember it means 8 extra bytes per UTXO

hashdag:
I vote for cancelling this functionality, let's ship the MVP

someone235:
Once IBD is triggered (from an orphan block as usual), the flow will look like this:
[...]

PR #1750, opened 2021-06-13
Add database prefix

someone235:
This is part of #1734

### DIFF
--- a/domain/consensus/datastructures/blockstatusstore/block_status_store.go
+++ b/domain/consensus/datastructures/blockstatusstore/block_status_store.go
@@ -7,6 +7,7 @@ import (
 	"github.com/kaspanet/kaspad/domain/consensus/model"
 	"github.com/kaspanet/kaspad/domain/consensus/model/externalapi"
 	"github.com/kaspanet/kaspad/domain/consensus/utils/lrucache"
+	"github.com/kaspanet/kaspad/domain/prefixmanager"
 )
 
 var bucketName = []byte("block-statuses")
@@ -18,10 +19,10 @@ type blockStatusStore struct {
 }
 
 // New instantiates a new BlockStatusStore
-func New(prefix byte, cacheSize int, preallocate bool) model.BlockStatusStore {
+func New(prefix *prefixmanager.Prefix, cacheSize int, preallocate bool) model.BlockStatusStore {
 	return &blockStatusStore{
 		cache:  lrucache.New(cacheSize, preallocate),
-		bucket: database.MakeBucket([]byte{prefix}).Bucket(bucketName),
+		bucket: database.MakeBucket(prefix.Serialize()).Bucket(bucketName),
 	}
 }
 
--- a/domain/consensus/datastructures/blockstatusstore/block_status_store.go
+++ b/domain/consensus/datastructures/blockstatusstore/block_status_store.go
@@ -7,7 +7,7 @@ import (
 	"github.com/kaspanet/kaspad/domain/consensus/model"
 	"github.com/kaspanet/kaspad/domain/consensus/model/externalapi"
 	"github.com/kaspanet/kaspad/domain/consensus/utils/lrucache"
-	"github.com/kaspanet/kaspad/domain/prefixmanager"
+	"github.com/kaspanet/kaspad/domain/prefixmanager/prefix"
 )
 
 var bucketName = []byte("block-statuses")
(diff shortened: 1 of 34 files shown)


domain/consensus/datastructures/blockstatusstore/block_status_store.go
```diff
@@ -9,17 +9,19 @@ import (
...
 // blockStatusStore represents a store of BlockStatuses
 type blockStatusStore struct {
-	cache *lrucache.LRUCache
+	cache  *lrucache.LRUCache
+	bucket model.DBBucket
 }
 
 // New instantiates a new BlockStatusStore
-func New(cacheSize int, preallocate bool) model.BlockStatusStore {
+func New(prefix byte, cacheSize int, preallocate bool) model.BlockStatusStore {
```
svarogg, line 21: I'd rather if all the `New`s got prefix as byte array already (or if you want to go an extra mile - new type called DBPrefix which is alias to byte array - just make sure MakeBucket is able to get it).

The one who should make sure that the prefix is 1-byte long is the place which creates the prefixes.

svarogg, line 21: I'd rather if all the `New`s got prefix as byte array already (or if you want to go an extra mile - new type called DBPrefix which is alias to byte array - just make sure MakeBucket is able to get it).

The one who should make sure that the prefix is 1-byte long is the place which creates the prefixes.

EDIT: even better idea: DBPrefix should be alias to [1]byte, and have a `.Slice()` function.

domain/domain.go
```diff
@@ -4,32 +4,169 @@ import (
...
 }
 
+func (d *domain) CreateTemporaryConsensus() error {
+	_, hasInactivePrefix, err := prefixmanager.InactivePrefix(d.db)
+	if err != nil {
+		return err
+	}
+
+	if hasInactivePrefix {
+		return errors.Errorf("cannot have more than one inactive prefix")
```
svarogg, line 49: I'd revise to "Cannot create temporary consensus when a temporary consensus already exists".
Having an inactive prefix is an implementation detail.

domain/domain.go
```diff
@@ -4,32 +4,169 @@ import (
...
+	if activePrefix == 0 {
+		inactivePrefix = 1
+	}
+
+	err = prefixmanager.SetPrefixAsInactive(d.db, inactivePrefix)
+	if err != nil {
+		return err
+	}
+
+	consensusFactory := consensus.NewFactory()
```
svarogg, line 72: Not sure this is very important, but maybe we can have one factory for every time?

domain/domain.go
```diff
@@ -4,32 +4,169 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	d.tempConsensus = &consensusInstance
+	return nil
+}
+
+func (d *domain) CommitTemporaryConsensus() error {
+	transaction, err := d.db.Begin()
```
svarogg, line 83: Please rename to dbTx or something, transaction confuses with kaspa transaction

domain/domain.go
```diff
@@ -4,32 +4,169 @@ import (
...
+	}
+
+	err = transaction.Commit()
+	if err != nil {
+		return err
+	}
+
+	// We delete anything associated with the old prefix outside
+	// of the transaction in order to save memory.
+	err = prefixmanager.DeleteInactivePrefix(d.db)
```
svarogg, line 125: Did we benchmark how much time this takes for a big database?
If it's too much, maybe better to do it in separate thread?

svarogg, line 125: Also, if we crash in the middle, we have half a prefix.
I think maybe we want in CreateTemporaryConsensus to call DeleteInactivePrefix(P) right before creating P, just in case there's something there.

EDIT: just read that in domain.New you are doing DeleteInactivePrefix, so that might be sufficient. Leaving this comment just to make sure.

svarogg (changes requested):

domain/domain.go
someone235: IMO keeping the factory as part of domain state adds unnecessary complexity with no real gain

domain/domain.go
someone235: It's about 50 seconds for 3.2GB database

svarogg (dismissed):

stasatdaglabs (changes requested):

stasatdaglabs (approved):

domain/prefixmanager/prefix/prefix.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// Equal returns whether p equals to other
+func (p *Prefix) Equal(other *Prefix) bool {
+	return p.value == other.value
+}
+
+// Flip returns the opposite of the current prefix
+func (p *Prefix) Flip() *Prefix {
+	value := prefixZero
```
svarogg, line 27: Optional: All this can be written in one line 
```
return &Prefix {value: 1-p.value }
```

svarogg (changes requested):

svarogg (dismissed):

domain/prefixmanager/prefix/prefix.go
someone235: I think it's clearer the way it is

svarogg (approved):

stasatdaglabs (approved):

Issue #1290, opened 2020-12-27
Stress tests for DAA


FestinaLente666:
We need a platform that would allow us to simulate various scenarios of hash rate change and test the DAA against them.

That is, we need to be able, given a function h(t), we need to be able to simulate a network in which the total hash rate at time t is approx. h(t). (@svarogg has already created such a simulation for h(t) = c.)

This platform should have the following desiderata:
- It could simulate the given hash rate
- It could report the actual hash rate and the corresponding block rate, averaged over windows of approx. 10 seconds
- It could introduce random noise to the hash rate from a given random variable (or at least from a normal distribution with a any desired variance, though some biased noise sources would probably afford some more insight)

Once such a system is operational, the DAA could be stress tested in several ways, including:
- Constant exponential increase in hash rate: h(t) = e^{ct} for some c>0
- Sudden increase/decrease: for some parameters c_1, c_2, t_0 we have h(t) = c_1 for t<t_0 and c_2 for t>t_0. This has a lot of variations, including deliberately choosing t_0 so small that the DAA didn't have time to stabilize
- Limited peak: for some c_1, c_2, t_0, t_1 we have h(t) = c_2 for t_0 < t < t_1 and c_1 otherwise. Would be interesting to see how the network responds to small t_1 - t_0 and large c_2 - c_1
- Noisy constant hash rate
- Periodic peaks: simulating an adversary with a lot of hash rate (say 20%) which constantly turns their machines on and off in various patterns, e.g. turns them on for 5 minutes every hour or so.

The goal of the stress tests are twofold. First, we want to determine if the current DAA eventually stabilizes to the desired blockrate (or to an acceptable deviance therefrom, e.g. slightly higher blockrate in light of constantly increasing hashrate) under the various scenarios. Second, it could indicate that we could recalibrate the DAA parameters:

The current parametrization of the DAA parameters defaultDifficultyAdjustmentWindowSize, defaultTimestampDeviationTolerance were chosen to ensure that no honest block would be delayed (assuming a maximal clock drift of 2 minutes), but that a <50% attacker could not manipulate the difficulty by more than 10%. If we find that the algorithm is very responsive, we could parametrize it to only allow difficulty manipulation of say 5% while still maintaining good responsiveness (the calculation is available [here](https://research.kas.pa/t/handling-timestamp-manipulations/97)).

elichai:
This will probably take a minimum of 1 week

hashdag:
Post mainnet

[...]

PR #1767, opened 2021-06-24
Implement a stability test to stress test the DAA algorithm

stasatdaglabs:
Closes https://github.com/kaspanet/kaspad/issues/1290


stability-tests/daa/daa_test.go
```diff
@@ -0,0 +1,340 @@
...
+			// takes a few hundred nanoseconds, so we spin in place instead of e.g. calling time.Sleep()
+			for {
+				targetHashNanoseconds := targetHashNanosecondsFunction(time.Since(startTime))
+				hashElapsedTimeNanoseconds := time.Since(hashStartTime).Nanoseconds()
+				if hashElapsedTimeNanoseconds >= targetHashNanoseconds {
+					break
+				}
+			}
+
+			hashDuration := time.Since(hashStartTime)
```
svarogg, line 222: Wouldn't this always be more-or-less equal to targetHashNanoseconds?
I think you should run this before the loop.

(Also not very sure why are you measuring this, but guess I'll find out later down the PR).

stability-tests/daa/daa_test.go
```diff
@@ -0,0 +1,340 @@
...
+			hashDurations = append(hashDurations, hashDuration)
```
svarogg, line 223: Not sure if relevant or not, but this append might affect the durations of things, since it will be copying 100k items once it reaches capacity.

stability-tests/daa/daa_test.go
```diff
@@ -0,0 +1,318 @@
...
+	if err != nil {
+		t.Fatalf("NewRPCClient: %s", err)
+	}
+
+	var hashDurations []time.Duration
+	var miningDurations []time.Duration
+
+	startTime := time.Now()
+	runForDuration(runDuration, func() {
+		getBlockTemplateResponse, err := rpcClient.GetBlockTemplate(miningAddress)
```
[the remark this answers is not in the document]
svarogg: And in general, break this monsterfunction into subroutines, very hard to read it as a single piece.

stability-tests/daa/daa_test.go
```diff
@@ -0,0 +1,340 @@
...
+		for !isFinished {
+			runFunction(&isFinished)
+		}
+	}()
+	time.Sleep(duration)
+	isFinished = true
+}
+
+func runKaspad(t *testing.T) func() {
+	appDir, err := common.TempDir("kaspad-daa-test")
```
svarogg, line 291: Don't we already have some runKaspad function in other stability tests?

Maybe move this to common?

stability-tests/daa/daa_test.go
```diff
@@ -0,0 +1,340 @@
...
+	if err != nil {
+		t.Fatalf("TempDir: %s", err)
+	}
+
+	kaspadRunCommand, err := common.StartCmd("KASPAD",
+		"kaspad",
+		common.NetworkCliArgumentFromNetParams(&dagconfig.DevnetParams),
+		"--appdir", appDir,
+		"--logdir", appDir,
```
svarogg, line 300: No need to specify logdir if you want it inside appDir.
What current config will do is a weird structure where you have 
```
appDir/
--- > kaspa-devnet
--- > --- > data
--- > kaspad.log
```

svarogg (changes requested):

stability-tests/daa/daa_test.go
stasatdaglabs: It would, but note that if it reaches `averageHashRateSampleSize` the next couple of lines pop the oldest entry.
Also note that the capacity is defined as `averageHashRateSampleSize + 1`

stability-tests/daa/daa_test.go
stasatdaglabs: `hashDurations` is for logging+debugging only. That is to say, I collect a sample size of 100k durations and print their average every time a block is found.

`targetHashNanoseconds` is the *lower bound* of a `hash duration`, and if, for whatever reason, `hashDuration` is much greater than `targetHashNanoseconds` I want to see it.

stability-tests/daa/daa_test.go
```diff
@@ -0,0 +1,340 @@
...
+			headerForMining.SetNonce(i)
+			if pow.CheckProofOfWorkWithTarget(headerForMining, targetDifficulty) {
+				templateBlock.Header = headerForMining.ToImmutable()
+				break
+			}
+
+			// Yielding a thread in Go takes up to a few milliseconds whereas hashing once
+			// takes a few hundred nanoseconds, so we spin in place instead of e.g. calling time.Sleep()
+			for {
+				targetHashNanoseconds := targetHashNanosecondsFunction(time.Since(startTime))
```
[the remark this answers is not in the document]
stasatdaglabs: `time.Since(startTime)` is different in each loop. I don't think it's possible to move it outside of it.

stability-tests/daa/daa_test.go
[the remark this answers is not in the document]
stasatdaglabs: > I think better to move all the setup to outside runForDuration.

This isn't really setup, but rather part of mining. I've renamed runForDuration to loopForDuration to make it more clear as to what's going on.

stability-tests/daa/daa_test.go
```diff
@@ -158,154 +163,186 @@ func measureMachineHashNanoseconds(t *testing.T) int64 {
...
+	hashDurations = append(hashDurations, hashDuration)
+	if len(hashDurations) > averageHashRateSampleSize {
+		hashDurations = hashDurations[1:]
+	}
+	return hashDurations
 }
 
-func runKaspad(t *testing.T) func() {
-	dataDir, err := common.TempDir("kaspad-daa-test")
+func appendMiningDuration(miningDurations []time.Duration, miningDuration time.Duration) []time.Duration {
```
svarogg, line 299: Consider combining appendHashDuration and appendMiningDuration, they are functionally the same.

Maybe create a type `durationsSlice` or whatever, that has a `maxItems` and an `append` and `average` methods.

svarogg (changes requested):

stability-tests/common/run-kaspad.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"github.com/kaspanet/kaspad/domain/dagconfig"
+	"os"
+	"sync/atomic"
+	"syscall"
+	"testing"
+)
+
+// RunKaspadForTesting runs kaspad for testing purposes
+func RunKaspadForTesting(t *testing.T, testName string, rpcAddress string) func() {
```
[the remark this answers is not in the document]
stasatdaglabs: This `runKaspad` is different from the others in that it relies on testing.T.
Going forward, we should probably be writing tests using the test framework. Right now, though, converting all of them is a bit out of the scope of this ticket.

svarogg (approved):

PR #1777, opened 2021-07-04
Adds name to route, and writes it in every error message 

svarogg:


app/rpc/rpc.go
```diff
@@ -53,7 +53,7 @@ func (m *Manager) routerInitializer(router *router.Router, netConnection *netada
 	for messageType := range handlers {
 		messageTypes = append(messageTypes, messageType)
 	}
-	incomingRoute, err := router.AddIncomingRoute(messageTypes)
+	incomingRoute, err := router.AddIncomingRoute("router", messageTypes)
```
stasatdaglabs, line 56: This should probably be "rpc - incoming" rather than "router"

infrastructure/network/netadapter/router/router.go
```diff
@@ -45,7 +46,7 @@ func (r *Router) AddIncomingRoute(messageTypes []appmessage.MessageCommand) (*Ro
 // AddIncomingRouteWithCapacity registers the messages of types `messageTypes` to
 // be routed to the given `route` with a capacity of `capacity`
 func (r *Router) AddIncomingRouteWithCapacity(capacity int, messageTypes []appmessage.MessageCommand) (*Route, error) {
-	route := newRouteWithCapacity(capacity)
+	route := newRouteWithCapacity("", capacity)
```
stasatdaglabs, line 49: AddIncomingRouteWithCapacity should also accept a name and pass it to newRouteWithCapacity.

infrastructure/network/rpcclient/rpcrouter.go
```diff
@@ -14,7 +14,7 @@ func buildRPCRouter() (*rpcRouter, error) {
 	router := routerpkg.NewRouter()
 	routes := make(map[appmessage.MessageCommand]*routerpkg.Route, len(appmessage.RPCMessageCommandToString))
 	for messageType := range appmessage.RPCMessageCommandToString {
-		route, err := router.AddIncomingRoute([]appmessage.MessageCommand{messageType})
+		route, err := router.AddIncomingRoute("rpc", []appmessage.MessageCommand{messageType})
```
stasatdaglabs, line 17: Maybe call it "rpc client" instead to differentiate it from the server.

stasatdaglabs (changes requested):
Please remove the `590bca9` from the PR title

app/rpc/rpc.go
svarogg: Changed to just "rpc router", the " - incoming" is added by AddIncomingRoute

stasatdaglabs (approved):

PR #1790, opened 2021-07-07
Limit block mass instead of merge set limit + Introduce  SigOpCount to TransactionInput

### LINKED ISSUE #1693: Rethink mass numbers
1) Make MaxMassAcceptedByBlock a number that cannot be included in one block (otherwise when blocks are full most of the merge set transactions will be filtered)
2) Rethink defaultMassPerSigOp to allow more than 50 inputs transactions.

svarogg:
This PR fixes #1693 .
See the issue for more details.


domain/consensus/processes/transactionvalidator/mass.go
```diff
@@ -20,29 +18,30 @@ func (v *transactionValidator) transactionMassStandalonePart(tx *externalapi.Dom
...
-		isP2SH := txscript.IsPayToScriptHash(utxoEntry.ScriptPublicKey())
-		sigOpsCount += uint64(txscript.GetPreciseSigOpCount(sigScript, utxoEntry.ScriptPublicKey(), isP2SH))
+	// calculate mass for size
+	size := estimatedsize.TransactionEstimatedSerializedSize(tx)
+	massForSize := size * v.massPerTxByte
+
+	// calculate mass for scriptPubKey
+	totalScriptPubKeySize := uint64(0)
+	for _, output := range tx.Outputs {
+		totalScriptPubKeySize += 2 //output.ScriptPublicKey.Version (uint16)
```
someone235, line 33: I know this is not related to this PR, but do we really want to give extra weight to the version number?

domain/consensus/processes/transactionvalidator/transaction_in_isolation.go
```diff
@@ -9,7 +9,11 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (v *transactionValidator) ValidateTransactionInIsolation(tx *externalapi.DomainTransaction) error {
+// ValidateTransactionInIsolationAndPopulateMass validates the pares of the transaction that can be validated
```
someone235, line 12: pares->parts

domain/consensus/utils/consensushashing/calculate_signature_hash.go
```diff
@@ -102,6 +103,9 @@ func calculateSignatureHash(tx *externalapi.DomainTransaction, inputIndex int, t
 	sequencesHash := getSequencesHash(tx, hashType, reusedValues)
 	infallibleWriteElement(hashWriter, sequencesHash)
 
+	sigOpCountsHash := getSigOpCountsHash(tx, hashType, reusedValues)
+	infallibleWriteElement(hashWriter, sigOpCountsHash)
```
someone235, line 107: Shouldn't you always sign on your own sigopcount as well?

infrastructure/network/netadapter/server/grpcserver/protowire/rpc.proto
```diff
@@ -61,6 +61,7 @@ message RpcTransactionInput {
   RpcOutpoint previousOutpoint = 1;
   string signatureScript = 2;
   uint64 sequence = 3;
+  uint32 sigOpCount = 5;
```
someone235, line 64: You should update rpc.md

someone235 (changes requested):

domain/consensus/processes/transactionvalidator/mass.go
svarogg: As long as it is kept in the UTXOSet (And I believe it is), then absolutely.

The block mass reflects the social cost, and the fact something is mandatory or not doesn't change the social cost.

domain/consensus/utils/consensushashing/calculate_signature_hash.go
svarogg: Whooopppsss... good call.

domain/consensus/processes/transactionvalidator/mass.go
someone235: The question is why do we give extra weight script pub key size. If it's related to script execution, script version has nothing to do with it. And if it's related to UTXO set size, we should also give the same weight to the output value field.

domain/miningmanager/mempool/check_transaction_standard.go
```diff
@@ -129,7 +128,7 @@ func (mp *mempool) IsTransactionOutputDust(output *externalapi.DomainTransaction
 	// The most common scripts are pay-to-pubkey, and as per the above
 	// breakdown, the minimum size of a p2pk input script is 148 bytes. So
 	// that figure is used.
-	totalSerializedSize := estimatedsize.TransactionOutputEstimatedSerializedSize(output) + 148
+	totalSerializedSize := transactionvalidator.TransactionOutputEstimatedSerializedSize(output) + 148
```
someone235, line 131: Don't we prefer to just change it to use mass?

infrastructure/network/netadapter/server/grpcserver/protowire/rpc.proto
```diff
@@ -91,7 +91,7 @@ message RpcUtxoEntry {
 message RpcTransactionVerboseData{
   string transactionId = 1;
   string hash = 2;
-  uint64 size = 3;
+  uint64 mass = 3;
```
someone235, line 94: Give it another number to support backward compatability

someone235 (changes requested):

domain/miningmanager/mempool/check_transaction_standard.go
svarogg: So just to be clear (I think I agree, but want to make sure we're on the same page):
We remove maximumStandardSignatureScriptSize completely, because we SignatureScriptSize is embedded into mass, and the limit comes from there?

domain/consensus/processes/transactionvalidator/mass.go
svarogg: The answer is it's all about UTXOSet size.
Script execution limiting is through SigOpCount (the rest of the Ops are negligable).

So I've created an issue:
https://github.com/kaspanet/kaspad/issues/1796

someone235 (approved):

Issue #1721, opened 2021-05-13
Transactions are not validated for locktime when admitted into mempool


svarogg:
Transactions whose Locktime has not yet been reached should be rejected, but not cause a ban. The same way we treat immature coinbase spends.

hashdag:
If such txns currently enter the mempool (rather than cause a ban) -- defer to post mainnet

someone235:
I think the best approach is to have a third function in `TransactionValidator` that will be called `ValidateTransactionInContext` that will validate only `isFinalizedTransaction`. This will be called by both `ValidateBodyInContext` and `ValidateTransactionAndPopulateWithConsensusData`

PR #1794, opened 2021-07-13
Validate locktime when admitted into mempool and when building a block.

talelbaz:
#1721  
Validate locktime when admitted into mempool and when building a block.
Also, fix isFinalizedTransaction to use DAAscore instead of blue score.

### DIFF
--- a/domain/consensus/model/interface_processes_transactionvalidator.go
+++ b/domain/consensus/model/interface_processes_transactionvalidator.go
@@ -8,6 +8,8 @@ import (
 // it's possible to determine whether a transaction is valid
 type TransactionValidator interface {
 	ValidateTransactionInIsolation(transaction *externalapi.DomainTransaction) error
+	ValidateTransactionWithContextAndWithoutUTXOContext(stagingArea *StagingArea, tx *externalapi.DomainTransaction,
+		povBlockHash *externalapi.DomainHash) error
 	ValidateTransactionInContextAndPopulateMassAndFee(stagingArea *StagingArea,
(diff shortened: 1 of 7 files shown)


domain/consensus/model/interface_processes_transactionvalidator.go
```diff
@@ -8,6 +8,8 @@ import (
 // it's possible to determine whether a transaction is valid
 type TransactionValidator interface {
 	ValidateTransactionInIsolation(transaction *externalapi.DomainTransaction) error
+	ValidateTransactionWithContextAndWithoutUTXOContext(stagingArea *StagingArea, tx *externalapi.DomainTransaction,
```
stasatdaglabs, line 11: Consider renaming this to ValidateTransactionInContextIgnoringUTXO, so that we won't have a new concept called "utxo context"

domain/consensus/processes/blockbuilder/block_builder.go
```diff
@@ -145,6 +145,11 @@ func (bb *blockBuilder) validateTransaction(
 		return err
 	}
 
+	err = bb.transactionValidator.ValidateTransactionWithContextAndWithoutUTXOContext(stagingArea, transaction, model.VirtualBlockHash)
```
stasatdaglabs, line 148: ValidateTransactionInContextAndPopulateMassAndFee already calls this. Why call it twice?

stasatdaglabs (changes requested):

domain/consensus/processes/blockbuilder/block_builder.go
talelbaz: I can't find where, can you show me maybe?

stasatdaglabs (approved):

PR #1800, opened 2021-07-15
Disable relative time lock by time

### LINKED ISSUE #1753: Solve time-based relative timelocks after pruning headers
The current situation is that the UTXO-Set only remembers the output's DAAScore.

Then if the output has a time-based relative timelock, kaspad searches over the selectedParentChain to find the relevant block, and retrieve it's timestamp.
Once pruning headers is implemented, we no longer will be able to find such blocks the moment they are out of the pruning window, rendering the whole feature useless.

Possible alternatives:
1. Disable relative time-based timelocks. 
Some might argue that relative time-based timelocks are a redundant feature anyway, and DAA-score based timelocks can be used in all situations anyway. 
This might be correct, but requires further research to make sure there are no valid use-cases.
There's a general feeling of "we don't know why it might be used, but it definitively should be available"

2. Add AcceptingBlockTimestamp to UTXOEntry. 
This will solve the problem, and is not very complicated, but adds ~10% more data to the UTXO-Set - a constrained resource to start with.

talelbaz:
Closes #1753 
Considers each relative timelock to be interpreted as DAA score (ignore the type flag -> DAA score/time)
** There few changes in the test TestValidateTransactionInContextAndPopulateFee in transactionvalidator_test.go , but Its going to be all rewritten right after this pull request


domain/consensus/utils/txscript/opcode.go
```diff
@@ -1069,10 +1069,10 @@ func opcodeReturn(op *parsedOpcode, vm *Engine) error {
 }
 
 // verifyLockTime is a helper function used to validate locktimes.
-func verifyLockTime(txLockTime, threshold, lockTime uint64) error {
+func verifyLockTime(txLockTime, threshold, lockTime uint64, isRelativeLockTime bool) error {
 	// The lockTimes in both the script and transaction must be of the same
 	// type.
-	if !((txLockTime < threshold && lockTime < threshold) ||
+	if !isRelativeLockTime && !((txLockTime < threshold && lockTime < threshold) ||
```
someone235, line 1075: Why is it needed?

someone235 (changes requested):

domain/consensus/utils/txscript/opcode.go
talelbaz: because if its relative timelock you don't need to check the types - since there is only one type

domain/consensus/utils/txscript/opcode.go
```diff
@@ -1068,17 +1068,16 @@ func opcodeReturn(op *parsedOpcode, vm *Engine) error {
...
-		(txLockTime >= threshold && lockTime >= threshold)) {
-		str := fmt.Sprintf("mismatched locktime types -- tx locktime "+
-			"%d, stack locktime %d", txLockTime, lockTime)
-		return scriptError(ErrUnsatisfiedLockTime, str)
+func verifyLockTimeWithThreshold(txLockTime, threshold, lockTime uint64) error {
+	err := verifyLockTime(txLockTime, lockTime)
+	if err != nil {
+		return err
 	}
+	return verifyLockTimeTypes(txLockTime, threshold, lockTime)
```
someone235, line 1076: Please choose between `threshold` and `types`. I also prefer that `verifyLockTimeTypes` will happen before `verifyLockTime.`

someone235 (changes requested):

someone235 (approved):

PR #1820, opened 2021-08-16
Fix the RPCClient leaking connections

stasatdaglabs:


infrastructure/network/rpcclient/rpcclient.go
```diff
@@ -73,11 +61,23 @@ func (c *RPCClient) connect() error {
 	c.rpcRouter = rpcRouter
 
 	log.Infof("Connected to %s", c.rpcAddress)
+
+	getInfoResponse, err := c.GetInfo()
+	if err != nil {
+		return errors.Errorf("error making GetInfo request")
```
someone235, line 67: This should be wrapped

someone235 (changes requested):

someone235 (approved):

PR #1831, opened 2021-10-17
Switch to a fixed block subsidy after a certain work threshold

stasatdaglabs:
This depends on https://github.com/kaspanet/kaspad/pull/1830
Do not merge this PR before that one

The switch function is outlined here: https://hashdag.medium.com/kaspa-launch-plan-9a63f4d754a6


domain/consensus/processes/coinbasemanager/block_reward_switch.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/kaspanet/kaspad/domain/consensus/model"
+	"math/big"
+)
+
+func (c *coinbaseManager) isBlockRewardFixed(stagingArea *model.StagingArea) (bool, error) {
+	if c.hasBlockRewardSwitchedToFixed {
+		return true, nil
+	}
+
+	currentPruningPointIndex, err := c.pruningStore.CurrentPruningPointIndex(c.databaseContext, stagingArea)
```
someone235, line 13: You should check from the pruning point of the current checked block POV (the one in its header), otherwise you can encounter subjectivity issues.

someone235 (changes requested):
Please write a test for this behaviour and resolve https://github.com/kaspanet/kaspad/issues/1834

domain/consensus/processes/coinbasemanager/block_reward_switch.go
```diff
@@ -40,10 +37,28 @@ func (c *coinbaseManager) isBlockRewardFixed(stagingArea *model.StagingArea) (bo
...
+func (c *coinbaseManager) findPruningPointIndex(stagingArea *model.StagingArea, pruningPointHash *externalapi.DomainHash) (uint64, error) {
+	currentPruningPointHash, err := c.pruningStore.PruningPoint(c.databaseContext, stagingArea)
+	if err != nil {
+		return 0, err
+	}
+	currentPruningPointIndex, err := c.pruningStore.CurrentPruningPointIndex(c.databaseContext, stagingArea)
+	if err != nil {
+		return 0, err
+	}
+	for !currentPruningPointHash.Equal(pruningPointHash) && currentPruningPointIndex > 0 {
```
someone235, line 56: Why `currentPruningPointIndex > 0`? Don't we want to return an error if we get to zero and still don't find it?

domain/consensus/processes/blockbuilder/test_block_builder.go
```diff
@@ -200,7 +200,11 @@ func (bb *testBlockBuilder) buildBlockWithParents(stagingArea *model.StagingArea
 
 	bb.acceptanceDataStore.Stage(stagingArea, tempBlockHash, acceptanceData)
 
-	coinbase, err := bb.coinbaseManager.ExpectedCoinbaseTransaction(stagingArea, tempBlockHash, coinbaseData)
+	pruningPoint, err := bb.newBlockPruningPoint(stagingArea, tempBlockHash)
```
someone235, line 203: Can you reuse this pruning point later down the road?

someone235 (changes requested):

someone235 (approved):

PR #1844, opened 2021-11-07
Cache the miner state

elichai:
This allows a miner to cache the matrix generation to improve the performance of the miner.
I implemented a `MinerCache` that caches: the matrix, the target, the PrePowHash.

### DIFF
--- a/domain/consensus/processes/blockvalidator/pruning_violation_proof_of_work_and_difficulty.go
+++ b/domain/consensus/processes/blockvalidator/pruning_violation_proof_of_work_and_difficulty.go
@@ -148,7 +148,7 @@ func (v *blockValidator) validateDifficulty(stagingArea *model.StagingArea,
 //    difficulty is not performed.
 func (v *blockValidator) checkProofOfWork(header externalapi.BlockHeader) error {
 	// The target difficulty must be larger than zero.
-	state := pow.NewMinerState(header.ToMutable())
+	state := pow.NewState(header.ToMutable())
 	target := &state.Target
 	if target.Sign() <= 0 {
 		return errors.Wrapf(ruleerrors.ErrNegativeTarget, "block target difficulty of %064x is too low",
--- a/domain/consensus/processes/blockvalidator/pruning_violation_proof_of_work_and_difficulty.go
+++ b/domain/consensus/processes/blockvalidator/pruning_violation_proof_of_work_and_difficulty.go
@@ -8,7 +8,6 @@ import (
 	"github.com/kaspanet/kaspad/domain/consensus/utils/virtual"
 	"github.com/kaspanet/kaspad/infrastructure/db/database"
 	"github.com/kaspanet/kaspad/infrastructure/logger"
-	"github.com/kaspanet/kaspad/util/difficulty"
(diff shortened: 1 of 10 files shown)


domain/consensus/utils/pow/pow.go
```diff
@@ -12,52 +12,76 @@ import (
...
 
-// CheckProofOfWorkWithTarget check's if the block has a valid PoW according to the provided target
-// it does not check if the difficulty itself is valid or less than the maximum for the appropriate network
-func CheckProofOfWorkWithTarget(header externalapi.MutableBlockHeader, target *big.Int) bool {
-	// The block pow must be less than the claimed target
-	powNum := CalculateProofOfWorkValue(header)
-
-	// The block hash must be less or equal than the claimed target.
-	return powNum.Cmp(target) <= 0
+type MinerState struct {
```
someone235, line 15: Can you rename it to something more mining agnostic? Maybe just `State`? (and then it'll be `pow.State`).
It's used as part of validation, so the mining terminology feels unrelated.

domain/consensus/processes/blockvalidator/pruning_violation_proof_of_work_and_difficulty.go
```diff
@@ -163,7 +163,7 @@ func (v *blockValidator) checkProofOfWork(header externalapi.BlockHeader) error
 
 	// The block pow must be valid unless the flag to avoid proof of work checks is set.
 	if !v.skipPoW {
-		valid := pow.CheckProofOfWorkWithTarget(header.ToMutable(), target)
+		valid := state.CheckProofOfWork()
```
someone235, line 166: Maybe we should still keep some wrapper like `CheckProofOfWorkWithTarget` to make it more readable? The state here isn't reused.

domain/consensus/utils/pow/pow.go
```diff
@@ -78,7 +102,7 @@ func BlockLevel(header externalapi.BlockHeader) int {
 		return constants.MaxBlockLevel
 	}
 
-	proofOfWorkValue := CalculateProofOfWorkValue(header.ToMutable())
+	proofOfWorkValue := NewMinerState(header.ToMutable()).CalculateProofOfWorkValue()
```
someone235, line 105: nit: Use a wrapper

someone235 (changes requested):
Notice that you also need to `go fmt`

domain/consensus/processes/blockvalidator/pruning_violation_proof_of_work_and_difficulty.go
elichai: The state is used here, for the target.
I agree that a wrapper for no target is nice because it's simple, but a wrapper for no target, and a wrapper for with target, and maybe some other wrapper(for with matrix or with prePowHash) are exactly the reason I made this state :)

domain/consensus/utils/pow/pow.go
elichai: That's inside of the `pow` package, do you really feel it requires a wrapper?

someone235 (approved):

Issue #1860, opened 2021-11-28
When IBD finishes, the UTXO index in not updated accordingly


someone235:
We probably want ResolveVirtual to return a diff for each small resolve, and use that diff to update the UTXO index

cbitensky:
Working on it.

PR #1865, opened 2021-12-02
ResolveVirtual updates UTXOIndex

cbitensky:
Fixes https://github.com/kaspanet/kaspad/issues/1860


domain/utxoindex/utxoindex.go
```diff
@@ -93,6 +93,36 @@ func (ui *UTXOIndex) isSynced() (bool, error) {
 	return externalapi.HashesEqual(virtualInfo.ParentHashes, utxoIndexVirtualParents), nil
 }
 
+func (ui *UTXOIndex) UpdateUtxoDiff(utxoDiff externalapi.UTXODiff) (*UTXOChanges, error) {
+	ui.mutex.Lock()
+	defer ui.mutex.Unlock()
+
+	log.Tracef("Updating UTXO index with VirtualUTXODiff: %+v", utxoDiff)
```
someone235, line 100: You don't update the virtual parents in this method. IMHO you should make `ResolveVirtual` return the virtual parents and then just call `Update`

someone235 (changes requested):

someone235:
Fixed in https://github.com/kaspanet/kaspad/pull/1868

PR #1878, opened 2021-12-11
Changes to libkaspawallet to support Kaspaper

svarogg:


cmd/kaspawallet/keys/keys.go
```diff
@@ -89,6 +90,22 @@ func (d *File) toJSON() *keysFileJSON {
 	}
 }
 
+func NewFileFromMnemonics(params *dagconfig.Params, mnemonics string, password string) (*File, error) {
```
someone235, line 93: Please change `mnemonics` to `mnemonic`

someone235 (changes requested):

someone235 (approved):

PR #1892, opened 2021-12-19
Implement the new monetary policy

stasatdaglabs:


domain/consensus/processes/coinbasemanager/coinbasemanager.go
```diff
@@ -182,97 +179,47 @@ func acceptanceDataFromArrayToMap(acceptanceData externalapi.AcceptanceData) map
...
-	// If there's ghostdag data with trusted data we prefer it because we need the original merge set non-pruned merge set.
-	if database.IsNotFoundError(err) {
-		ghostdagData, err = c.ghostdagDataStore.Get(c.databaseContext, stagingArea, blockHash, false)
-		if err != nil {
-			return nil, err
-		}
-	}
+func (c *coinbaseManager) calcDeflationaryPeriodBlockSubsidy(blockDaaScore uint64) uint64 {
+	const secondsPerMonth = 2629800
+	monthsSinceDeflationaryPhaseStarted := (blockDaaScore - c.deflationaryPhaseDaaScore) / secondsPerMonth
```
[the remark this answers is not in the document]
michaelsutton: I think we should avoid premature planning of BPS changes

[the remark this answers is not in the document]
michaelsutton: @someone235 what do you think?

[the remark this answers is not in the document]
someone235: Anyway @stasatdaglabs, you need to be consistent. Because in the above line you didn't take the block rate into account, and a few lines under you did.

domain/dagconfig/consensus_defaults.go
```diff
@@ -76,4 +74,6 @@ const (
 	defaultTargetTimePerBlock = 1 * time.Second
 
 	defaultPruningProofM = 1000
+
+	defaultDeflationaryPhaseDaaScore = 15778800 // Half a year
```
someone235, line 78: From my calculations this should be triggered at 10/6/2022, which is not 6 months from mainnet start (7/11/2021)

someone235 (changes requested):

domain/consensus/processes/coinbasemanager/coinbasemanager.go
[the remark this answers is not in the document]
stasatdaglabs: Got rid of it for now

domain/consensus/processes/coinbasemanager/coinbasemanager.go
```diff
@@ -182,97 +179,47 @@ func acceptanceDataFromArrayToMap(acceptanceData externalapi.AcceptanceData) map
...
-
-	// If there's ghostdag data with trusted data we prefer it because we need the original merge set non-pruned merge set.
-	if database.IsNotFoundError(err) {
-		ghostdagData, err = c.ghostdagDataStore.Get(c.databaseContext, stagingArea, blockHash, false)
-		if err != nil {
-			return nil, err
-		}
-	}
+func (c *coinbaseManager) calcDeflationaryPeriodBlockSubsidy(blockDaaScore uint64) uint64 {
+	const secondsPerMonth = 2629800
```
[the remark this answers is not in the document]
stasatdaglabs: Added a comment

domain/dagconfig/consensus_defaults.go
stasatdaglabs: Updated the calculation and added a comment

someone235 (dismissed):

domain/consensus/processes/coinbasemanager/coinbasemanager.go
```diff
@@ -182,97 +177,44 @@ func acceptanceDataFromArrayToMap(acceptanceData externalapi.AcceptanceData) map
...
+	monthsSinceDeflationaryPhaseStarted := (blockDaaScore - c.deflationaryPhaseDaaScore) / secondsPerMonth
 
-	// If there's ghostdag data with trusted data we prefer it because we need the original merge set non-pruned merge set.
-	if database.IsNotFoundError(err) {
-		ghostdagData, err = c.ghostdagDataStore.Get(c.databaseContext, stagingArea, blockHash, false)
-		if err != nil {
-			return nil, err
-		}
-	}
+	// enlargedDeflationConstant is the estimated deflation constant `2^1/12` multiplied by a big
```
michaelsutton, line 202: Can we add the following full explanation?:

/*
Denote `B`, `M` to be `deflationaryPhaseBaseSubsidy` and `monthsSinceDeflationaryPhaseStarted` respectively.

We want to compute the following function in a deterministic way:

`subsidy = B/(2^(1/12))^M`

Set `C = enlargedDeflationConstant` and `F = enlargeFactor` large enough s.t. C/F ~= 2^(1/12).

We start with `c=1 (enlargedDeflationMultiplier)`, `f=1 (enlargeFactorDivisor)` and repeat c *= C, f *= F for M times
 s.t. eventually  `c/f ~= (2^(1/12))^M`.  

Finally we compute `subsidy = B * (f/c)`

*/ 

P.S. 1: I would love it if var names were actually short like in this snippet^, it's much harder to read equations with sentences then with simple in-context clear names such as `c`, `f`

P.S. 2: with current base subsidy, the subsidy will reach zero when `M=425`, so effectively we can pre-compute this and place it in a table. Just saying. (for instance benchmarking this function shows that with `M=30*12` it takes ~2ms)

domain/consensus/processes/coinbasemanager/coinbasemanager_test.go
```diff
@@ -78,3 +79,50 @@ func TestCalcDeflationaryPeriodBlockSubsidy(t *testing.T) {
 		}
 	}
 }
+
+func TestBuildSubsidyTable(t *testing.T) {
```
someone235, line 83: Add t.Skip() so this test won't run every time the test suite is run

someone235 (changes requested):

domain/consensus/processes/coinbasemanager/coinbasemanager_test.go
michaelsutton: But I want it to run in case base subsidy is modified and this method is not modified too

michaelsutton: Ok, fixed it so now table printing is using the test logger and not console print

someone235 (approved):

domain/consensus/processes/coinbasemanager/coinbasemanager.go
```diff
@@ -182,97 +177,68 @@ func acceptanceDataFromArrayToMap(acceptanceData externalapi.AcceptanceData) map
...
-	for i := 0; i < binomialSteps; i++ {
-		step := firstTwoBytes & 1
-		firstTwoBytes >>= 1
-		binomialSum += int64(step)
-	}
-	return binomialSum - (binomialSteps / 2), nil
+func (c *coinbaseManager) calcDeflationaryPeriodBlockSubsidyFloatCalc(month uint64) uint64 {
+	baseSubsidy := c.deflationaryPhaseBaseSubsidy
+	subsidy := float64(baseSubsidy) / math.Pow(2, float64(month)/12)
+	return uint64(subsidy)
```
elichai, line 241: Can we please not deploy this code as long as it has floats in it?
would appreciate if someone could explain to me the new monetary policy (or link an explanation) so we can come up with a calculation that doesn't involve floats

michaelsutton: Please read a few lines before as well...

michaelsutton: In short, we used this method to create the table above, and only the table is actually used

domain/consensus/processes/coinbasemanager/coinbasemanager.go
```diff
@@ -182,97 +177,68 @@ func acceptanceDataFromArrayToMap(acceptanceData externalapi.AcceptanceData) map
...
-		coinbaseTransaction := mergeSetBlock.Transactions[transactionhelper.CoinbaseTransactionIndex]
-		_, _, mergeSetBlockSubsidy, err := c.ExtractCoinbaseDataBlueScoreAndSubsidy(coinbaseTransaction)
-		if err != nil {
-			return nil, err
-		}
-		mergeSetSubsidySum += int64(mergeSetBlockSubsidy)
-	}
-	return big.NewRat(mergeSetSubsidySum, 1), nil
+/*
+	This table was pre-calculated by calling `calcDeflationaryPeriodBlockSubsidyFloatCalc` for all months until reaching 0 subsidy.
```
michaelsutton, line 207: @elichai see here

PR #1899, opened 2021-12-21
Address Manager: Don't seed if we have enough Addresses + Remove Address that failed connection too many times

### LINKED ISSUE #1588: If kaspad runs out of addresses - no new ones will be ever added

svarogg:
This PR does two things:
1. Fixes #1588 which was improperly fixed in a way that seedFromDNS is called on every ConnectionManager loop.
After this PR seedFromDNS will be called only if we don't have enough Addresses in AddressManager
2. Removes an Address that has failed more then 10 connection attempts.


someone235:
Ideally I wouldn't want to remove any address when the address manager isn't full, and instead of that just give lower priority to addresses with many failures. We should randomize addresses with preference to addresses with lower amount of failures.

stasatdaglabs (approved):
AddressManager.addAddressNoLock contains some logic that relies on connectionFailedCount. Please check whether that's still relevant. Approved otherwise

svarogg:
> 

Yes, it is still relevant - in case we have too much addresses, it's still desirable to remove someone who has 2 connectFailedCount, even though it's smaller then the amount for auto-removal.

svarogg:
> AddressManager.addAddressNoLock contains some logic that relies on connectionFailedCount. Please check whether that's still relevant. Approved otherwise

@stasatdaglabs 

Yes, it is still relevant - in case we have too much addresses, it's still desirable to remove someone who has 2 connectFailedCount, even though it's smaller then the amount for auto-removal.

michaelsutton (approved):

someone235 (approved):

PR #1904, opened 2021-12-22
Wallet: show balance by addresses

cbytensky:
Also added two digits of precision. Now it shows exact number of Sompis.
Result of `kaspawallet balance -v` now looks like:
```
Address                                                                       Available             Pending
-----------------------------------------------------------------------------------------------------------
kaspa:qqcl3q8s8qdwxtm42fkjw6atu9wwafamv3tza26r4wvayh3jerkdv59z8l50d    3248804.12960945
kaspa:qzc34p2g0m04prt9akh3d7y0hqgxq07rl3yhskhjg6a9tx4vq38kq7mzwcylu        353.11366474
-----------------------------------------------------------------------------------------------------------
                                                              Total    3249157.24327419
```
`kaspawallet balance`:
```
                        Available             Pending
Total balance    3249157.24327419
```


cmd/kaspawallet/balance.go
```diff
@@ -22,10 +30,13 @@ func balance(conf *balanceConfig) error {
 		return err
 	}
 
-	fmt.Printf("Balance:\t\tKAS %f\n", float64(response.Available)/constants.SompiPerKaspa)
-	if response.Pending > 0 {
-		fmt.Printf("Pending balance:\tKAS %f\n", float64(response.Pending)/constants.SompiPerKaspa)
+	println("Address                                                                       Available             Pending")
```
someone235, line 33: I would make the detailed format optional, and the default view should be only balance and pending balance IMHO

cmd/kaspawallet/daemon/server/balance.go
```diff
@@ -3,8 +3,12 @@ package server
 import (
 	"context"
 	"github.com/kaspanet/kaspad/cmd/kaspawallet/daemon/pb"
+	"github.com/kaspanet/kaspad/cmd/kaspawallet/libkaspawallet"
 )
 
+type balancesType struct{ Available, Pending uint64 }
```
someone235, line 9: Available and Pending should be private

someone235 (changes requested):

cmd/kaspawallet/balance.go
```diff
@@ -22,10 +30,21 @@ func balance(conf *balanceConfig) error {
 		return err
 	}
 
-	fmt.Printf("Balance:\t\tKAS %f\n", float64(response.Available)/constants.SompiPerKaspa)
-	if response.Pending > 0 {
-		fmt.Printf("Pending balance:\tKAS %f\n", float64(response.Pending)/constants.SompiPerKaspa)
+	if conf.Verbose {
+		print("Address                                               ")
+	}
+	println("                        Available             Pending")
```
someone235, line 36: I think displaying it as a table doesn't make sense when you have only one line, so I think it's better that the display will remain as before in non verbose mode.

someone235 (changes requested):

cmd/kaspawallet/balance.go
cbytensky: There are two lines. Second about pending balance.

cbytensky: There _was_ a table before. With two lines and one column. Aligned by tab character.

The code will be more complex. The output will be ununified with verbose mode.

I am sure that now is better.

cbytensky: Update:
Made it in one line

cbytensky:
```
Total balance, KAS    3249157.24327419
```

someone235: It's very weird that the KAS comes before the actual number. What was bad with the way it was before?

someone235: And generally speaking UX decisions shouldn't be affected by code complexity considerations when the difference in code complexity terms is so minor.

cbytensky: It financial information systems is _very_ usual, when unit of measure is taken out column data. And there _is_ column data. Because when there is pending balance, output looks like
```
Total balance, KAS    3249157.24327419        500.00000000 (pending)
```
And with `--verbose`:
```
Address                                                                       Available             Pending
-----------------------------------------------------------------------------------------------------------
kaspa:qqcl3q8s8qdwxtm42fkjw6atu9wwafamv3tza26r4wvayh3jerkdv59z8l50d    3248804.12960945        500.00000000
kaspa:qzc34p2g0m04prt9akh3d7y0hqgxq07rl3yhskhjg6a9tx4vq38kq7mzwcylu        353.11366474
-----------------------------------------------------------------------------------------------------------
                                                 Total balance, KAS    3249157.24327419        500.00000000
```
It will be wierd when `KAS` will repeat after every number.

Now in `dev` branch `balance` output is:
```
Balance:		KAS 3249157.243274
```
There _is_ `KAS` before number. Why?

cbytensky: I also not care about code complexity if UX requires it.
When I told about code complexity, I have in mind namely UX complexity. I made interface as unified as I can. Unified between output with and without `--verbose`. I tried to make as more common elements as I can, because more common elements means more simple UX (in general of cause).

someone235 (approved):

PR #1908, opened 2021-12-23
fix CONTRIBUTING.md text

Aleoami:


cmd/kaspawallet/keys/keys.go
```diff
@@ -367,7 +367,7 @@ func (d *File) detectNumThreads(password []byte, encryptedMnemonic *EncryptedMne
 
 		_, err := decryptMnemonic(numThreadsGuess, encryptedMnemonic, password)
 		if err != nil {
-			const maxTries = 32
+			const maxTries = 256
```
someone235, line 370: The max number of threads that argon2.IDKey can get is 255

someone235 (changes requested):

cmd/kaspawallet/keys/keys.go
Aleoami:
```suggestion
			const maxTries = 255
```

someone235 (approved):

Issue #1896, opened 2021-12-19
Ignore red blocks from submitBlock RPC command


someone235:
There's no point in mining blocks that are red from the miner POV, so the miner node should ignore them. We can implement this idea by checking on submitBlock if the submitted block blue score diff from the selected tip blue score is more than `K+1`.
The user should also be able to transfer `allowRedBlocks=true` when using the submit block in case he really wants to add that block.

Edit:
Instead of red blocks check for blocks outside of the DAA window. We can check if the blue score diff is greater than 2641, and change the flag name to `allowNonDAABlocks`.

michaelsutton:
I think that even if `allowRedBlocks=true` we should reject blocks which are out of the DAA window. Of course one can still modify the code, but such blocks are essentially spam (usually a result of miner/kaspad software bugs) and there is no reason to prorogate them.

someone235:
> I think that even if `allowRedBlocks=true` we should reject blocks which are out of the DAA window. Of course one can still modify the code, but such blocks are essentially spam (usually a result of miner/kaspad software bugs) and there is no reason to prorogate them.

The idea behind the flag is that maybe you want to sync your node manually, by sending it block by block via submitBlock, but your node might have the wrong selected chain so all of the submitted blocks will be rejected

michaelsutton:
> > I think that even if `allowRedBlocks=true` we should reject blocks which are out of the DAA window. Of course one can still modify the code, but such blocks are essentially spam (usually a result of miner/kaspad software bugs) and there is no reason to prorogate them.
> 
> The idea behind the flag is that maybe you want to sync your node manually, by sending it block by block via submitBlock, but your node might have the wrong selected chain so all of the submitted blocks will be rejected

I still see a difference between red blocks and non-DAA blocks. One can think that a large miner would like to use red blocks since he might be the one merging them as well.

PR #1914, opened 2021-12-28
Reject out of date blocks submitted via RPC 

michaelsutton:
In order to avoid lagging miners from spamming the network, we reject by default all submitted blocks with `DAA score < virtual DAA score - DAA window size`. Such blocks are not rewarded anyhow. See issue #1896


app/rpc/rpchandlers/submit_block.go
```diff
@@ -34,6 +34,22 @@ func HandleSubmitBlock(context *rpccontext.Context, _ *router.Router, request ap
...
 	}
 
+	if !context.Config.AllowSubmittingNonDAABlocks {
+		virtualDAAScore, err := context.Domain.Consensus().GetVirtualDAAScore()
+		if err != nil {
+			return nil, err
+		}
+		// A simple heuristic check which signals that the mined block is out of date
+		// and should not be accepted unless user explicitly requests
+		if int(virtualDAAScore-domainBlock.Header.DAAScore()) > context.Config.NetParams().DifficultyAdjustmentWindowSize {
```
michaelsutton, line 44: @someone235 I'm not sure if I should test the case `virtualDAAScore < domainBlock.Header.DAAScore()` and what does it even mean to be in such state. Currently this will overflow and the condition will be true, hence block rejected

infrastructure/config/config.go
```diff
@@ -122,6 +122,7 @@ type Flags struct {
 	IsArchivalNode                  bool          `long:"archival" description:"Run as an archival node: don't delete old block data when moving the pruning point (Warning: heavy disk usage)'"`
 	AllowSubmitBlockWhenNotSynced   bool          `long:"allow-submit-block-when-not-synced" hidden:"true" description:"Allow the node to accept blocks from RPC while not synced (this flag is mainly used for testing)"`
 	EnableSanityCheckPruningUTXOSet bool          `long:"enable-sanity-check-pruning-utxo" hidden:"true" description:"When moving the pruning point - check that the utxo set matches the utxo commitment"`
+	AllowSubmittingNonDAABlocks     bool          `long:"allow-non-daa-blocks" hidden:"false" description:"When submitting blocks via RPC - accept blocks which are out of virtual daa window (although such blocks are not rewarded anyhow)"`
```
michaelsutton, line 125: @someone235  is `hidden:"false"` needed here ?

someone235: I think it's better to move this flag directly to the submitBlock RPC command

someone235 (changes requested):

svarogg (approved):

someone235 (approved):

PR #1885, opened 2021-12-14
Add request balance by address to kaspactl

FestinaLente666:

PR #1915, opened 2021-12-29
Fix protobuf generation for PR #1885

michaelsutton:


michaelsutton:
Fixes protobuf generation for PR #1885. Some changes were made to field names, but the command `go generate .` was not re runed. Note this too @FestinaLente666

someone235 (approved):

PR #1918, opened 2021-12-30
Remove DNS seeder managed by Elichai Turkel (its offline)

michaelsutton:
This DNS seeder by Elichai Turkel @elichai is offline


elichai (approved):
ACK

PR #1921, opened 2022-01-02
P2P upgrade mechanism

someone235:


stasatdaglabs (changes requested):
* I suspect that connections between v3 and v4 will fail, since v3 will attempt to send an address exchange message and v4 will disconnect. In either case please write an integration test to check this.
* If a flow is identical between v3 and v4 please reuse the code. As a policy, I think we should only copy+paste once there is a difference, but not before.

app/protocol/flows/handshake/receiveversion.go
```diff
@@ -18,6 +19,8 @@ var (
 	// minAcceptableProtocolVersion is the lowest protocol version that a
 	// connected peer may support.
 	minAcceptableProtocolVersion = uint32(3)
+
+	defaultMaxProtocolVersion = uint32(4)
```
stasatdaglabs, line 23: The word "default" here makes it sound like the max protocol may change during runtime. Why not maxAcceptableProtocolVersion or defaultProtocolVersion?

stasatdaglabs (changes requested):

stasatdaglabs (approved):

Issue #1922, opened 2022-01-03
Add compound feature to kaspawallet


svarogg:
With current limits it's only possible to send up to ~50k KAS of mined outputs before the transaction becomes too large to be accepted into the mempool (== fit in a block).

I suggest to modify kaspawallet so that it detects large transactions, and offers to compound multiple transactions into one.

More specifically: assume we need to spend **n+m** outputs where **n** is the maximum outputs a transaction can spend.
Kaspawallet will create transaction **tx1** which will spend the first **n** outputs into some exchange address and **tx2** that will spend the remaining m outputs into same change address.
Then kaspawallet will create a third **compounding_tx** which will send the outputs of **tx1** and **tx2** to the specified recipient.

svarogg:
Alternatively what we can do:
1. If kaspawallet creates a too-big-transaction - print an error message and exit.
2. Add to kaspawallet a `compound` command, which will gather all funds into as little utxos as possible.

someone235:
tx2 is redundant in the case that `m<n`. You can directly create the compound tx by using tx1.output and the rest m outputs.
Or in more general, if you need to use `k` outputs, use `ceil(k/n)` transactions.

someone235:
And part of the issue should be too change the sign and broadcast commands to accept an array of transactions instead of a single one, and to change the createRawTransaction command to output an array.

someone235:
BTW, I'm not sure if this is what you meant, but tx1.address should just be a regular change address.

PR #1924, opened 2022-01-04
Make kaspawallet store the utxos sorted by amount

svarogg:
This is a preliminary PR for #1922 .

I'm sending this as a separate PR, since it will immidiately help, even without the compound feature, which might take some time to implement properly.

This changes kaspawallet daemon to always keeps a sorted list of UTXOs.
Thus, on average, less UTXOs would be needed for any created transaction, making it less likely that a too-big transaction would be created.

Note this does incur some performance hit, especially for large wallets, but profiling showed it's probably negligible.


michaelsutton:
Note that this approach also has some negative effect, since also a medium-size send will use the largest UTXO, while it could of merged a few small UTXOs with out passing the mempool/block limit

svarogg:
Anyway, the main thing this PR does is make sure the utxos are sorted, from which end do we pick them can be discussed and decided upon later down the road.

cmd/kaspawallet/daemon/server/sync.go
```diff
@@ -219,18 +219,30 @@ func (s *server) addEntryToUTXOSet(entry *appmessage.UTXOsByAddressesEntry, addr
...
+	})
 
 	return nil
 }
 
+// insertUTXO inserts the given utxo into s.utxosSortedByAmount, while keeping it sorted.
+func (s *server) insertUTXO(utxo *walletUTXO) {
+	s.utxosSortedByAmount = append(s.utxosSortedByAmount, utxo)
+	// bubble up the new UTXO to keep the UTXOs sorted by value
+	index := len(s.utxosSortedByAmount) - 1
```
michaelsutton, line 235: This can be a binary search

michaelsutton: e.g. same logic as here https://github.com/kaspanet/kaspad/blob/master/domain/consensus/processes/reachabilitymanager/ordered_tree_node_set.go#L42

cmd/kaspawallet/daemon/server/sync.go
```diff
@@ -244,13 +256,14 @@ func (s *server) refreshExistingUTXOs() error {
 		return err
 	}
 
-	s.utxos = make(map[externalapi.DomainOutpoint]*walletUTXO, len(getUTXOsByAddressesResponse.Entries))
+	s.utxosSortedByAmount = make([]*walletUTXO, 0, len(getUTXOsByAddressesResponse.Entries))
```
michaelsutton, line 259: Binary search above is especially important here since it will reduce the overall insertion cost from n^2 to nlogn

cmd/kaspawallet/daemon/server/sync.go
someone235: You can also use `sort.Search`, like in `findTransactionIndex`

someone235: nit: I think it's more readable to use code such as `func (tobf *TransactionsOrderedByFeeRate) Push`, and maybe even more efficient

someone235 (changes requested):

app/appmessage/rpc_get_balance_by_address.go
```diff
@@ -9,7 +9,7 @@ type GetBalanceByAddressRequestMessage struct {
 
 // Command returns the protocol command string for the message
 func (msg *GetBalanceByAddressRequestMessage) Command() MessageCommand {
-	return CmdGetBalanceByAddressRequestMessage
+	return CmdGetBalancesByAddressesRequestMessage
```
michaelsutton, line 12: was this file suppose to change?

app/appmessage/rpc_get_balance_by_address.go
```diff
@@ -30,7 +30,7 @@ type GetBalanceByAddressResponseMessage struct {
 
 // Command returns the protocol command string for the message
 func (msg *GetBalanceByAddressResponseMessage) Command() MessageCommand {
-	return CmdGetBalanceByAddressResponseMessage
+	return CmdGetBalancesByAddressesResponseMessage
```
michaelsutton, line 33: same here, seems like a mistake and it should remain `CmdGetBalanceByAddressResponseMessage`

cmd/kaspawallet/daemon/server/sync.go
```diff
@@ -200,71 +189,49 @@ func (s *server) refreshExistingUTXOsWithLock() error {
...
+		}
+		utxos[i] = &walletUTXO{
+			Outpoint:  outpoint,
+			UTXOEntry: utxoEntry,
+			address:   address,
+		}
 	}
 
-	getUTXOsByAddressesResponse, err := s.rpcClient.GetUTXOsByAddresses(addressSet.strings())
+	sort.Slice(utxos, func(i, j int) bool { return utxos[i].UTXOEntry.Amount() > utxos[j].UTXOEntry.Amount() })
```
someone235, line 221: Maybe it's better to test it on testnet first with a very big wallet?

someone235 (approved):

cmd/kaspawallet/daemon/server/sync.go
michaelsutton: `n log n` should be fine any way

PR #1925, opened 2022-01-09
Filter redundant blocks from daa window

someone235:


domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -1133,3 +1127,37 @@ func (pm *pruningManager) isPruningPointInPruningDepth(stagingArea *model.Stagin
 
 	return blockGHOSTDAGData.BlueScore() >= pruningPointHeader.BlueScore()+pm.pruningDepth, nil
 }
+
+func (pm *pruningManager) TrustedBlockAssociatedGHOSTDAGDataBlockHashes(stagingArea *model.StagingArea, blockHash *externalapi.DomainHash) ([]*externalapi.DomainHash, error) {
+	blockHashes := make([]*externalapi.DomainHash, 0, pm.k)
+	current := blockHash
+	isTrustedData := false
+	for i := externalapi.KType(0); i < pm.k+1; i++ {
```
stasatdaglabs, line 1135: This might be splitting hairs, but pm.k+1 might overflow. Please write i <= pm.k instead

domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -1133,3 +1127,37 @@ func (pm *pruningManager) isPruningPointInPruningDepth(stagingArea *model.Stagin
...
+
+		blockHashes = append(blockHashes, current)
+
+		if ghostdagData.SelectedParent().Equal(pm.genesisHash) {
+			break
+		}
+
+		if current.Equal(pm.genesisHash) {
+			break
+		}
```
stasatdaglabs, line 1157: Why the above two conditions? Wouldn't the second one suffice?

domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -1133,3 +1127,37 @@ func (pm *pruningManager) isPruningPointInPruningDepth(stagingArea *model.Stagin
...
+
+		current = ghostdagData.SelectedParent()
```
stasatdaglabs, line 1159: Please make sure there aren't any scenarios where this can fail after the first prune.

app/protocol/flows/v4/blockrelay/handle_pruning_point_and_its_anticone_requests.go
```diff
@@ -57,8 +58,73 @@ func HandlePruningPointAndItsAnticoneRequests(context PruningPointAndItsAnticone
 				return err
 			}
 
+			windowSize := context.Config().NetParams().DifficultyAdjustmentWindowSize
+			daaWindowBlocks := make([]*externalapi.TrustedDataDataDAAHeader, 0, windowSize)
+			daaWindowHashesToIndex := make(map[externalapi.DomainHash]int, windowSize)
+			trustedDataDAABlockIndexes := make(map[externalapi.DomainHash][]uint64)
+
+			ghostdagData := make([]*externalapi.BlockGHOSTDAGDataHashPair, 0)
```
stasatdaglabs, line 66: Why did you leave out the capacity here?

stasatdaglabs (changes requested):
Please add the following tests:
* v3 syncs from v4
* v4 syncs from v3

app/appmessage/p2p_msgtrusteddata.go
```diff
@@ -0,0 +1,19 @@
+package appmessage
+
+// MsgTrustedData represents a kaspa TrustedData message
+type MsgTrustedData struct {
```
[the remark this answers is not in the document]
someone235: One of the slow stability tests checks it

domain/consensus/processes/pruningmanager/pruningmanager.go
someone235: current can start from genesis, and then the above condition won't break the loop

domain/consensus/processes/pruningmanager/pruningmanager.go
someone235: We don't prune headers yet, but when we do, we shouldn't delete the last K blocks from the selected chain of each pruning point anticone block.

app/protocol/flows/v4/blockrelay/handle_pruning_point_and_its_anticone_requests.go
someone235: Because it's hard to estimate the capacity

testing/integration/ibd_test.go
```diff
@@ -199,6 +217,12 @@ func TestIBDWithPruning(t *testing.T) {
 
 	// Test a situation where a node with pruned headers syncs another fresh node.
 	testSync(syncee1, syncee2)
+
+	// Test syncing from p2p v4 to p2p v3
+	testSync(syncee2, synceeV3)
+
+	// Test syncing from p2p v3 to p2p v4
+	testSync(synceeV3, synceeV4)
```
stasatdaglabs, line 225: synceeV4 appears to have protocolVersion: 3

stasatdaglabs (changes requested):

stasatdaglabs (dismissed):

stasatdaglabs (approved):

PR #1927, opened 2022-01-16
Add --transaction-file options to the `sign` and `broadcast` wallet subcommands

stasatdaglabs:


someone235 (approved):

demisrael:
what about save hex when creating unsigned transaction? it's 120Mb of console output!

PR #1930, opened 2022-01-23
Extract IBD management from invs relay flow to a new separated flow 

michaelsutton:
Currently if a node has performance issues during IBD and takes more than 100 seconds to process a bulk of headers, the connection to the IBD peer is aborted due to route capacity limit reached (since invs are not dequeued). This in turn causes the IBD to abort and delete the DB prefix, starting the whole process all over, sometimes in an infinite loop.

This PR fixes this by completely separating the IBD flow from invs relay flow and using a channel to communicate between them.

For now I only applied the changes to v3. Once we finalize this I'll copy the changes to v4.

### DIFF
--- a/app/protocol/flows/v3/register.go
+++ b/app/protocol/flows/v3/register.go
@@ -62,8 +62,10 @@ func registerBlockRelayFlows(m protocolManager, router *routerpkg.Router, isStop
 			appmessage.CmdInvRelayBlock, appmessage.CmdBlock, appmessage.CmdBlockLocator,
 		},
 			isStopping, errChan, func(incomingRoute *routerpkg.Route, peer *peerpkg.Peer) error {
-				return blockrelay.HandleRelayInvs(m.Context(), incomingRoute,
+				err := blockrelay.HandleRelayInvs(m.Context(), incomingRoute,
 					outgoingRoute, peer, ibdChannel)
+				close(ibdChannel)
+				return err
 			},
 		),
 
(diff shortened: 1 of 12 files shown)


app/protocol/flows/v3/register.go
```diff
@@ -49,6 +50,7 @@ func registerAddressFlows(m protocolManager, router *routerpkg.Router, isStoppin
 
 func registerBlockRelayFlows(m protocolManager, router *routerpkg.Router, isStopping *uint32, errChan chan error) []*common.Flow {
 	outgoingRoute := router.OutgoingRoute()
+	ibdChannel := make(chan *externalapi.DomainBlock)
```
stasatdaglabs, line 53: Instead of creating the channel here and passing it to both flows, please add it to FlowContext.

stasatdaglabs (changes requested):

app/protocol/flows/v3/register.go
michaelsutton: But FlowContext is shared between all peers, The IBD channel is per peer

stasatdaglabs: Good point. Please add it to Peer instead

michaelsutton: Are you sure? it seems a bit out of context for the Peer struct

stasatdaglabs: It's about as in-context as the last ping time. Certainly not more out of context than the flow registration function

someone235 (approved):

stasatdaglabs (approved):

Issue #1920, opened 2022-01-02
Kaspad crashes on out-of-memory during IBD from scratch


stasatdaglabs:
This happens fairly consistently on my 8gb VM
Please profile memory during the entire IBD process and see whether it may be optimized to not take up so much RAM

PR #1937, opened 2022-01-30
Split ApplyPruningPointProof to multiple small database transactions

stasatdaglabs:
This fixes https://github.com/kaspanet/kaspad/issues/1920

ApplyPruningPointProof is always called on the Staging Consensus during IBD, so there's no risk of data inconsistency.
This PR splits the writes from this function to the database to multiple small transactions, so that the garbage collector has a chance to clean up while data is being written
Note that it's still a big, quick write, so RAM consumption isn't reduced to zero. On my VM I've watched it consume up to 1gb additional memory throughout the process

### DIFF
--- a/domain/consensus/processes/pruningproofmanager/pruningproofmanager.go
+++ b/domain/consensus/processes/pruningproofmanager/pruningproofmanager.go
@@ -608,6 +608,10 @@ func (ppm *pruningProofManager) dagProcesses(
 	return reachabilityManagers, dagTopologyManagers, ghostdagManagers
 }
 
+// ApplyPruningPointProof applies the given pruning proof to the current consensus. Specifically,
+// it's meant to be used against the StagingConsensus during headers-proof IBD. Note that for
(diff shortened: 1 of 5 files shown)


domain/consensus/processes/pruningproofmanager/pruningproofmanager.go
```diff
@@ -607,13 +608,15 @@ func (ppm *pruningProofManager) dagProcesses(
 	return reachabilityManagers, dagTopologyManagers, ghostdagManagers
 }
 
-func (ppm *pruningProofManager) ApplyPruningPointProof(stagingArea *model.StagingArea, pruningPointProof *externalapi.PruningPointProof) error {
+func (ppm *pruningProofManager) ApplyPruningPointProof(pruningPointProof *externalapi.PruningPointProof) error {
```
someone235, line 611: Can you add a comment that this function is not atomic and if it fails the outside scope should probably delete the database?

someone235 (changes requested):

someone235 (approved):

Issue #1912, opened 2021-12-26
Display progress of IBD process in Kaspad logs


hashdag:
Currently the operator has no easy way to discern whether the IBD process is close to or far from completion.

Add a log that prints e.g. % of IBD process that has been completed so far.

For instance, use DAA score of the IBD-triggering orphan vs the DAA score of the highest properly synced block (essentially -- the syncee's selected).

Should be done after IBD indeed operates with a linear progress process #1895

stasatdaglabs:
Implemented in https://github.com/kaspanet/kaspad/pull/1938

PR #1938, opened 2022-01-30
Display progress of IBD process in Kaspad logs

stasatdaglabs:
This implements https://github.com/kaspanet/kaspad/issues/1912

Reports for both protocol version 3 and 4 are added, for both header and block downloads

The reports look like this:
`2022-01-30 16:37:14.450 [INF] PROT: IBD: Processed 63600 blocks (28%)`


someone235 (changes requested):
I think we should also report the progress of block body progress.

someone235 (approved):

PR #1949, opened 2022-02-18
Add progress indication for virtual resolution

someone235:


app/protocol/flows/v4/blockrelay/ibd.go
```diff
@@ -590,6 +590,10 @@ func (flow *handleIBDFlow) resolveVirtual(estimatedVirtualDAAScoreTarget uint64)
 			if err != nil {
 				return err
 			}
+			percents := int(float64(virtualDAAScore-virtualDAAScoreStart) / float64(estimatedVirtualDAAScoreTarget-virtualDAAScoreStart) * 100)
+			if percents < 0 {
```
michaelsutton, line 594: How can this become negative? is it possible that `virtualDAAScore < virtualDAAScoreStart` ?
If it's possible that `estimatedVirtualDAAScoreTarget < virtualDAAScoreStart` then I would set `estimatedVirtualDAAScoreTarget = virtualDAAScoreStart` before the `for` loop

app/protocol/flows/v4/blockrelay/ibd.go
```diff
@@ -590,6 +590,10 @@ func (flow *handleIBDFlow) resolveVirtual(estimatedVirtualDAAScoreTarget uint64)
...
+				percents = 100
+			}
 			log.Infof("Resolving virtual. Estimated progress: %d%%",
```
michaelsutton, line 597: You were probably planning on using `percents` here

michaelsutton (changes requested):

app/protocol/flows/v4/blockrelay/ibd.go
someone235: Because in some edge cases `virtualDAAScoreStart` can be higher than `estimatedVirtualDAAScoreTarget`

michaelsutton (approved):

PR #1951, opened 2022-02-20
Add support for auto-compound in `kaspawallet send`

svarogg:
This PR fixes a big problem in kaspawallet - when the number of inputs is greater then one can fit into a single transaction (such as often happens when a big miner wants to send a lot of kaspa) - the transaction generated is rejected by the node.

This PR splits such transaction into multiple smaller transactions, sending the funds into a change address, then adds another, merge transaction, which would send the funds to their final destination.

Note: there are a quite a few redundant serialize/deserialize of PartiallySignedTransaction(s)
This is there because libkaspawallet exposes methods that work with byte-slices rather then PartiallySignedTransaction(s)
I plan to refactor libkaspawallet to expose only methods for deserialized data in an upcoming PR.


someone235 (changes requested):

cmd/kaspawallet/daemon/server/split_transaction.go
```diff
@@ -0,0 +1,199 @@
...
+	transactionMass, err := s.estimateMassAfterSignatures(transaction)
+	if err != nil {
+		return nil, err
+	}
+
+	if transactionMass < mempool.MaximumStandardTransactionMass {
+		return []*serialization.PartiallySignedTransaction{transaction}, nil
+	}
+
+	splitCount := int(transactionMass / mempool.MaximumStandardTransactionMass)
```
someone235, line 119: Let's say the tx has 100 inputs, 1 output, mass 1000, and MaximumStandardTransactionMass is 90.
So we'll try to split it to 12 transactions.
So if for example each input mass is 10, and the output mass is 10, there's be a split transaction which will have 9 inputs (`ceil(100/12)`) and 1 output, so the tx mass will be 100, which is above MaximumStandardTransactionMass

cmd/kaspawallet/daemon/server/split_transaction.go
```diff
@@ -70,29 +67,34 @@ func (s *server) mergeTransaction(splitTransactions []*serialization.PartiallySi
...
-	var payments []*libkaspawallet.Payment
-	if totalValue >= sentValue {
-		payments = []*libkaspawallet.Payment{{
-			Address: targetAddress,
-			Amount:  sentValue,
-		}, {
+	if totalValue < sentValue {
+		// sometimes the fees from compound transactions make the total output higher than what's available from selected
+		// utxos, in such cases - find one more UTXO and use it.
+		oneMoreUTXO, err := s.oneMoreUTXOForMergeTransaction(utxos, sentValue-totalValue)
```
someone235, line 79: Why do we assume that one UTXO should be enough?

someone235 (changes requested):

cmd/kaspawallet/daemon/server/split_transaction.go
```diff
@@ -0,0 +1,199 @@
...
+	if numOutputs > 2 || numOutputs == 0 {
+		return nil, errors.Errorf("original transaction has %d outputs, while 1 or 2 are expected",
+			len(originalTransaction.Tx.Outputs))
+	}
+
+	totalValue := uint64(0)
+	sentValue := originalTransaction.Tx.Outputs[0].Value
+	utxos := make([]*libkaspawallet.UTXO, len(splitTransactions))
+	for i, splitTransaction := range splitTransactions {
+		output := splitTransaction.Tx.Outputs[0]
```
stasatdaglabs, line 63: We don't appear to be merging the change. Is it not something we want to do at this stage?

stasatdaglabs (changes requested):

cmd/kaspawallet/daemon/server/split_transaction.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/kaspanet/kaspad/cmd/kaspawallet/libkaspawallet/serialization"
+	"github.com/kaspanet/kaspad/domain/consensus/model/externalapi"
+	"github.com/kaspanet/kaspad/domain/consensus/utils/consensushashing"
+	"github.com/kaspanet/kaspad/domain/consensus/utils/constants"
+	"github.com/kaspanet/kaspad/domain/consensus/utils/utxo"
+	"github.com/kaspanet/kaspad/domain/miningmanager/mempool"
+	"github.com/kaspanet/kaspad/util"
+)
+
+func (s *server) maybeAutoCompoundTransaction(transactionBytes []byte) ([][]byte, error) {
```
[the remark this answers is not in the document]
svarogg: I added a large comment for maybeAutoCompoundTransaction that I think does a good explanation. If you think it's insufficient - lmk.

cmd/kaspawallet/daemon/server/split_transaction.go
```diff
@@ -0,0 +1,194 @@
...
+	splitCount := int(transactionMass / mempool.MaximumStandardTransactionMass)
+	if transactionMass%mempool.MaximumStandardTransactionMass > 0 {
+		splitCount++
+	}
+	inputCountPerSplit := len(transaction.Tx.Inputs) / splitCount
+
+	splitTransactions := make([]*serialization.PartiallySignedTransaction, splitCount)
+	for i := 0; i < splitCount; i++ {
+		startIndex := i * inputCountPerSplit
+		endIndex := startIndex + inputCountPerSplit
```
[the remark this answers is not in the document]
svarogg:
```
	inputCountPerSplit := len(transaction.Tx.Inputs) / splitCount
	if len(transaction.Tx.Inputs)%splitCount > 0 {
		splitCount++
	}
```
You missed this part (note the logic is a bit weird, I'm incrementing `splitCount`, and not `inputCountPerSplit`, since incrementing `inputCountPerSplit` might make mass too high.
I'll add a comment to clarify that.

cmd/kaspawallet/daemon/server/create_unsigned_transaction.go
```diff
@@ -102,3 +102,32 @@ func (s *server) selectUTXOs(spendAmount uint64, feePerInput uint64) (
...
+
+	var utxo *walletUTXO
+	for _, utxo = range s.utxosSortedByAmount {
+		if !isUTXOSpendable(utxo, dagInfo.VirtualDAAScore, s.params.BlockCoinbaseMaturity) {
+			continue
+		}
+		if utxo.UTXOEntry.Amount() < requiredAmount+feePerInput {
+			return nil, errors.Errorf("Insufficient funds for merge transaction")
+		}
+		for _, alreadySelectedUTXO := range alreadySelectedUTXOs {
```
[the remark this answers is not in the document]
svarogg: I guess because that will require to implement a walletUTXO set, and that would be a premature optimization

cmd/kaspawallet/daemon/server/split_transaction.go
svarogg: The sum of the outputs of the splits is equal to the total of their inputs. This includes the change as well.

Change is re-calculated again in mergeTransaction:
```
	if totalValue > sentValue {
		payments = append(payments, &libkaspawallet.Payment{
			Address: changeAddress,
			Amount:  totalValue - sentValue,
		})
``` 
(Note the change is going to be a bit lower, because there's fee for the mergeTransaction)

cmd/kaspawallet/daemon/server/split_transaction.go
svarogg: I think this should be foolproof now. Please take a careful look in and around `splitAndInputPerSplitCounts`. (Also suggestions for a better name welcome).

cmd/kaspawallet/daemon/server/split_transaction.go
```diff
@@ -0,0 +1,199 @@
...
+	mergeTransactionBytes, err := libkaspawallet.CreateUnsignedTransaction(s.keysFile.ExtendedPublicKeys,
+		s.keysFile.MinimumSignatures, payments, utxos)
+	if err != nil {
+		return nil, err
+	}
+
+	return serialization.DeserializePartiallySignedTransaction(mergeTransactionBytes)
+}
+
+func (s *server) maybeSplitTransaction(transaction *serialization.PartiallySignedTransaction,
```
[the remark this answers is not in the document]
svarogg: I feel ya, and also @someone235 has defined said edge case, but I have a feeling what you suggest is an overkill.

It also will probably be quite annoying, clocking at O(n^2) with a possibly high n, and high constants (with cloning of the tx, etc.), in an operation that is supposed to be instantaneous.

I think the way I did it now is much more foolproof.
Please take a careful look in and around splitAndInputPerSplitCounts. (Also suggestions for a better name welcome).

cmd/kaspawallet/daemon/server/split_transaction.go
```diff
@@ -0,0 +1,258 @@
...
+			Outpoint:       utxo.Outpoint,
+			UTXOEntry:      utxo.UTXOEntry,
+			DerivationPath: s.walletAddressPath(utxo.address)})
+		totalValueAdded += utxo.UTXOEntry.Amount() - feePerInput
+		if totalValueAdded >= requiredAmount {
+			break
+		}
+	}
+	if totalValueAdded < requiredAmount {
+		return nil, 0, errors.Errorf("Insufficient funds for merge transaction")
```
stasatdaglabs, line 254: Do we really want to deny a user from sending funds just because they don't have enough money for a merge transaction?

cmd/kaspawallet/daemon/server/split_transaction.go
```diff
@@ -0,0 +1,258 @@
...
+			UTXOEntry:      utxo.NewUTXOEntry(output.Value, output.ScriptPublicKey, false, constants.UnacceptedDAAScore),
+			DerivationPath: s.walletAddressPath(changeWalletAddress),
+		}
+		totalValue += output.Value
+		totalValue -= feePerInput
+	}
+
+	if totalValue < sentValue {
+		// sometimes the fees from compound transactions make the total output higher than what's available from selected
+		// utxos, in such cases - find one more UTXO and use it.
```
stasatdaglabs, line 87: Please update this comment

cmd/kaspawallet/daemon/server/split_transaction.go
```diff
@@ -0,0 +1,258 @@
...
+		splitTransactions[i], err = s.createSplitTransaction(transaction, changeAddress, startIndex, endIndex)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return splitTransactions, nil
+}
+
+func (s *server) splitAndInputPerSplitCounts(transaction *serialization.PartiallySignedTransaction, transactionMass uint64) (
```
stasatdaglabs, line 144: For the sake of argument, let's assume our transaction contains:
* 9 pay to public key inputs, each one has mass 100
* 1 multisig input with mass 900
* Mass limit of 1000
* A transaction without any inputs has mass 50

I believe this method would produce:
* massWithoutInputs = 50
* massForInputs = 1800
* inputCount = 10
* massPerInput = 180
* inputPerSplitCount = 5
* splitCount = 2

The splitCount is correct, but if we try to have inputPerSplitCount of 5, we'd get:
* One transaction with mass 550
* Another transaction with mass **1350**

1350 is over the mass limit of 1000

stasatdaglabs (changes requested):

cmd/kaspawallet/daemon/server/split_transaction.go
someone235: What's the alternative?

cmd/kaspawallet/sign.go
```diff
@@ -1,63 +1,73 @@
...
 	}
-
 	privateKeys, err := keysFile.DecryptMnemonics(conf.Password)
 	if err != nil {
 		return err
 	}
 
-	updatedPartiallySignedTransaction, err := libkaspawallet.Sign(conf.NetParams(), privateKeys, partiallySignedTransaction, keysFile.ECDSA)
+	transactionsHex := conf.Transaction
+	if conf.TransactionFile != "" {
```
someone235, line 31: Can you call it TransactionsFile?

cmd/kaspawallet/daemon/server/split_transaction.go
```diff
@@ -197,3 +222,37 @@ func (s *server) estimateMassAfterSignatures(transaction *serialization.Partiall
...
+	dagInfo, err := s.rpcClient.GetBlockDAGInfo()
+	if err != nil {
+		return nil, 0, err
+	}
+
+	for _, utxo := range s.utxosSortedByAmount {
+		if !isUTXOSpendable(utxo, dagInfo.VirtualDAAScore, s.params.BlockCoinbaseMaturity) {
+			continue
+		}
+		for _, alreadySelectedUTXO := range alreadySelectedUTXOs {
```
someone235, line 238: Replying to
> I guess because that will require to implement a walletUTXO set, and that would be a premature optimization

You just need to convert it to a set of the type `map[externalapi.DomainOutpoint]struct{}`. No need to use `walletUTXOSet` or something like that.

cmd/kaspawallet/transactions_hex_encoding.go
```diff
@@ -0,0 +1,33 @@
...
+
+func encodeTransactionsToHex(transactions [][]byte) string {
+	transactionsInHex := make([]string, len(transactions))
+	for i, transaction := range transactions {
+		transactionsInHex[i] = hex.EncodeToString(transaction)
+	}
+	return strings.Join(transactionsInHex, hexTransactionsSeparator)
+}
+
+func decodeTransactionsFromHex(transactionsHex string) ([][]byte, error) {
```
someone235, line 20: nit: I think it's best to use a protobuf type that will hold an array of transactions, instead of parsing with separator manually

someone235 (changes requested):

cmd/kaspawallet/broadcast.go
```diff
@@ -28,27 +28,28 @@ func broadcast(conf *broadcastConfig) error {
...
-	response, err := daemonClient.Broadcast(ctx, &pb.BroadcastRequest{Transaction: transaction})
-	if err != nil {
-		return err
+	for _, transaction := range transactions {
+		response, err := daemonClient.Broadcast(ctx, &pb.BroadcastRequest{Transaction: transaction})
+		if err != nil {
+			return err
+		}
+		fmt.Println("Transaction was sent successfully")
+		fmt.Printf("Transaction ID: \t%s\n", response.TxID)
```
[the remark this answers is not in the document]
svarogg: Did something a little bit different, see what I did and lmk if you no like

cmd/kaspawallet/daemon/server/split_transaction.go
svarogg: Not having funds for merge transaction is more-or-less the same as not having enough fees, just send a little bit less and you'll be fine.

cmd/kaspawallet/daemon/server/split_transaction.go
svarogg: We assume that the transaction was created by kaspawallet, thus it wouldn't have inputs with different masses.
Therefore your scenario is impossible.

cmd/kaspawallet/broadcast.go
```diff
@@ -28,27 +28,28 @@ func broadcast(conf *broadcastConfig) error {
 		return errors.Errorf("Both --transaction and --transaction-file cannot be passed at the same time")
 	}
 
-	transactionHex := conf.Transaction
+	transactionsHex := conf.Transaction
```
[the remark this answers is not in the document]
svarogg: Did that (also for TransactionsFile), but left the cli flag to still say "--transaction" in singular, because the user should be transparent to the split at this point.

cmd/kaspawallet/daemon/server/split_transaction.go
```diff
@@ -0,0 +1,259 @@
...
+
+	massForInputs := transactionMass - massWithoutInputs
+
+	inputCount := len(transaction.Tx.Inputs)
+	massPerInput := massForInputs / uint64(inputCount)
+	if massForInputs%uint64(inputCount) > 0 {
+		massPerInput++
+	}
+
+	inputPerSplitCount = int((mempool.MaximumStandardTransactionMass - massWithoutInputs) / massPerInput)
```
someone235, line 158: Maybe rename to input**s**PerSplitCount?

someone235: Each split tx has only one output right? So wouldn't it make more sense to reduce the mass of one output, instead of all outputs?

someone235 (changes requested):

cmd/kaspawallet/daemon/server/split_transaction.go
svarogg: I've added a comment.
Adding a sanity test means either refactoring the CalculateMass code to have CalculateMassForInput, or using some not-entirely-accurate proxy such as input size, and I'd rather not do either.

stasatdaglabs (dismissed):

cmd/kaspawallet/daemon/server/split_transaction.go
someone235: You can check if all inputs are multisig with the same number of required signatures, or all of them are p2pk

cmd/kaspawallet/daemon/server/split_transaction.go
someone235: Using the outputs of the original transaction as an upper bound is confusing from the reader POV (why use such outputs that are unrelated to the actual split transaction), and the calculation might go wrong: e.g. P2SH is 34 bytes long and P2PK is 33 bytes long, so when you're using a multisig wallet and wanna pay to P2PK, and say your original transaction had one output, the output of the split transaction will use a P2SH address, which requires more mass, so your split tx can go above limit.

So if you don't want to export another function for calculating output mass, can't you just make a stub output that will use some change address and then calculate the output mass in the same way you calculated `massForInputs`?

someone235 (approved):

PR #1953, opened 2022-02-20
Implement a `parse` sub command in the walllet

stasatdaglabs:
Parses and prints details about a transaction. The output looks like this:

photo5805693648795449514


cmd/kaspawallet/parse.go
```diff
@@ -0,0 +1,82 @@
...
+	allOutputSompi := uint64(0)
+	for index, output := range partiallySignedTransaction.Tx.Outputs {
+		scriptPublicKeyType, scriptPublicKeyAddress, err := txscript.ExtractScriptPubKeyAddress(output.ScriptPublicKey, conf.ActiveNetParams)
+		if err != nil {
+			return err
+		}
+
+		addressString := scriptPublicKeyAddress.EncodeAddress()
+		if scriptPublicKeyType == txscript.NonStandardTy {
+			addressString = hex.EncodeToString(output.ScriptPublicKey.Script)
```
someone235, line 69: Can you give some indication that this is not an address? Maybe write "<Unparseable scipt pub key: %s>" or something like this?

someone235 (changes requested):

someone235 (approved):

PR #1962, opened 2022-03-01
Optimize expected header pruning point

someone235:


domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -995,7 +995,10 @@ func (pm *pruningManager) ExpectedHeaderPruningPoint(stagingArea *model.StagingA
 		return nil, err
 	}
 
-	if hasPruningPointInItsSelectedChain && pm.finalityScore(ghostdagData.BlueScore()) > pm.finalityScore(selectedParentPruningPointHeader.BlueScore()+pm.pruningDepth) {
+	minRequiredBlueScoreForNextPruningPoint := (pm.finalityScore(selectedParentPruningPointHeader.BlueScore()) + 1) * pm.finalityInterval
+
+	if hasPruningPointInItsSelectedChain &&
+		minRequiredBlueScoreForNextPruningPoint+pm.pruningDepth <= ghostdagData.BlueScore() {
```
michaelsutton, line 1001: Can you write here the formula for the next pruning point and reference more places in the codebase that use it ?

michaelsutton: Extended discussion: https://research.kas.pa/t/pruning-point-heuristic-formula/127

michaelsutton (approved):

domain/consensus/processes/pruningmanager/pruningmanager.go
michaelsutton: The change was verified to be safe

Issue #1957, opened 2022-02-23
Kaspawallet - Force "send transaction" from specific address


thesheepcat:
Currently, kaspawallet does not allow users to spend specific funds on a specific address when creating a transaction.

I suggest to add the possibility to force a specific address when sending funds.

Example command: `./kaspawallet send -t RECEIVING_ADDRESS -a FROM_ADDRESS -v AMOUNT`

In case the address does not have enough funds, an error is generated.

Eventually, in the future, multiple addresses can be passed as parameter and the wallet knows which addresses can take the coins from.

PR #1964, opened 2022-03-03
Add support for `from` address in `kaspawallet send`

tmrlvi:
- Add an option to choose the from address in a transactions when sending KAS
- Do not use change address is change is 0 (transaction with 0 output is not accepted by kaspad anyway)


tmrlvi:
This solves Issue https://github.com/kaspanet/kaspad/issues/1957

cmd/kaspawallet/daemon/server/create_unsigned_transaction.go
```diff
@@ -28,9 +28,17 @@ func (s *server) CreateUnsignedTransaction(_ context.Context, request *pb.Create
 		return nil, err
 	}
 
+	var fromAddress util.Address = nil
```
someone235, line 31: nit: The `=nil` here is redundant

cmd/kaspawallet/config.go
```diff
@@ -54,13 +54,15 @@ type sendConfig struct {
 	Password      string  `long:"password" short:"p" description:"Wallet password"`
 	DaemonAddress string  `long:"daemonaddress" short:"d" description:"Wallet daemon server to connect to (default: localhost:8082)"`
 	ToAddress     string  `long:"to-address" short:"t" description:"The public address to send Kaspa to" required:"true"`
+	FromAddress   string  `long:"from-address" short:"a" description:"Specific public address to send Kaspa from" required:"false"`
```
someone235, line 57: If it's not a big deal, I would prefer to directly implement it with multiple addresses (`FromAddresses`)

cmd/kaspawallet/daemon/server/create_unsigned_transaction.go
```diff
@@ -68,7 +80,9 @@ func (s *server) selectUTXOs(spendAmount uint64, feePerInput uint64) (
 	}
 
 	for _, utxo := range s.utxosSortedByAmount {
-		if !isUTXOSpendable(utxo, dagInfo.VirtualDAAScore, s.params.BlockCoinbaseMaturity) {
+		addr, err := s.walletAddressString(utxo.address)
+		if err != nil || (from != nil && addr != from.String()) ||
```
someone235, line 84: Why do you swallow the error?

someone235 (changes requested):
In addition to what I wrote in the comments, I think it's better to wait for https://github.com/kaspanet/kaspad/pull/1951 to be merged so the compounded transaction will also be created only from the `FromAddresses`.

cmd/kaspawallet/daemon/server/create_unsigned_transaction.go
tmrlvi: I thought to ignore invalid UTXOs (which cannot be formatted to address). Maybe a better approach would be to convert `from` to `walletAddress` type and explicitly throw an error if failing. I'll look into it

tmrlvi:
I've updated the coded with changes to `FromAddresses` (plural) 
I'm still testing (don't merge yet)

tmrlvi:
Fixed and tested

tmrlvi:
Discussion point: currently, I implement multiple addresses using comma, which make argument parsing a bit cumbersome. There are two option to make it better (that I can think of): 

1. Use `go-flags` native option for multiple values (repeating flags)
2. Implement custom parser for `go-flags`

Do you think we should change the parsing, or is the way it is written now ok?

cmd/kaspawallet/daemon/server/create_unsigned_transaction.go
```diff
@@ -83,7 +98,8 @@ func (s *server) selectUTXOs(spendAmount uint64, feePerInput uint64) (
 	}
 
 	for _, utxo := range s.utxosSortedByAmount {
-		if !isUTXOSpendable(utxo, dagInfo.VirtualDAAScore, s.params.BlockCoinbaseMaturity) {
+		if (fromAddresses != nil && !slices.Contains(fromAddresses, utxo.address)) ||
```
someone235, line 101: Nice! First use of generics in the codebase!

someone235:
> Discussion point: currently, I implement multiple addresses using comma, which make argument parsing a bit cumbersome. There are two option to make it better (that I can think of):
> 
> 1. Use `go-flags` native option for multiple values (repeating flags)
> 2. Implement custom parser for `go-flags` (as described in https://stackoverflow.com/a/28323276/3051715)
> 
> Do you think we should change the parsing, or is the way it is written now ok?

I think it's better to stick with the `go-flags` standard

tmrlvi:
> > Discussion point: currently, I implement multiple addresses using comma, which make argument parsing a bit cumbersome. There are two option to make it better (that I can think of):
> > 
> > 1. Use `go-flags` native option for multiple values (repeating flags)
> > 2. Implement custom parser for `go-flags` (as described in https://stackoverflow.com/a/28323276/3051715)
> > 
> > Do you think we should change the parsing, or is the way it is written now ok?
> 
> I think it's better to stick with the `go-flags` standard

Ok, so I'm changing `--from-addresses` to `--from-address`, and allow for multiple uses. Example:
```
./kaspawallet create-unsigned-transaction -a <from_addr_1> -a <from_addr_2> -t <to_addr> -v <value>
```

someone235 (approved):

PR #1978, opened 2022-03-14
Improve output of non-critical protocol errors to avoid user panic

michaelsutton:
This PR improves log output. It avoids the full verbose error in the case of ping timeout, and frames the call-stack output in other cases. It looks like this:

2022-03-15 00:45:35.974 [INF] PROT: Non-critical peer protocol error from '', printing the full stack for debug purposes: 
        ====================================================
        timeout expired
        route '' got timeout after 2m0s
        github.com/kaspanet/kaspad/app/protocol/flows/v5/ping.TestPingErr
        	C:/Users/msssu/Documents/Git/kaspanet/kaspad/app/protocol/flows/v5/ping/ping_error_test.go:13
        testing.tRunner
        	C:/Program Files/Go/src/testing/testing.go:1259
        runtime.goexit
        	C:/Program Files/Go/src/runtime/asm_amd64.s:1581 
        ====================================================

### DIFF
--- a/app/protocol/flowcontext/errors.go
+++ b/app/protocol/flowcontext/errors.go
@@ -2,11 +2,17 @@ package flowcontext
 
 import (
 	"errors"
+	"strings"
 	"sync/atomic"
 
 	"github.com/kaspanet/kaspad/infrastructure/network/netadapter/router"
 
 	"github.com/kaspanet/kaspad/app/protocol/protocolerrors"
+)
+
+var (
+	// ErrPingTimeout signifies that a ping operation timed out.
+	ErrPingTimeout = protocolerrors.New(false, "timeout expired on ping")
 )
 
 // HandleError handles an error from a flow,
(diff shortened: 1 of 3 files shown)


app/protocol/flows/v4/blockrelay/ibd_with_headers_proof.go
```diff
@@ -24,6 +24,7 @@ func (flow *handleIBDFlow) ibdWithHeadersProof(highHash *externalapi.DomainHash,
 			return err
 		}
 
+		log.Infof("IBD with pruning proof from %s was unsuccessful. Deleting the staging consensus.", flow.peer)
```
michaelsutton, line 27: Right after this, the user sees the line `Deleting database prefix &{0}` which seems worrying. We want to explain first.

app/protocol/flows/v4/blockrelay/ibd_with_headers_proof.go
```diff
@@ -32,6 +33,8 @@ func (flow *handleIBDFlow) ibdWithHeadersProof(highHash *externalapi.DomainHash,
 		return err
 	}
 
+	log.Infof("Header download stage of IBD with pruning proof completed successfully from %s. "+
```
michaelsutton, line 36: Same here. Sometimes the consensus commit triggers a previous consensus delete with the line `Deleting database prefix &{0}`. We should explain first.

domain/consensus/processes/syncmanager/antipast.go
```diff
@@ -145,15 +145,15 @@ func (sm *syncManager) missingBlockBodyHashes(stagingArea *model.StagingArea, hi
 		lowHash = selectedChild
 	}
 	if !foundHeaderOnlyBlock {
-		if lowHash == highHash {
+		if lowHash.Equal(highHash) {
```
michaelsutton, line 148: I found a case where the error log line below was falsely triggered due to the wrong equality test

svarogg (approved):

PR #1986, opened 2022-03-23
Add names to nameless routes

Aleoami:
Adding names to nameless routes of capacity of 131172 to avoid messages like "route '' reached capacity of 131172" where it's unclear what exactly this route is.


infrastructure/network/netadapter/router/router.go
```diff
@@ -24,10 +24,10 @@ type Router struct {
 }
 
 // NewRouter creates a new empty router
-func NewRouter() *Router {
+func NewRouter(name string) *Router {
 	router := Router{
 		incomingRoutes: make(map[appmessage.MessageCommand]*Route),
-		outgoingRoute:  newRouteWithCapacity("", outgoingRouteMaxMessages),
+		outgoingRoute:  newRouteWithCapacity(name, outgoingRouteMaxMessages),
```
someone235, line 30: Can you change it to `fmt.Sprintf("%s - outgoing route")` or something like this, so we can know this route is an outgoing route?

someone235 (changes requested):

infrastructure/network/netadapter/router/router.go
Aleoami: This will mean then that every route created via NewRoute() is the outgoing one, right?

Aleoami: Done

someone235 (approved):

PR #1988, opened 2022-03-28
Optimize the miner-kaspad flow and latency

michaelsutton:
This PR includes initial changes to improve the miner-to-kaspad communication cycle, including the following:

- A dedicated "new block template" event for notifying miners only when necessary

A following PR will include additional improvements


app/appmessage/message.go
```diff
@@ -156,6 +156,9 @@ const (
 	CmdVirtualDaaScoreChangedNotificationMessage
 	CmdGetBalancesByAddressesRequestMessage
 	CmdGetBalancesByAddressesResponseMessage
+	CmdNotifyNewBlockTemplateRequestMessage
```
someone235, line 159: What about StopNotify?

someone235 (approved):

app/appmessage/message.go
michaelsutton: It's not implemented for most events. But we should add it eventually.

PR #1996, opened 2022-03-31
New definition for "out of sync"

michaelsutton:
This PR aims to modify "out of sync" definition such that miners will mine under all possible circumstances (wip, under discussion)

### DIFF
--- a/app/protocol/flowcontext/should_mine.go
+++ b/app/protocol/flowcontext/should_mine.go
@@ -2,21 +2,12 @@ package flowcontext
 
 import "github.com/kaspanet/kaspad/util/mstime"
 
-const (
-	maxSelectedParentTimeDiffToAllowMiningInMilliSeconds = 60 * 60 * 1000 // 1 Hour
-)
-
-// ShouldMine returns whether it's ok to use block template from this node
-// for mining purposes.
-func (f *FlowContext) ShouldMine() (bool, error) {
+// IsNearlySynced returns whether this node is considered synced or close to being synced. This info
+// is used to determine if it's ok to use a block template from this node for mining purposes.
+func (f *FlowContext) IsNearlySynced() (bool, error) {
 	peers := f.Peers()
 	if len(peers) == 0 {
-		log.Debugf("The node is not connected, so ShouldMine returns false")
-		return false, nil
-	}
-
-	if f.IsIBDRunning() {
-		log.Debugf("IBD is running, so ShouldMine returns false")
+		log.Debugf("The node is not connected to peers, so IsNearlySynced returns false")
 		return false, nil
 	}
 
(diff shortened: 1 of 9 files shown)


app/protocol/flowcontext/should_mine.go
```diff
@@ -2,21 +2,12 @@ package flowcontext
...
-const (
-	maxSelectedParentTimeDiffToAllowMiningInMilliSeconds = 60 * 60 * 1000 // 1 Hour
-)
-
-// ShouldMine returns whether it's ok to use block template from this node
-// for mining purposes.
-func (f *FlowContext) ShouldMine() (bool, error) {
+// IsNearlySynced returns whether this node is considered synced or close to being synced. This info
+// is used to determine if it's ok to use a block template from this node for mining purposes.
+func (f *FlowContext) IsNearlySynced() (bool, error) {
```
michaelsutton, line 7: Not sure how to call this function.

app/protocol/flowcontext/should_mine.go
```diff
@@ -35,13 +26,15 @@ func (f *FlowContext) ShouldMine() (bool, error) {
 	}
 
 	now := mstime.Now().UnixMilliseconds()
```
michaelsutton, line 28: The queries to virtual selected parent info are relatively expensive. It might make sense to maintain the `virtual selected parent timestamp` by pushing virtual changes and only querying the current time against it with out acquiring the consensus lock.

app/protocol/flows/v4/transactionrelay/handle_relayed_transactions.go
```diff
@@ -50,7 +50,12 @@ func (flow *handleRelayedTransactionsFlow) start() error {
 			return err
 		}
 
-		if flow.IsIBDRunning() {
+		isNearlySynced, err := flow.IsNearlySynced()
```
michaelsutton, line 53: This might be a performance-sensitive location, if we don't optimize `IsNearlySynced` using push updates as suggested above.

app/protocol/flows/v4/blockrelay/handle_relay_invs.go
```diff
@@ -106,9 +107,13 @@ func (flow *handleRelayInvsFlow) start() error {
...
 
-		// Block relay is disabled during IBD
-		if flow.IsIBDRunning() {
-			log.Debugf("Got block %s while in IBD. continuing...", inv.Hash)
+		// Block relay is disabled if the node is out of sync and is already during IBD
+		isNearlySynced, err := flow.IsNearlySynced()
+		if err != nil {
+			return err
+		}
+		if !isNearlySynced && flow.IsIBDRunning() {
```
someone235, line 115: IsIBDRunning should be pretty fast, so I think it's better to check it first

app/protocol/flows/v4/transactionrelay/handle_relayed_transactions.go
someone235: Maybe it's fine to keep using `IsIBDRunning` in here

someone235 (changes requested):

someone235 (approved):

Issue #1990, opened 2022-03-28
Add "sweep" command to kaspawallet


someone235:
Add a `sweep` command where you can pass in `--private-key` a hex private key in the same format that is used by `genkeypair`. The command will use all of the private key funds and send them to a change address of the wallet.
A similar example can be seen here: https://github.com/svarogg/private-key-wallet

michaelsutton:
Done at #2018

PR #2018, opened 2022-04-13
Adds a "sweep" command to `kaspawallet`

D-Stacks:
Implements #1990 

Overveiw:

1) allows for broadcasting of serialized `externalapi.DomainTransaction`s
2) allows the kaspawallet server to query UTXOs of external addresses
3) implements logic of: extracting a associated address, as well as, utxos from a schnorr private key. Creates a new wallet address. Creates, signs and splits transaction(s) with a schnorr private key, and finnally broadcasts it.


cmd/kaspawallet/sweep.go
```diff
@@ -0,0 +1,267 @@
...
+	"github.com/kaspanet/kaspad/util"
+	"github.com/kaspanet/kaspad/util/txmass"
+	"github.com/pkg/errors"
+)
+
+const feePerInput = 10000
+
+func sweep(conf *sweepConfig) error {
+
+	//Sweep assumes the passed private key is a schnorr private key: I see no function to explictly test the type of key
```
someone235, line 31: Please delete `I see no function to explictly test the type of key`, and generally avoid and `I` phrases. Such comments could be added as a comment to the PR if you want the reviewer to read it.

cmd/kaspawallet/sweep.go
```diff
@@ -0,0 +1,267 @@
...
+	//It will almost certainly fail somewhere if this is not the case
+	privateKeyBytes, err := hex.DecodeString(conf.PrivateKey)
+	if err != nil {
+		return err
+	}
+
+	if err != nil {
```
someone235, line 38: this `if` is redundant

cmd/kaspawallet/sweep.go
```diff
@@ -0,0 +1,267 @@
...
+		return err
+	}
+	publicKeybytes, err := libkaspawallet.PublicKeyFromPrivateKey(privateKeyBytes)
+	if err != nil {
+		return err
+	}
+
+	//NewAddress might seem confusing, but the function is entirely deterministic based on the public key
```
someone235, line 46: Please delete this comment. If you think the name is misleading you can change it from `NewAddressPublicKey` to `AddressPublicKeyFromPublicKey` or something like this (but then you should also change `NewAddressPublicKeyECDSA` etc to keep the same convention)

cmd/kaspawallet/sweep.go
```diff
@@ -0,0 +1,267 @@
...
+	if err != nil {
+		return err
+	}
+
+	totalExtracted := uint64(0)
+
+	fmt.Println("\nTransaction ID(s):")
+	for i, txID := range response.TxIDs {
+		fmt.Printf("\t%s\n", txID)
+		fmt.Println("\tSweeped:\t", utils.FormatKas(splitTransactions[i].Outputs[0].Value), " KAS")
```
someone235, line 126: I think `Swept` is the correct form

cmd/kaspawallet/sweep.go
```diff
@@ -0,0 +1,267 @@
...
+	totalSplitAmount := uint64(0)
+
+	ScriptPublicKey, err := txscript.PayToAddrScript(toAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	dummyTransactionWindow := make([]*externalapi.DomainTransaction, 2)
+	//[0] is the last build that didn't violate mass, that can be added as a split
+	dummyTransactionWindow[0] = newDummyTransaction()
```
someone235, line 174: No slice properties are actually being used, so maybe it's better to change this to `dummyTransaction1` and `dummyTransaction2` (although I would use more meaningful names, like `lastTriedValidTx` and `workInProgressTx` or something like this)

cmd/kaspawallet/sweep.go
```diff
@@ -0,0 +1,267 @@
...
+
+func createSplitTransactionsWithSchnorrPrivteKey(
+	params *dagconfig.Params,
+	selectedUTXOs []*libkaspawallet.UTXO,
+	toAddress util.Address,
+	feePerInput int) ([]*externalapi.DomainTransaction, error) {
+
+	var splitTransactions []*externalapi.DomainTransaction
+
+	// I need to add extra mass to transaction, txmasscalculater seems to undercalculate, by around this amount.
```
someone235, line 156: It doesn't undercalculate. You just don't take signatures into account. Please rephrase the comment to reflect it. But anyway, please try to use `we` instead of `I`.

cmd/kaspawallet/sweep.go
```diff
@@ -0,0 +1,267 @@
...
+
+		if massCalculater.CalculateTransactionMass(dummyTransactionWindow[1])+extraMass >= mempool.MaximumStandardTransactionMass {
+			splitTransactions = append(splitTransactions, dummyTransactionWindow[0])
+			currentIdxOfSplit = 0
+			totalSplitAmount = 0
+			totalAmount = totalAmount + currentAmount
+			dummyTransactionWindow[0] = newDummyTransaction()
+			dummyTransactionWindow[1] = newDummyTransaction()
+			if i == len(selectedUTXOs)-1 {
+				splitTransactions = append(splitTransactions, dummyTransactionWindow[1])
```
someone235, line 218: I don't get it. This is a newly created dummy transactions. How can you append it to `splitTransactions`?

cmd/kaspawallet/sweep.go
```diff
@@ -0,0 +1,267 @@
...
+				break
+			}
+			continue
+		}
+
+		//Special case, end of inputs, with no violation, where we can assign dummyWindow[1] to split and break
```
someone235, line 224: There's no such thing as `dummyWindow[1]`

cmd/kaspawallet/sweep.go
```diff
@@ -0,0 +1,267 @@
...
+	massCalculater := txmass.NewCalculator(params.MassPerTxByte, params.MassPerScriptPubKeyByte, params.MassPerSigOp)
+
+	//for refernece this keeps track in respect to dummyWindow[0], not dummyWindow[1]
+	currentIdxOfSplit := 0
+
+	totalAmount := uint64(0)
+
+	totalSplitAmount := uint64(0)
+
+	ScriptPublicKey, err := txscript.PayToAddrScript(toAddress)
```
someone235, line 167: Starting a local variable with a capital letter makes it look like an exported variable, which is kinda weird. It's better to start only exported variables with a capital letter

cmd/kaspawallet/sweep.go
```diff
@@ -0,0 +1,267 @@
...
+	toAddress util.Address,
+	feePerInput int) ([]*externalapi.DomainTransaction, error) {
+
+	var splitTransactions []*externalapi.DomainTransaction
+
+	// I need to add extra mass to transaction, txmasscalculater seems to undercalculate, by around this amount.
+	extraMass := uint64(7000)
+	massCalculater := txmass.NewCalculator(params.MassPerTxByte, params.MassPerScriptPubKeyByte, params.MassPerSigOp)
+
+	//for refernece this keeps track in respect to dummyWindow[0], not dummyWindow[1]
```
someone235, line 160: Please rename `dummyWindow` in the comment

cmd/kaspawallet/sweep.go
```diff
@@ -0,0 +1,267 @@
...
+		currentIdxOfSplit++
+
+		dummyTransactionWindow[0] = dummyTransactionWindow[1].Clone()
+		dummyTransactionWindow[1].Outputs = make([]*externalapi.DomainTransactionOutput, 1)
+
+	}
+	return splitTransactions, nil
+}
+
+func signWithSchnorrPrivteKey(params *dagconfig.Params, privateKeyBytes []byte, domainTransactions []*externalapi.DomainTransaction) ([][]byte, error) {
```
someone235, line 240: `Privte` -> `Private`

someone235 (changes requested):

cmd/kaspawallet/sweep.go
```diff
@@ -153,43 +146,38 @@ func createSplitTransactionsWithSchnorrPrivteKey(
...
-	dummyTransactionWindow[0] = newDummyTransaction()
-	//[1] represents the tested unsigned transaction
-	dummyTransactionWindow[1] = newDummyTransaction()
+	lastValidTx := newDummyTransaction()
+	currentTx := newDummyTransaction() //i.e. the tested tx
 
 	//loop through utxos commit segments that don't violate max mass
 	for i, currentUTXO := range selectedUTXOs {
 
 		if currentIdxOfSplit == 0 {
```
someone235, line 170: I think this variable is actually doing nothing. On the first iteration `lastValidTx` and `currentTx` are actually fresh and empty, and you reset them every time you set `currentIdxOfSplit` to `0`

someone235 (changes requested):

cmd/kaspawallet/sweep.go
D-Stacks: `currentIdxOfSplit`? this is needed to reset `lastValidTx` and `currentTx`, for a new split transaction. I agree it doen't do anything during the first iteration, so if you want I can do an extra check with `&& i != 0` before entering the scope. But it is needed to reset the Txs after each split, excluding the first.

someone235 (approved):

PR #2023, opened 2022-04-15
Block template cache improvement

biospb:
* Avoid concurrent calls to template builder
* Clear cache on new block event
* Cache ShouldMine call


domain/miningmanager/miningmanager.go
```diff
@@ -21,6 +21,7 @@ type MiningManager interface {
 	ValidateAndInsertTransaction(transaction *externalapi.DomainTransaction, isHighPriority bool, allowOrphan bool) (
 		acceptedTransactions []*externalapi.DomainTransaction, err error)
 	RevalidateHighPriorityTransactions() (validTransactions []*externalapi.DomainTransaction, err error)
+	RemoveTransactions(txs []*externalapi.DomainTransaction, removeRedeemers bool) error
```
michaelsutton, line 24: Seems like this change slipped in here (which is why factory.go seems to have a compilation error). 

Also, please make sure `go fmt` is not required for miningmanager.go, the tests seem to fail on something there.

michaelsutton (dismissed):

michaelsutton (approved):

Issue #2017, opened 2022-04-13
`rpc error: code = Unknown desc = server is not synced` - when too many new / change addresses are created?


D-Stacks:
Both I and @Aleoami have noticed that when spamming the network with Txs, the wallet deamon will have a hard time syncing. the `isSynced` function for the wallet server references both external and internal indexes in order to decided when it is synced. I therefore deduce that this is somehow caused by creating corpious amounts of new, and/or change addresses. I havn't dug deeper into the code, but I wish to document this issue.

D-Stacks:
although this was partially expected behaviour, sync speed was fixed by #2025 and sync state was displayed by #2045

PR #2025, opened 2022-04-15
Wallet synchronization improvement

biospb:
* Much faster sync on startup
* Avoid double scan of the same address range


D-Stacks:
nice catch! I suspect this might be related to #2017, wanted to look into this next.

biospb:
yes. this is fix for it

cmd/kaspawallet/daemon/server/sync.go
```diff
@@ -103,16 +112,24 @@ func (s *server) maxUsedIndex() uint32 {
...
+		err := s.collectAddressesWithLock(index, index+1000)
 		if err != nil {
 			return err
 		}
+		maxUsedIndex = s.maxUsedIndex()
+	}
+
+	s.lock.Lock()
+	if index > s.nextSyncStartIndex {
+		s.nextSyncStartIndex = index
```
someone235, line 130: Maybe the name doesn't reflect it, but `s.nextSyncStartIndex` is used only for `collectFarAddresses`, so I don't understand why would you want to reset it to the last checked recent address

someone235 (changes requested):

cmd/kaspawallet/daemon/server/sync.go
biospb: it is the main fix of slow sync. as it used in isSynced() 
otherwise we need to wait until collectFarAddresses() will reach it

someone235 (dismissed):

someone235 (approved):

PR #2026, opened 2022-04-16
Fix RPC connections counting

biospb:
* show incomming connections count


infrastructure/network/netadapter/server/grpcserver/grpc_server.go
```diff
@@ -115,23 +116,23 @@ func (s *gRPCServer) handleInboundConnection(ctx context.Context, stream grpcStr
...
 }
 
-func (s *gRPCServer) incrementInboundConnectionCountAndLimitIfRequired() error {
+func (s *gRPCServer) incrementInboundConnectionCountAndLimitIfRequired() (int, error) {
 	s.inboundConnectionCountLock.Lock()
 	defer s.inboundConnectionCountLock.Unlock()
 
 	if s.maxInboundConnections > 0 && s.inboundConnectionCount == s.maxInboundConnections {
-		return errors.Errorf("limit of %d inbound connections has been exceeded", s.maxInboundConnections)
+		log.Warnf("Limit of %d inbound connections has been exceeded", s.maxInboundConnections)
```
someone235, line 130: This is pretty normal, so I wouldn't use `Warnf`. `Debugf` Seems to fit.

someone235 (changes requested):

infrastructure/network/netadapter/server/grpcserver/grpc_server.go
biospb: it is ultimately critical for mining nodes. as you can see this msg only in miner or in debug log mode, which is not an option on mining node...

michaelsutton (approved):

infrastructure/network/netadapter/server/grpcserver/grpc_server.go
michaelsutton: I think it is important for a node operator to notice he has too many RPC clients

svarogg (dismissed):

michaelsutton (approved):

someone235 (approved):

PR #2029, opened 2022-04-17
Fixing parse to handle conditional transactions

tmrlvi:
Small fix in `kaspawallet parse` to account for conditional transactions


tmrlvi:
I just saw there was the same PR already

PR #2031, opened 2022-04-20
GetBlockTemplate: Add Donation mechanism.

D-Stacks:
Regarding community wishes to add donation addresses to kaspads

note: 
~~1) untested~~
2) perhaps I still need better parsing of the flags / config.
3) alternative solution to broadcast donations might be to include it into the getInfo rpc call, wich would only needs to be queried once, instead of being supplied with every block template.


D-Stacks:
- closed

PR #2033, opened 2022-04-20
Prevent memory leaking in binaryserializer and domainhash

biospb:
ByteSlice() needs some extra checking

### DIFF
--- a/domain/consensus/model/externalapi/hash.go
+++ b/domain/consensus/model/externalapi/hash.go
@@ -65,16 +65,13 @@ func (hash DomainHash) String() string {
 }
 
 // ByteArray returns the bytes in this hash represented as a byte array.
-// The hash bytes are cloned, therefore it is safe to modify the resulting array.
 func (hash *DomainHash) ByteArray() *[DomainHashSize]byte {
-	arrayClone := hash.hashArray
-	return &arrayClone
+	return &hash.hashArray
 }
 
 // ByteSlice returns the bytes in this hash represented as a byte slice.
-// The hash bytes are cloned, therefore it is safe to modify the resulting slice.
 func (hash *DomainHash) ByteSlice() []byte {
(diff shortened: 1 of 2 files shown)


util/binaryserializer/binaryserializer.go
```diff
@@ -113,19 +113,21 @@ func PutUint16(w io.Writer, val uint16) error {
 // buffer from the free list and writes the resulting four bytes to the given
 // writer.
 func PutUint32(w io.Writer, val uint32) error {
-	var buf [4]byte
-	binary.LittleEndian.PutUint32(buf[:], val)
-	_, err := w.Write(buf[:])
+	buf := Borrow()[:4]
```
someone235, line 116: Can you explain how does it leak memory? `buf` is an array, so it should be stored entirely on the stack. I wouldn't expect it to leak.

domain/consensus/model/externalapi/hash.go
```diff
@@ -65,16 +65,13 @@ func (hash DomainHash) String() string {
 }
 
 // ByteArray returns the bytes in this hash represented as a byte array.
-// The hash bytes are cloned, therefore it is safe to modify the resulting array.
 func (hash *DomainHash) ByteArray() *[DomainHashSize]byte {
-	arrayClone := hash.hashArray
-	return &arrayClone
+	return &hash.hashArray
```
someone235, line 69: Which kind of leak does it fix? Is it a temporary leak that is cleaned by the gc or is it permanent?

util/binaryserializer/binaryserializer.go
someone235: And also, `Borrow` is pretty discouraged because it uses a `select` which can be quite slow. It's better to get rid of it where possible.

someone235 (changes requested):

PR #2034, opened 2022-04-20
Improved staging shard performance

biospb:
using map instead of extremely growing array


domain/consensus/model/staging_area.go
```diff
@@ -51,8 +52,16 @@ func (sa *StagingArea) Commit(dbTx DBTransaction) error {
 		return errors.New("Attempt to call Commit on already committed stagingArea")
 	}
 
-	for _, shard := range sa.shards {
-		if shard == nil { // since sa.shards is an array and not a map, some shard slots might be empty.
+	// order by ID
```
someone235, line 55: No need to order

domain/consensus/model/staging_area.go
```diff
@@ -51,8 +52,16 @@ func (sa *StagingArea) Commit(dbTx DBTransaction) error {
...
+	var keys []StagingShardID
+	for id := range sa.shards {
+		keys = append(keys, id)
+	}
+	sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })
+
+	for _, id := range keys {
+		shard, ok := sa.shards[id]
+		if !ok || shard == nil {
```
someone235, line 64: No need to compare to nil

someone235 (changes requested):

someone235 (approved):

someone235 (approved):

PR #2036, opened 2022-04-26
Add AcceptedTransactionIDs to ChainChanged notification and VirtualSelectedParentChain RPC

svarogg:
This PR was requested by TxBit, since it's complicated for them to follow balances tracking UTXO changes.

And example usage can be seen at: https://github.com/svarogg/kaspadotnet/blob/main/Kaspadotnet.Examples/FollowTransactions.cs


app/rpc/manager.go
```diff
@@ -196,7 +196,7 @@ func (m *Manager) notifyVirtualSelectedParentChainChanged(virtualChangeSet *exte
 	defer onEnd()
 
 	notification, err := m.context.ConvertVirtualSelectedParentChainChangesToChainChangedNotificationMessage(
-		virtualChangeSet.VirtualSelectedParentChainChanges)
+		virtualChangeSet.VirtualSelectedParentChainChanges, true)
```
someone235, line 199: You're populating the notification with tx data even if the user didn't subscribe to RPC notifications. I guess it could be a heavy operation.

testing/integration/selected_parent_chain_test.go
```diff
@@ -15,7 +16,7 @@ func TestVirtualSelectedParentChain(t *testing.T) {
 
 	// Register to virtual selected parent chain changes
 	onVirtualSelectedParentChainChangedChan := make(chan *appmessage.VirtualSelectedParentChainChangedNotificationMessage)
-	err := kaspad1.rpcClient.RegisterForVirtualSelectedParentChainChangedNotifications(
```
someone235, line 18: nit: Edit this integration test to check if it works propertly

someone235 (changes requested):

testing/integration/selected_parent_chain_test.go
svarogg: Hmmmm... I looked into that, it would be quite complex to modify this test to include checking acceptedTransactionIDs because this tests checks something completely different.

So I did the only responsible thing possible, and created an issue: https://github.com/kaspanet/kaspad/issues/2040

someone235 (changes requested):

someone235 (approved):

Issue #2035, opened 2022-04-22
Crashing kaspads with invalid kaspactl arguments - add testing for all RPC calls.


D-Stacks:
I noticed it is possible to crash kaspads by querying the kaspad with invalid arguments, in my example i queried `./kaspaclt GetBalanceByAddress <address>` using an address that doesn't contain any prefix i.e. ('kaspa:', 'kaspatest'). I suspect the underlying problem is that execeptions, due to bad rpc input, are being raised in the kaspad, while processing the rpc call. These should be error handled properly, and sent back as an error over the rpc response. I suspect this might be the case for other calls also. I propose that Full testing for all rpc calls, could negate these kind of bugs in the future, and would ensure such bugs do not make it into kaspad in the future. 

Below is the full stack trace:

 ```
 2022-04-22 11:59:16.229 [INF] TXMP: RPC Incoming connection from 127.0.0.1:40580
2022-04-22 11:59:16.234 [CRT] RPCS: Exiting: Fatal error in goroutine `routerInitializer-handleIncomingMessages 265`: Couldn't decode address 'qrddu0lv6zwxk50rajelqxgl5jr63d6uwnrv4cfqxaz2as7z09udg75f739nj': decoded address is of unknown format: invalid index of ':'
2022-04-22 11:59:16.234 [CRT] RPCS: Goroutine stack trace: goroutine 429 [running]:
runtime/debug.Stack()
	/usr/lib/go-1.18/src/runtime/debug/stack.go:24 +0x65
github.com/kaspanet/kaspad/util/panics.GoroutineWrapperFunc.func1({0xc2abd3, 0x28}, 0xc0220461e0)
	/home/user/Coding/kaspad/kaspad/util/panics/panics.go:30 +0x35
github.com/kaspanet/kaspad/app/rpc.(*Manager).routerInitializer(0xc01a5b4600, 0xc022046000, 0xc0220460f0)
	/home/user/Coding/kaspad/kaspad/app/rpc/rpc.go:65 +0x222
github.com/kaspanet/kaspad/infrastructure/network/netadapter.newNetConnection({0xd2db00?, 0xc014f58150}, 0xc01a5a3570, {0xc1449e?, 0x60?})
	/home/user/Coding/kaspad/kaspad/infrastructure/network/netadapter/netconnection.go:41 +0xf2
[...]

PR #2037, opened 2022-04-27
Quick fix: RPC clients crashing kaspads

D-Stacks:
This is currently just a preliminary quick fix for  #2035

-  replace `panic(err)` in `Manager.handleError` with a log Warning

I decided to do a PR since I think this quick fix is better then the current situation, although it is still far from ideal.

more info can be gathered from sourcecode comments, but the ideal situation would be returning the err as an rpcerror in the respective request's response.


D-Stacks:
btw I cannot reproduce the failing test on my local branch.

D-Stacks:
update: seems like test just suddenly passed.

D-Stacks:
Will close & wait for proper fix.

PR #2027, opened 2022-04-17
Allow blank address in NotifyUTXOsChanged to get all updates

svarogg:

PR #2044, opened 2022-05-11
Kaspawallet tracking

D-Stacks:
draft for wallet tracking, 

1) follow broadcasted transaction ids and wait for acceptance into the DAG, before reusing utxos.
2) reserving utxos after being used for unsigned transactions.
3) use #2027 for utxo change updates, instead of calling getutxosbyaddresses each secound.

TO DO:

1) find a way to query the orphan pool of a mempool.


D-Stacks:
Closing: kind of made redunant by #2053 but i suppose i will rework this into two different prs. one for notification updates in the wallet, the other for better / more synced tracking, instead of waiting for 1 minute after broadcasting.

PR #2045, opened 2022-05-11
Clarify wallet message concerning a wallet daemon sync state 

Aleoami:
And a couple of memory allocation optimizations

### DIFF
--- a/cmd/kaspawallet/daemon/server/address.go
+++ b/cmd/kaspawallet/daemon/server/address.go
@@ -39,10 +39,10 @@ func (s *server) ShowAddresses(_ context.Context, request *pb.ShowAddressesReque
 	defer s.lock.Unlock()
 
 	if !s.isSynced() {
-		return nil, errors.New("server is not synced")
+		return nil, errors.Errorf("wallet daemon is not synced yet, %s", s.formatSyncStateReport())
 	}
 
-	addresses := make([]string, 0)
+	addresses := make([]string, 0, s.keysFile.LastUsedExternalIndex())
 	for i := uint32(1); i <= s.keysFile.LastUsedExternalIndex(); i++ {
 		walletAddr := &walletAddress{
 			index:         i,
@@ -65,7 +65,7 @@ func (s *server) NewAddress(_ context.Context, request *pb.NewAddressRequest) (*
 	defer s.lock.Unlock()
 
 	if !s.isSynced() {
-		return nil, errors.New("server is not synced")
+		return nil, errors.Errorf("wallet daemon is not synced yet, %s", s.formatSyncStateReport())
 	}
 
 	err := s.keysFile.SetLastUsedExternalIndex(s.keysFile.LastUsedExternalIndex() + 1)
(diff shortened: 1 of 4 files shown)


cmd/kaspawallet/daemon/server/address.go
```diff
@@ -39,10 +39,10 @@ func (s *server) ShowAddresses(_ context.Context, request *pb.ShowAddressesReque
 	defer s.lock.Unlock()
 
 	if !s.isSynced() {
-		return nil, errors.New("server is not synced")
+		return nil, errors.Errorf("wallet daemon is not synced yet, %s", s.formatSyncStateReport())
 	}
 
-	addresses := make([]string, 0)
+	addresses := make([]string, 0, s.keysFile.LastUsedExternalIndex())
```
svarogg, line 45: Please use `make([]string, s.keysFile.LastUsedExternalIndex())`
Than in line 57 instead of append use `addresses[i] = address.String()`.

It's not very different, but a little safer.

cmd/kaspawallet/daemon/server/create_unsigned_transaction.go
```diff
@@ -89,7 +89,7 @@ func (s *server) createUnsignedTransactions(address string, amount uint64, fromA
 func (s *server) selectUTXOs(spendAmount uint64, feePerInput uint64, fromAddresses []*walletAddress) (
 	selectedUTXOs []*libkaspawallet.UTXO, changeSompi uint64, err error) {
 
-	selectedUTXOs = []*libkaspawallet.UTXO{}
+	selectedUTXOs = make([]*libkaspawallet.UTXO, 0, len(s.utxosSortedByAmount))
```
svarogg, line 92: Please undo this. 
Usually selectedUTXOs is not going to include all of s.utxosSortedByAmount, often just a tiny fraction, so you are over-allocating here.

Since it's not known ahead-of-time how much to allocate, better let append do it's thing.

cmd/kaspawallet/daemon/server/external_spendable_utxos.go
```diff
@@ -40,8 +39,9 @@ func (s *server) selectExternalSpendableUTXOs(externalUTXOs *appmessage.GetUTXOs
 	daaScore := dagInfo.VirtualDAAScore
 	maturity := s.params.BlockCoinbaseMaturity
 
-	//we do not make because we do not know size, because of unspendable utxos
-	var selectedExternalUtxos []*pb.UtxosByAddressesEntry
+	// even though we do not know size because of unspendable utxos,
+	// we still can make an assumption about the maximum possible UTXOs
+	selectedExternalUtxos := make([]*pb.UtxosByAddressesEntry, 0, len(externalUTXOs.Entries))
```
svarogg, line 44: Same here.

cmd/kaspawallet/daemon/server/sync.go
```diff
@@ -231,5 +236,16 @@ func (s *server) refreshUTXOs() error {
 }
 
 func (s *server) isSynced() bool {
-	return s.nextSyncStartIndex > s.keysFile.LastUsedInternalIndex() && s.nextSyncStartIndex > s.keysFile.LastUsedExternalIndex()
+	return s.nextSyncStartIndex > s.maxUsedIndex()
+}
+
+func (s *server) formatSyncStateReport() string {
+	maxUsed := s.maxUsedIndex()
```
svarogg, line 243: Please rename to maxUsedIndex

svarogg (changes requested):

cmd/kaspawallet/daemon/server/address.go
Aleoami: It's rather addresses[i-1] = address.String() because this cycle is 1-based

Aleoami (commented):
Remarks fixed

svarogg (dismissed):

someone235 (approved):

PR #2050, opened 2022-05-14
RPC: filter utxos not consensusState - perhaps fixes the orphan problem. 

D-Stacks:
Check to see if utxos supplied by the utxoindex is also found in the consensusState (which is used by the mempool) before sending rpc responses pertaining to the utxos.

This is a hacky PR that might solve the orphan problem (I still need testnet access to acually test it), if this were to fix the orphan problem, it none-the-less would not fix the underlying issue of utxoindex and consensusState not being in sync about utxos. 

Also the code might require a clean-up, and it fails some tests, for reasons i do not know.

### DIFF
--- a/app/rpc/rpccontext/context.go
+++ b/app/rpc/rpccontext/context.go
@@ -44,7 +44,7 @@ func NewContext(cfg *config.Config,
 		UTXOIndex:         utxoIndex,
 		ShutDownChan:      shutDownChan,
 	}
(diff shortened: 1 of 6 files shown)


D-Stacks:
Closing: most likly made redunant by #2052

D-Stacks:
Closing: most likly made redunant by https://github.com/kaspanet/kaspad/pull/2052

PR #2052, opened 2022-05-18
Use a channel for utxo change events 

michaelsutton:
Use a channel from within consensus in order to raise virtual change events in order.

### DIFF
--- a/app/protocol/flowcontext/orphans.go
+++ b/app/protocol/flowcontext/orphans.go
@@ -15,12 +15,6 @@ import (
 // on: 2^orphanResolutionRange * PHANTOM K.
 const maxOrphans = 600
 
-// UnorphaningResult is the result of unorphaning a block
-type UnorphaningResult struct {
-	block            *externalapi.DomainBlock
-	virtualChangeSet *externalapi.VirtualChangeSet
-}
-
 // AddOrphan adds the block to the orphan set
 func (f *FlowContext) AddOrphan(orphanBlock *externalapi.DomainBlock) {
 	f.orphansMutex.Lock()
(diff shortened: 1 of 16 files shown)


someone235 (dismissed):

app/protocol/flowcontext/orphans.go
```diff
@@ -66,7 +60,7 @@ func (f *FlowContext) UnorphanBlocks(rootBlock *externalapi.DomainBlock) ([]*Uno
 	rootBlockHash := consensushashing.BlockHash(rootBlock)
 	processQueue := f.addChildOrphansToProcessQueue(rootBlockHash, []externalapi.DomainHash{})
 
-	var unorphaningResults []*UnorphaningResult
+	var unorphaningResults []*externalapi.DomainBlock
```
someone235, line 63: Can you just call it unorphanedBlocks?

domain/consensus/consensus.go
```diff
@@ -196,16 +196,42 @@ func (s *consensus) ValidateAndInsertBlock(block *externalapi.DomainBlock, shoul
...
+		return nil, err
+	}
+
 	return virtualChangeSet, nil
 }
 
-func (s *consensus) onVirtualChangeSet(virtualChangeSet *externalapi.VirtualChangeSet) {
-	if s.virtualChangeChan == nil {
-		return
+func (s *consensus) onVirtualChanged(virtualChangeSet *externalapi.VirtualChangeSet, wasVirtualUpdated bool) error {
```
someone235, line 208: `onVirtualChange` is a better name, no?

someone235 (changes requested):

someone235 (approved):

PR #2062, opened 2022-05-24
Add wallet daemon state messages to a terminal window log

Aleoami:
Aimed to clarify to the user the state and operational progress of the wallet daemon as it starts: shows launch process stages and UTXO set gathering progress, updating the info as the new addresses appear in the derived chain

### DIFF
--- a/cmd/kaspawallet/daemon/server/sync.go
+++ b/cmd/kaspawallet/daemon/server/sync.go
@@ -14,12 +14,6 @@ var keyChains = []uint8{libkaspawallet.ExternalKeychain, libkaspawallet.Internal
 
 type walletAddressSet map[string]*walletAddress
 
-var (
-	isProgressLogFinalLineShown bool   = false
-	maxProcessedAddressesForLog uint32 = 0
-	maxUsedAddressesForLog      uint32 = 0
-)
-
 func (was walletAddressSet) strings() []string {
 	addresses := make([]string, 0, len(was))
 	for addr := range was {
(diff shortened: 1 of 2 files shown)


cmd/kaspawallet/daemon/server/sync.go
```diff
@@ -274,3 +283,26 @@ func (s *server) refreshUTXOs() error {
 func (s *server) isSynced() bool {
 	return s.nextSyncStartIndex > s.keysFile.LastUsedInternalIndex() && s.nextSyncStartIndex > s.keysFile.LastUsedExternalIndex()
 }
+
+func (s *server) updateSyncingProgressLog(currProcessedAddresses, currMaxUsedAddresses uint32) {
+	if currMaxUsedAddresses > maxUsedAddressesForLog {
+		maxUsedAddressesForLog = currMaxUsedAddresses
```
someone235, line 289: Please don't use global variables for state, instead, store the relevant data inside `server`. This makes testing easier.

someone235 (changes requested):

cmd/kaspawallet/daemon/server/sync.go
```diff
@@ -274,3 +277,30 @@ func (s *server) refreshUTXOs() error {
...
+
+	if s.maxProcessedAddressesForLog >= s.maxUsedAddressesForLog {
+		if !s.isLogFinalProgressLineShown {
+			log.Infof("Wallet is synced, ready for queries")
+			s.isLogFinalProgressLineShown = true
+		}
+	} else {
+		percentProcessed := float64(s.maxProcessedAddressesForLog) / float64(s.maxUsedAddressesForLog) * 100.0
+
+		log.Infof("Gathering UTXOs set, %d addresses of %d processed (%.2f%%)...",
```
someone235, line 303: This is misleading, because the caller function doesn't really gather UTXOs

someone235 (changes requested):

PR #2070, opened 2022-05-30
[Ready] - RPC & UtxoIndex: keep track of, query and test circulating supply.

D-Stacks:
1) Adds, and keeps track of circulating supply to UtxoIndex.Store via utxo index commits.
2) Allows for querying of the circulating supply via a new rpc call `GetCoinSupply`
      -> also shows max supply via maxSompi constant.
3) Adds tests for circulating supply vs both number of mining rounds in loop and vs blockCount RPC call
      -> I just don't fully understand why circulating supply indicates 2 less rewards then block count, one is the genesis block, the other virtual?


D-Stacks:
Practical testing is giving me some wierd results, might have to go back debugging:

```
{
    "getCoinSupplyResponse": {
        "totalSompi": "2900000000000000000",
        "circulatingSompi": "32287296062282",
        "error": null
    }
}
```

domain/utxoindex/utxoindex.go
```diff
@@ -56,6 +62,11 @@ func (ui *UTXOIndex) Reset() error {
 		return err
 	}
 
+	err = ui.store.intalizeCirculatingSompiSupply() //calculate sompi supply from database before adding virtuals
```
someone235, line 65: The DB here is always empty. You can change the comment to "calling intalizeCirculatingSompiSupply so the `circulatingSupplyKey` key will be set"

infrastructure/network/netadapter/server/grpcserver/protowire/rpc.md
```diff
@@ -1817,6 +1819,33 @@ See NotifyNewBlockTemplateRequestMessage
...
+
+<a name="protowire.GetCoinSupplyResponseMessage"></a>
+
+### GetCoinSupplyResponseMessage
+
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| totalSompi | [uint64](#uint64) |  | note: this is a hard coded totalSupply, actual totalSupply is expected to deviate by upto -5%, but cannot be measured exactly. |
```
someone235, line 1841: Why can't it be measured exactly?

someone235 (changes requested):

app/rpc/rpchandlers/get_coin_supply.go
```diff
@@ -0,0 +1,29 @@
...
+		return errorMessage, nil
+	}
+
+	circulatingSompiSupply, err := context.UTXOIndex.GetCirculatingSompiSupply()
+	if err != nil {
+		return nil, err
+	}
+
+	response := appmessage.NewGetCoinSupplyResponseMessage(
+		constants.MaxSompi,
```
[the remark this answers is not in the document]
D-Stacks: I mentioned this as a comment in rpc.md, i think that would suffice? alternative would be an estimation function, and we do not / cannot know our max supply.

infrastructure/network/netadapter/server/grpcserver/protowire/rpc.md
D-Stacks: you even mention this on kaspanet.org i.e. a few hundred million kas will always be +/- . this is due to daa window and deflationary daa scores, i think.

domain/utxoindex/utxoindex.go
D-Stacks: intalizeCirculatingSompiSupply actually goes through the database and adds together all amounts found in it (also adds the key). does this even have any use? i see we delete the whole db a few lines up and fill it with virtuals afterwards. so wondering why i did this.

that being said if looping through an empty database is no overhead i can just leave it I suppose.

will change the comment

domain/utxoindex/store.go
```diff
@@ -313,3 +342,91 @@ func (uis *utxoIndexStore) deleteAll() error {
...
 }
+
+func (uis *utxoIndexStore) intalizeCirculatingSompiSupply() error {
+
+	cursor, err := uis.database.Cursor(utxoIndexBucket)
+	if err != nil {
+		return err
+	}
+	circulatingSompiSupplyInDatabase := uint64(0)
+	defer cursor.Close()
```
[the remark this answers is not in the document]
D-Stacks: will move it up.

domain/utxoindex/store.go
```diff
@@ -313,3 +342,91 @@ func (uis *utxoIndexStore) deleteAll() error {
 
 	return nil
 }
+
+func (uis *utxoIndexStore) intalizeCirculatingSompiSupply() error {
```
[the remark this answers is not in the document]
D-Stacks: will fix the typo

app/appmessage/rpc_get_coin_supply.go
```diff
@@ -34,7 +34,7 @@ func (msg *GetCoinSupplyResponseMessage) Command() MessageCommand {
 // NewGetCoinSupplyResponseMessage returns a instance of the message
 func NewGetCoinSupplyResponseMessage(totalSompi uint64, circulatingSompi uint64) *GetCoinSupplyResponseMessage {
```
someone235: Please call it maxSompi in here as well

someone235 (changes requested):

app/appmessage/rpc_get_coin_supply.go
D-Stacks: i cannot find this one in my code. i think it was changed by commit #72dcf87 but you reveiwed #ab1e7dc

someone235 (approved):

PR #2074, opened 2022-06-01
Unite multiple `GetBlockAcceptanceData` consensus calls to one

michaelsutton:
It is preferred to reduce the number of times consensus lock is acquired during consensus event processing. This should create a better balance between the consensus-event-channel producer (consensus code triggered by inserting blocks and resolving virtual), and the channel consumer (event processing loop). 

My tests show that this change reduces the max channel length from ~2000 to ~500 (during a full IBD and with `includeAcceptedTransactionIDs` listeners)

Note that number of blocks which will be passed to `GetBlocksAcceptanceData` is implicitly limited by the max number of blocks resolved by virtual in a single call (currently 100).


svarogg (approved):
Lgtm

PR #2079, opened 2022-06-07
Add ids to wallet server interactions.

D-Stacks:
I want to use this as a basis for future wallet imporvements for wallet utxo tracking /management as well as asnychronious usage. 

for example can be used to try to release utxos from a tracked state when wallet client commands err, and to track utxos as soon as they are used to create an unsigned transaction, until they are broadcasted.


D-Stacks:
repurposing this for a full implementation of a better utxo tracker, utilizing ids / uids.

PR #2084, opened 2022-06-13
Fix UTXO diff child error 

michaelsutton:
This PR fixes a bug which randomly occurred and indicated a corruption in virtual's state

### DIFF
--- a/domain/consensus/processes/pruningmanager/pruningmanager.go
+++ b/domain/consensus/processes/pruningmanager/pruningmanager.go
@@ -191,6 +191,46 @@ func (pm *pruningManager) UpdatePruningPointByVirtual(stagingArea *model.Staging
 	return nil
 }
 
+type blockIteratorFromOneBlock struct {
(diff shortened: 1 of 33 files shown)


domain/consensus/processes/pruningmanager/pruningmanager.go
```diff
@@ -222,12 +222,16 @@ func (pm *pruningManager) nextPruningPointAndCandidateByBlockHash(stagingArea *m
 		}
 	}
 
-	ghostdagData, err := pm.ghostdagDataStore.Get(pm.databaseContext, stagingArea, blockHash, false)
+	currentPruningPoint, err := pm.pruningStore.PruningPoint(pm.databaseContext, stagingArea)
```
someone235, line 225: Please use the same code that was used in https://github.com/kaspanet/kaspad/pull/2082

someone235 (approved):

PR #2087, opened 2022-06-16
Add explanation about max mass violation

Aleoami:
In order to have less questions to a support, including KDX's one since KDX seems taking that line directly from the node's response.


Aleoami:
Got it

PR #2089, opened 2022-06-17
Rename last references to blockheight - closes #1036

### LINKED ISSUE #1036: Go over remaining places that mention height - and change them to blue score or whatever fits
Block height doesn't make sense in a DAG. we should use the right terminology where it fits.

D-Stacks:
closes #1036
 
 Perhaps `confirmations` is more adequet? since bluescore is only partially synonymous with blockheigt?


domain/consensus/utils/txscript/data/tx_invalid.json
```diff
@@ -272,7 +272,7 @@
     "P2SH"
   ],
   [
-    "Argument negative with by-blockheight nLockTime=0"
+    "Argument negative with by-bluescore nLockTime=0"
```
someone235, line 275: It should be DAA score

domain/consensus/utils/txscript/data/tx_invalid.json
```diff
@@ -564,7 +564,7 @@
     "P2SH"
   ],
   [
-    "Argument negative with by-blockheight txin.nSequence=0"
+    "Argument negative with by-bluescore txin.nSequence=0"
```
someone235, line 567: It should be DAA score

someone235 (changes requested):

someone235 (approved):

PR #2095, opened 2022-06-19
RPC: add connection infos and uptime to getInfo

### LINKED ISSUE #2012: Add the "uptime" parameter to a GetInfo request
Or even create a separate GetStats request, answer to which could be populated with all kinds of info: like a number of p2p and RPC clients, a number of RPC requests responded total, TX requests and block template requests specifically, TXes included into submitted blocks, avg and max RPC request per second, blocks accepted via relay and via submitBlock, and whatever else could be interesting for the node runner.

D-Stacks:
this closes #2012, at least from what is written in the title, perhaps a `getStats` is required in addition.


D-Stacks:
Repurposing this to partition connection infos into a seperate call, since it does bloat this call.

PR #2096, opened 2022-06-19
Add RPC timeout parameter to wallet daemon

Aleoami:
PR provided as a response to the observed RPC call timeout panic of kaspawallet daemon that happened with a testnet wallet having a huge amount of UTXOs. 

Also address decoding operation in unsigned transaction creation procedure is moved several lines higher to increase the responsiveness of the wallet (so to precede the UTXO refreshment operation that takes a long time under the same testnet wallet condition, yet is a complete waste of time if the address is incorrect).

Certain tidying of the code, performed automatically by a VisualStudio Code.

### DIFF
--- a/cmd/kaspaminer/client.go
+++ b/cmd/kaspaminer/client.go
@@ -1,11 +1,12 @@
 package main
 
 import (
+	"time"
+
 	"github.com/kaspanet/kaspad/app/appmessage"
 	"github.com/kaspanet/kaspad/infrastructure/logger"
 	"github.com/kaspanet/kaspad/infrastructure/network/rpcclient"
 	"github.com/pkg/errors"
-	"time"
 )
 
 const minerTimeout = 10 * time.Second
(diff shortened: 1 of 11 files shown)


D-Stacks:
Just a note: this would require a rewrite of all clients using the native kaspad rpcClient to manually supply a timeout, at the very least this would be kgi and katnip.

Aleoami:
I don't think there's too many non-core clients so far, so it won't be a big deal, but on the other hand you're probably right about creating a CustomTimeout()-ed client. This way when the tout parameter is omitted the default client is created, otherwise the custom one. I don't really see the ChangeTimeout() call is useful (at least currently) because I can only know about the timeout violation when the walled daemon has crashed and thus there's nothing to change timeout to anymore :)

Aleoami:
Ok, I'm dumb :D there's already a SetTimeout() call available that should simply be utilized

Aleoami:
I've made a mess with Git, have to restart it all

D-Stacks:
I know that feeling all too well :)

PR #2107, opened 2022-06-26
Kaspawallet: implement fee calcs, centralize constant and logic

D-Stacks:
1) Move fee calcs, logic and constants to libkaspawallet
2) Allow for functions to calculate transaction(s) fees


D-Stacks:
Makes more sense to optionally return the whole tx, this can be used by clients to send, and get tx data for post-analysis, no reason to limit this feature in the `Send` call to just return fees.

D-Stacks:
I will split this into two dependent prs, one for feecalcs, one for sending back whole tx data, seems to make the most sense.

PR #2111, opened 2022-07-02
Mempool: Retrive stable state of the mempool. optimze get mempool entries by addresses

D-Stacks:
1) Adds a `clone` parameter to relevent mempool functions.
2) retrive only cloned mempool transactions (no pointers!) through relevent interfaces (i.e. outside of the mempool internal functioning)

3) when accessinging the mempool, query orphan and transactionpool together with one lock.

4) add retrival of mempool transactions by addresses through the interface
5) use point number 3 in order to not double loop transactions in get mempool entries by addresses

6) check if requested transaction is a known orphan before requesting. 

7) since both mempool and utxoindex now need accesses to a stringification and destringification of SciptPublicKeys for map building, I decided to add these as methods to externalapi.ScriptPublicKey

- `1)` and `2)`  ensures that accessing the mempool does not incur multiple threads accessing the same memory, it also ensures that retrived transactionpool transactions must be in a stable state with populated utxos.

- `3)` side-steps unintuitive mempool state as per rpc responses, where a transaction could be both an orphan and not an orphan, due to state changes between transactionpool and orphanpool queries. 

- `4)` and `5`) is an optimization and rewrite of the GetMempoolEntriesByAddresses RPC command. it ensures mempool transactions are only looped once instead of twice. 

- `6`) is a flow improvement for transaction requesting. i.e. it filters transactions that are already known transactions in the orphan pool.

- `7`) is for code maintenance.  

it should fix mempool syncing and missing utxos in retrived mempool transactions and perhaps a whole range of other mempool related issues, which may be caused by this issue, or flows which access mempool data.

### DIFF
--- a/app/protocol/flows/v5/transactionrelay/handle_requested_transactions.go
+++ b/app/protocol/flows/v5/transactionrelay/handle_requested_transactions.go
@@ -30,10 +30,6 @@ func (flow *handleRequestedTransactionsFlow) start() error {
 		}
 
 		for _, transactionID := range msgRequestTransactions.IDs {
-			//note: below ignores orphan txs that are requested
-			//find out if this is good or bad practice
-			//only reference i found to this, is that nodes don't do this in btc
-			//source: https://arxiv.org/abs/1912.11541 (2nd sentence in abstract)
 			tx, _, ok := flow.Domain().MiningManager().GetTransaction(transactionID, true, false)
 
 			if !ok {
--- a/app/protocol/flows/v5/transactionrelay/handle_requested_transactions.go
+++ b/app/protocol/flows/v5/transactionrelay/handle_requested_transactions.go
@@ -30,7 +30,11 @@ func (flow *handleRequestedTransactionsFlow) start() error {
 		}
 
 		for _, transactionID := range msgRequestTransactions.IDs {
-			tx, ok := flow.Domain().MiningManager().GetTransaction(transactionID)
+			//note: below ignores orphan txs that are requested
+			//find out if this is good or bad practice
+			//only reference i found to this, is that nodes don't do this in btc
+			//source: https://arxiv.org/abs/1912.11541 (2nd sentence in abstract)
+			tx, _, ok := flow.Domain().MiningManager().GetTransaction(transactionID, true, false)
 
 			if !ok {
 				msgTransactionNotFound := appmessage.NewMsgTransactionNotFound(transactionID)
@@ -40,7 +44,6 @@ func (flow *handleRequestedTransactionsFlow) start() error {
 				}
 				continue
 			}
-
 			err := flow.outgoingRoute.Enqueue(appmessage.DomainTransactionToMsgTx(tx))
 			if err != nil {
 				return err
--- a/domain/miningmanager/mempool/handle_new_block_transactions.go
+++ b/domain/miningmanager/mempool/handle_new_block_transactions.go
@@ -6,7 +6,7 @@ import (
 	"github.com/kaspanet/kaspad/domain/consensus/utils/transactionhelper"
 )
 
-func (mp *mempool) handleNewBlockTransactions(blockTransactions []*externalapi.DomainTransaction, clone bool) (
+func (mp *mempool) handleNewBlockTransactions(blockTransactions []*externalapi.DomainTransaction,) (
 	[]*externalapi.DomainTransaction, error) {
 
 	// Skip the coinbase transaction
@@ -30,7 +30,7 @@ func (mp *mempool) handleNewBlockTransactions(blockTransactions []*externalapi.D
 			return nil, err
 		}
 
-		acceptedOrphansFromThisTransaction, err := mp.orphansPool.processOrphansAfterAcceptedTransaction(transaction, clone)
+		acceptedOrphansFromThisTransaction, err := mp.orphansPool.processOrphansAfterAcceptedTransaction(transaction)
 		if err != nil {
 			return nil, err
 		}
--- a/domain/miningmanager/mempool/handle_new_block_transactions.go
+++ b/domain/miningmanager/mempool/handle_new_block_transactions.go
@@ -6,7 +6,7 @@ import (
 	"github.com/kaspanet/kaspad/domain/consensus/utils/transactionhelper"
 )
 
-func (mp *mempool) handleNewBlockTransactions(blockTransactions []*externalapi.DomainTransaction) (
+func (mp *mempool) handleNewBlockTransactions(blockTransactions []*externalapi.DomainTransaction, clone bool) (
 	[]*externalapi.DomainTransaction, error) {
 
 	// Skip the coinbase transaction
@@ -30,7 +30,7 @@ func (mp *mempool) handleNewBlockTransactions(blockTransactions []*externalapi.D
 			return nil, err
 		}
 
-		acceptedOrphansFromThisTransaction, err := mp.orphansPool.processOrphansAfterAcceptedTransaction(transaction)
+		acceptedOrphansFromThisTransaction, err := mp.orphansPool.processOrphansAfterAcceptedTransaction(transaction, clone)
 		if err != nil {
 			return nil, err
 		}
--- a/domain/miningmanager/mempool/handle_new_block_transactions.go
+++ b/domain/miningmanager/mempool/handle_new_block_transactions.go
@@ -6,7 +6,7 @@ import (
 	"github.com/kaspanet/kaspad/domain/consensus/utils/transactionhelper"
 )
 
-func (mp *mempool) handleNewBlockTransactions(blockTransactions []*externalapi.DomainTransaction,) (
+func (mp *mempool) handleNewBlockTransactions(blockTransactions []*externalapi.DomainTransaction) (
 	[]*externalapi.DomainTransaction, error) {
 
 	// Skip the coinbase transaction
--- a/domain/miningmanager/mempool/mempool.go
+++ b/domain/miningmanager/mempool/mempool.go
@@ -64,7 +64,7 @@ func (mp *mempool) GetTransaction(transactionID *externalapi.DomainTransactionID
 		isOrphan = true
 	}
 
-	return transaction, isOrphan, transactionfound 
+	return transaction, isOrphan, transactionfound
 }
 
 func (mp *mempool) GetTransactionsByAddresses(includeTransactionPool bool, includeOrphanPool bool) (
--- a/domain/miningmanager/mempool/mempool.go
+++ b/domain/miningmanager/mempool/mempool.go
@@ -4,7 +4,6 @@ import (
 	"sync"
 
 	"github.com/kaspanet/kaspad/domain/consensusreference"
-	"github.com/kaspanet/kaspad/domain/dagconfig"
 
 	"github.com/kaspanet/kaspad/domain/consensus/model/externalapi"
 	miningmanagermodel "github.com/kaspanet/kaspad/domain/miningmanager/model"
@@ -14,7 +13,6 @@ type mempool struct {
 	mtx sync.RWMutex
 
 	config             *Config
-	params             *dagconfig.Params
 	consensusReference consensusreference.ConsensusReference
 
 	mempoolUTXOSet   *mempoolUTXOSet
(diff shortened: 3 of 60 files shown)


domain/miningmanager/mempool/revalidate_high_priority_transactions.go
```diff
@@ -6,22 +6,35 @@ import (
...
+			}
 
-	for _, transaction := range mp.transactionsPool.highPriorityTransactions {
-		isValid, err := mp.revalidateTransaction(transaction)
-		if err != nil {
-			return nil, err
-		}
-		if !isValid {
-			continue
+			validTransactions = append(validTransactions, transaction.Transaction())
```
someone235, line 24: I think it's better to move the `clone` if-else here

domain/miningmanager/mempool/transactions_pool.go
```diff
@@ -204,17 +209,52 @@ func (tp *transactionsPool) limitTransactionCount() error {
...
 }
 
-func (tp *transactionsPool) getAllTransactions() []*externalapi.DomainTransaction {
-	allTransactions := make([]*externalapi.DomainTransaction, 0, len(tp.allTransactions))
+func (tp *transactionsPool) getTransactionsByAddresses(clone bool) (
+	sending model.ScriptPublicKeyStringToDomainTransaction,
+	receiving model.ScriptPublicKeyStringToDomainTransaction,
+	err error) {
+	sending = make(model.ScriptPublicKeyStringToDomainTransaction)
+	receiving = make(model.ScriptPublicKeyStringToDomainTransaction)
```
someone235, line 227: You could preallocate both of the to the size of `tp.allTransactions`

domain/miningmanager/mempool/transactions_pool.go
```diff
@@ -204,17 +209,52 @@ func (tp *transactionsPool) limitTransactionCount() error {
...
+	var transaction *externalapi.DomainTransaction
 	for _, mempoolTransaction := range tp.allTransactions {
-		allTransactions = append(allTransactions, mempoolTransaction.Transaction())
+		if clone {
+			transaction = mempoolTransaction.Transaction().Clone()
+		} else {
+			transaction = mempoolTransaction.Transaction()
+		}
+		for _, input := range transaction.Inputs {
+			if input.UTXOEntry == nil { //this should be fixed
```
someone235, line 236: Does it still happen?

domain/miningmanager/miningmanager_test.go
```diff
@@ -225,8 +225,10 @@ func TestHandleNewBlockTransactions(t *testing.T) {
 		if err != nil {
 			t.Fatalf("HandleNewBlockTransactions: %v", err)
 		}
-		if len(miningManager.AllTransactions()) != 0 {
-			blockIDs := domainBlocksToBlockIds(miningManager.AllTransactions())
+		mempoolTransactions, _ = miningManager.AllTransactions(true, false)
+		if len(mempoolTransactions) != 0 {
+			mempoolTransactions, _ = miningManager.AllTransactions(true, false)
```
someone235, line 230: Why have you added this line?

domain/miningmanager/miningmanager_test.go
```diff
@@ -605,7 +608,7 @@ func TestModifyBlockTemplate(t *testing.T) {
 				t.Fatalf("ValidateAndInsertTransaction: %v", err)
 			}
 		}
-		transactionsMempool := miningManager.AllTransactions()
+		transactionsMempool, _ := miningManager.AllTransactions(true, true)
```
someone235, line 611: Why have you changed it to include orphans?

domain/miningmanager/miningmanager_test.go
```diff
@@ -654,7 +657,7 @@ func TestModifyBlockTemplate(t *testing.T) {
 		if err != nil {
 			t.Fatalf("HandleNewBlockTransactions: %+v", err)
 		}
-		transactionsMempool = miningManager.AllTransactions()
+		transactionsMempool, _ = miningManager.AllTransactions(true, true)
```
someone235, line 660: Why have you changed it to include orphans?

testing/integration/tx_relay_test.go
```diff
@@ -103,6 +103,7 @@ func TestTxRelay(t *testing.T) {
 						t.Fatal("Error payer is reciving")
 					}
 				}
+
```
someone235, line 106: Please rollback this file (I know it's not important, but I don't want to spam the `git blame` log)

someone235 (changes requested):
As discussed in Discord, I'm still not sure the `clone` flag is actually needed. Maybe we should keep it as a temporary flag that is turned off for now.

app/protocol/flows/v5/transactionrelay/handle_requested_transactions.go
```diff
@@ -30,7 +30,11 @@ func (flow *handleRequestedTransactionsFlow) start() error {
 		}
 
 		for _, transactionID := range msgRequestTransactions.IDs {
-			tx, ok := flow.Domain().MiningManager().GetTransaction(transactionID)
+			//note: below ignores orphan txs that are requested
```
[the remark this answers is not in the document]
D-Stacks: removed comment

domain/miningmanager/mempool/handle_new_block_transactions.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/kaspanet/kaspad/domain/consensus/utils/transactionhelper"
 )
 
-func (mp *mempool) handleNewBlockTransactions(blockTransactions []*externalapi.DomainTransaction) (
+func (mp *mempool) handleNewBlockTransactions(blockTransactions []*externalapi.DomainTransaction, clone bool) (
```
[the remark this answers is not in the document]
D-Stacks: yes, that is useless, i will add the parameter into the function. I decided to clone everything out of the mempool because it seemed like it would be safer and good practice but my first implementation was with a parameter. But will keep it, just in case.

domain/miningmanager/mempool/orphan_pool.go
```diff
@@ -329,17 +333,58 @@ func (op *orphansPool) randomNonHighPriorityOrphan() *model.OrphanTransaction {
...
 }
 
-func (op *orphansPool) getAllOrphanTransactions() []*externalapi.DomainTransaction {
-	allOrphanTransactions := make([]*externalapi.DomainTransaction, 0, len(op.allOrphans))
+func (op *orphansPool) getOrphanTransactionsByAddresses(clone bool) (
+	sending model.ScriptPublicKeyStringToDomainTransaction,
+	receiving model.ScriptPublicKeyStringToDomainTransaction,
+	err error) {
+	sending = make(model.ScriptPublicKeyStringToDomainTransaction)
+	receiving = make(model.ScriptPublicKeyStringToDomainTransaction)
```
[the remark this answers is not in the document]
D-Stacks: we can have multiple outputs in one transaction, so `len(op.allOrphans)` would not suffice.

domain/miningmanager/mempool/mempool.go
```diff
@@ -39,42 +43,90 @@ func (mp *mempool) ValidateAndInsertTransaction(transaction *externalapi.DomainT
...
 
-	return mp.validateAndInsertTransaction(transaction, isHighPriority, allowOrphan)
+	return mp.validateAndInsertTransaction(transaction, isHighPriority, allowOrphan, true)
 }
 
-func (mp *mempool) GetTransaction(transactionID *externalapi.DomainTransactionID) (*externalapi.DomainTransaction, bool) {
+func (mp *mempool) GetTransaction(transactionID *externalapi.DomainTransactionID,
+	includeTransactionPool bool,
+	includeOrphanPool bool) (
+	transactionPoolTransaction *externalapi.DomainTransaction,
```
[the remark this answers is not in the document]
D-Stacks: I think it makes navigating these mempool retival calls more consistent, but can change if you want.

domain/miningmanager/mempool/transactions_pool.go
D-Stacks: no, haven't encountered it since cloning.

domain/miningmanager/mempool/transactions_pool.go
D-Stacks: can't one transaction can have multiple inputs / outputs from different addresses? was thinking: number of transactions doesn't have to be number of reciving or sending addresses, it can be more in the latter?

domain/miningmanager/mempool/handle_new_block_transactions.go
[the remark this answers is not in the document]
D-Stacks: yes, that is useless, i will change to cloning directly, where it is used, and add a comment why it is cloned.

domain/miningmanager/mempool/mempool.go
```diff
@@ -20,9 +23,10 @@ type mempool struct {
 }
 
 // New constructs a new mempool
-func New(config *Config, consensusReference consensusreference.ConsensusReference) miningmanagermodel.Mempool {
+func New(config *Config, params *dagconfig.Params, consensusReference consensusreference.ConsensusReference) miningmanagermodel.Mempool {
 	mp := &mempool{
 		config:             config,
+		params:             params,
```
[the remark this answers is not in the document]
D-Stacks: I can't quiet remember, but i think it is a relic from some previous iteration of the pr, will remove.

domain/miningmanager/mempool/mempool.go
```diff
@@ -39,42 +43,90 @@ func (mp *mempool) ValidateAndInsertTransaction(transaction *externalapi.DomainT
...
+	var orphanTransactionFound bool
 
-func (mp *mempool) AllTransactions() []*externalapi.DomainTransaction {
-	mp.mtx.RLock()
-	defer mp.mtx.RUnlock()
+	if includeTransactionPool {
+		transactionPoolTransaction, transactionfound = mp.transactionsPool.getTransaction(transactionID, true)
+	}
+	if includeOrphanPool {
+		orphanPoolTransaction, orphanTransactionFound = mp.orphansPool.getOrphanTransaction(transactionID, true)
```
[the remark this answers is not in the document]
D-Stacks: I think i did this for testing purposes (before this pr we had get orphan and get transaction as two seperate calls, it ended so that the rpc calls had both orphans and transaction pool transactions), ideally it shouldn't happen under one lock, i only see a slight usage case for testing the mempool. but will remove it, since it has no functional purpose.

testing/integration/tx_relay_test.go
D-Stacks: so I'll revert back without the whitespace.

domain/miningmanager/mempool/revalidate_high_priority_transactions.go
D-Stacks: i will make it clone by default as per other comment.

domain/miningmanager/miningmanager_test.go
D-Stacks: at least this line and the orphan retrival.

domain/miningmanager/mempool/mempool.go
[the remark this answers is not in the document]
D-Stacks: I think it makes navigating these mempool retival calls more consistent, will change though.

D-Stacks:
something failing tests, need to debug.

domain/miningmanager/mempool/orphan_pool.go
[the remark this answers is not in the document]
someone235: > we can have multiple outputs in one transaction, so `len(op.allOrphans)` would not suffice.

It's only an initial allocation. It would save the need to grow the map multiple times.

domain/miningmanager/mempool/transactions_pool.go
someone235: So let's remove the "this should be fixed" etc

domain/miningmanager/mempool/orphan_pool.go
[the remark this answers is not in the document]
D-Stacks: will do

someone235 (approved):

PR #2117, opened 2022-07-16
Set kaspawallet daemon send/recv message size limits to 100MB

michaelsutton:


michaelsutton:
This PR is suppose to allow huge wallet compounding operations to succeed (supporting hundreds of transactions).  However I tested this on testnet with the help of @biospb, and it seems that some of the later transactions do not get processed by the network, possibly due to mempool cleanup policies (it takes the network a few minutes to process hundreds of huge transactions).  It seems that only the first layer of compounding txs succeeds, while the following layers which depend on the first, are dropped eventually.
Any thoughts @someone235 @svarogg before I close this ?

PR #2121, opened 2022-08-01
Fix a subtle lock sync issue in consensus insert block 

michaelsutton:


domain/consensus/consensus.go
```diff
@@ -198,15 +198,33 @@ func (s *consensus) ValidateAndInsertBlock(block *externalapi.DomainBlock, updat
...
 		s.lock.Lock()
 		if s.virtualNotUpdated {
-			s.lock.Unlock()
-			err := s.ResolveVirtual(nil)
-			if err != nil {
-				return err
+			// We enter the loop in locked state
+			for {
+				// See comment about the const used at `ResolveVirtual`
+				_, isCompletelyResolved, err := s.resolveVirtualChunkNoLock(100)
```
svarogg, line 204: Please replace the 100 with constant (here and anywhere else it's used)

svarogg (changes requested):

svarogg (approved):

PR #2158, opened 2022-10-13
Make more checks if status is invalid even if the block exists

someone235:


app/rpc/rpchandlers/get_blocks.go
```diff
@@ -37,7 +37,7 @@ func HandleGetBlocks(context *rpccontext.Context, _ *router.Router, request appm
 			return nil, err
 		}
 
-		if !blockInfo.Exists {
+		if !blockInfo.Exists || blockInfo.BlockStatus == externalapi.StatusInvalid {
```
michaelsutton, line 40: you can also use `!blockInfo.HasHeader()`

michaelsutton (dismissed):

michaelsutton (approved):

PR #2162, opened 2022-10-26
Update version.go

rpringst:
Version Bump.


frenchtoasters (approved):
Also noticed this was not updated with the latest release.

someone235:
It's probably doesn't worth to make a release just for it, so we'll just fix it in v0.12.10

PR #2168, opened 2022-11-15
Update version.go

MatanHamilis:
Since version 0.12.9 was released, shouldn't the version in this file correspond to the alleged version?
Was unsure, but when running my node it was apparent to me that the reported version is inconsistent.


someone235:
Was fixed on #2167

PR #2169, opened 2022-11-19
Compare blue score with selected tip when checking if a pruning point…

someone235:
… proof is needed

### DIFF
--- a/app/protocol/flows/v5/blockrelay/ibd_with_headers_proof.go
+++ b/app/protocol/flows/v5/blockrelay/ibd_with_headers_proof.go
@@ -55,7 +55,12 @@ func (flow *handleIBDFlow) shouldSyncAndShouldDownloadHeadersProof(
 
 	var highestSharedBlockFound, isPruningPointInSharedBlockChain bool
 	if highestKnownSyncerChainHash != nil {
-		highestSharedBlockFound = true
+		blockInfo, err := flow.Domain().Consensus().GetBlockInfo(highestKnownSyncerChainHash)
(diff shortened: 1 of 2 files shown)


app/protocol/flows/v5/blockrelay/ibd_with_headers_proof.go
```diff
@@ -87,21 +92,21 @@ func (flow *handleIBDFlow) shouldSyncAndShouldDownloadHeadersProof(
...
 
 func (flow *handleIBDFlow) checkIfHighHashHasMoreBlueWorkThanSelectedTipAndPruningDepthMoreBlueScore(relayBlock *externalapi.DomainBlock) (bool, error) {
-	headersSelectedTip, err := flow.Domain().Consensus().GetHeadersSelectedTip()
+	virtualSelectedParent, err := flow.Domain().Consensus().GetVirtualSelectedParent()
 	if err != nil {
 		return false, err
 	}
 
-	headersSelectedTipInfo, err := flow.Domain().Consensus().GetBlockInfo(headersSelectedTip)
+	virtualSelectedTipInfo, err := flow.Domain().Consensus().GetBlockInfo(virtualSelectedParent)
```
michaelsutton, line 100: do we need to check if `virtualSelectedTipInfo.Exists` ? it probably by def will always exist + blue work/score checks below will work with zeros as well, so I'm not sure if needed

michaelsutton (approved):

michaelsutton (approved):

PR #2220, opened 2023-09-14
Fix off by small amounts in sent amount kaspa

coderofstuff:
**Problem:** Floating point math causes inconsistencies when converting kas to sompi.

**Solution:** Use a method that parses the amount as a string, the converts it to sompi string then parse to uint64

**Motivation:** Someone (some 3rd party integrating with kaspa) sent 33184.1489732 via kaspawallet and the other party got 33184.14897319 because the kaspawallet send command loses some precision when converting kas to sompi (floating point math kinda stuff), which is off by 0.00000001. Small difference but irreconcilable to accounting.


cmd/kaspawallet/send.go
```diff
@@ -99,3 +99,15 @@ func send(conf *sendConfig) error {
...
 }
+
+/**
+ */
+func kasToSompi(amount float64) uint64 {
+	amountInStr := fmt.Sprintf("%.8f", amount)
+
+	parts := strings.Split(amountInStr, ".")
+
+	convertedAmount, _ := strconv.ParseUint(strings.Join(parts, ""), 10, 64)
```
tmrlvi, line 110: Check the error value (and return error is it is set). Number could be too big, and then ParseUint will return (nil, error)

tmrlvi:
`conf.SendAmount` is accepted as float64, so another source of math inconsistencies would be in the conversions of string -> float64 -> string again. 

Maybe we should consider accepting string value, validating, and working directly with it (fixed precision type would be better, but I am not familiar with such solutions in go)

tmrlvi (changes requested):
`conf.SendAmount` is accepted as float64, so another source of math inconsistencies would be in the conversions of string -> float64 -> string again.

Maybe we should consider accepting string value, validating, and working directly with it (fixed precision type would be better, but I am not familiar with such solutions in go)

cmd/kaspawallet/send.go
coderofstuff: Addressed in [83c3d44](https://github.com/kaspanet/kaspad/pull/2220/commits/83c3d4458eb886ef2272130026d35ece56cbf946)

cmd/kaspawallet/utils/validate_amount.go
```diff
@@ -0,0 +1,34 @@
...
+	if !match {
+		return errors.Errorf("Invalid send amount")
+	}
+
+	if err != nil {
+		return err
+	}
+
+	// If it parses properly, then this is valid
+	_, err = strconv.ParseFloat(amount, 64)
```
tmrlvi, line 27: Isn't this redundant (assuming fixed regex)? (though it doesn't hurt)

tmrlvi (changes requested):

cmd/kaspawallet/utils/validate_amount.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/pkg/errors"
+)
+
+/**
+ * 1. May be an integer (no decimal components)
+ * 2. May be float with up to 8 decimal places
+ */
+func ValidateAmountFormat(amount string) error {
+	// Check whether it's an integer, or a float with max 8 digits
+	match, err := regexp.MatchString("^\\d{1,19}(.\\d{0,8})?$", amount)
```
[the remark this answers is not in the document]
tmrlvi: Also, should it be only 11 digits before the period? (in total 19)

cmd/kaspawallet/send.go
```diff
@@ -35,7 +35,13 @@ func send(conf *sendConfig) error {
...
 	var sendAmountSompi uint64
 	if !conf.IsSendAll {
-		sendAmountSompi = uint64(conf.SendAmount * constants.SompiPerKaspa)
+		parsedAmountSompi, err := utils.KasToSompi(conf.SendAmount)
+
+		if err != nil {
+			return err
+		}
+
+		sendAmountSompi = parsedAmountSompi
```
tmrlvi, line 44: Is there a reason it is not directly `sendAmountSompi, err := utils.KasToSompi(conf.SendAmount)`?

cmd/kaspawallet/config.go
```diff
@@ -296,17 +296,17 @@ func parseCommandLine() (subCommand string, config interface{}) {
 }
 
 func validateCreateUnsignedTransactionConf(conf *createUnsignedTransactionConfig) error {
-	if (!conf.IsSendAll && conf.SendAmount == 0) ||
-		(conf.IsSendAll && conf.SendAmount > 0) {
+	if (!conf.IsSendAll && conf.SendAmount == "") ||
+		(conf.IsSendAll && conf.SendAmount > "") {
```
tmrlvi, line 300: You probably want to change this as well to `!=` (to be consistent with below, but both work)

tmrlvi:
Seems good. I think you should fix the regex to raise parsing error earlier with a clearer description.

The other two comments are not material, just for cleaner code

cmd/kaspawallet/utils/validate_amount.go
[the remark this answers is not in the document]
coderofstuff: The maximum possible sompi for uint64 should be `18446744073709551615` - in float that'd be `184467440737.09551615` (12 digits to the left of the decimal). So the rule `"^\([1-9]d{0,11}|0)(\.\d{0,8})?$"` is enough.

More test cases added.

cmd/kaspawallet/utils/validate_amount.go
coderofstuff: It's redundant now that the regex is fixed

cmd/kaspawallet/send.go
coderofstuff: Yeah, because of shadowing: https://go.dev/play/p/vAX1AHDpmvs

However, we since `err` is already declared above the lines, I can do (and have done):

```
sendAmountSompi, err = utils.KasToSompi(conf.SendAmount)
```

(note that it's `=` and not `:=`)

tmrlvi (commented):
Just change

tmrlvi (approved):

tmrlvi (approved):

tmrlvi (approved):

cmd/kaspawallet/utils/format_kas.go
```diff
@@ -14,3 +16,28 @@ func FormatKas(amount uint64) string {
...
+		return 0, err
+	}
+
+	// after validation, amount can only be either an int OR
+	// a float with an int component and decimal places
+	parts := strings.Split(amount, ".")
+	amountStr := ""
+
+	if len(parts) == 2 {
+		amountStr = fmt.Sprintf("%s%-*s", parts[0], 8, parts[1]) // Padded with spaces at the end to fill for missing decimals: Sample "0.01234    "
```
someone235, line 34: Use `math.Log10(SompiPerKaspa)` to determine amount of zeroes (and also add a test that checks `SompiPerKaspa` is a multiple of 10.

cmd/kaspawallet/utils/format_kas.go
```diff
@@ -14,3 +16,28 @@ func FormatKas(amount uint64) string {
...
+		amountStr = strings.ReplaceAll(amountStr, " ", "0")      // Make the spaces be 0s. Sample "0.012340000"
+	} else {
+		amountStr = fmt.Sprintf("%s00000000", parts[0])
```
someone235, line 37: Use `math.Log10(SompiPerKaspa)` to determine amount of zeroes

cmd/kaspawallet/utils/validate_amount.go
```diff
@@ -0,0 +1,26 @@
...
+	"regexp"
+
+	"github.com/pkg/errors"
+)
+
+/**
+ * 1. May be an integer (no decimal components)
+ * 2. May be float with up to 8 decimal places
+ */
+func ValidateAmountFormat(amount string) error {
```
someone235, line 13: I wouldn't make it public, and rename it to reflect that it talks about KAS amount (maybe `validateKASAmountFormat`)

someone235: I would also move it to `format_kas.go` since it's only relevant for `KasToSompi`

someone235 (changes requested):

cmd/kaspawallet/utils/format_kas.go
```diff
@@ -30,14 +33,36 @@ func KasToSompi(amount string) (uint64, error) {
...
 
 	return convertedAmount, err
 }
+
+func validateAmountFormat(amount string) error {
+	// Check whether it's an integer, or a float with max 8 digits
+	match, err := regexp.MatchString("^([1-9]\\d{0,11}|0)(\\.\\d{0,8})?$", amount)
+
+	if !match {
+		return errors.Errorf("Invalid send amount")
```
someone235, line 60: Change to "invalid amount" (we're not in wallet context)

cmd/kaspawallet/utils/format_kas.go
```diff
@@ -30,14 +33,36 @@ func KasToSompi(amount string) (uint64, error) {
...
 
+	amountStr = fmt.Sprintf("%s%-*s", parts[0], decimalPlaces, decimalStr) // Padded with spaces at the end to fill for missing decimals: Sample "0.01234    "
+	amountStr = strings.ReplaceAll(amountStr, " ", "0")                    // Make the spaces be 0s. Sample "0.012340000"
+
 	convertedAmount, err := strconv.ParseUint(amountStr, 10, 64)
 
 	return convertedAmount, err
 }
+
+func validateAmountFormat(amount string) error {
```
someone235, line 55: change to `validateKASAmountFormat` (this function is accessible from the whole package so its name should be understandable in any context)

someone235 (changes requested):

someone235 (approved):

PR #2223, opened 2023-09-26
Anti-spam measurements against dust attack

someone235:


michaelsutton (approved):

michaelsutton:
Since this patch was widely accepted by the mining community [(anti-spam-patch)](https://github.com/kaspanet/kaspad/releases/tag/anti-spam-patch7) we merge it directly to master and will add a tag v0.12.14 + a formal release in order to effectively make it a well known and consolidated version. 

The patch introduces temporary tx relay rules and a local build block template strategy which avoids mining spam/dust txs. This is not a consensus rule change. 

Note: the patch was not merged to the `dev` branch

PR #2226, opened 2023-10-26
Set gas correctly when converting from rpcTx

coderofstuff:
Previously it was setting it from LockTime


someone235 (approved):

someone235:
Fixed by #2203

PR #2258, opened 2023-12-26
Lazy wallet utxo sync after broadcasting a tx

someone235:


cmd/kaspawallet/daemon/server/create_unsigned_transaction.go
```diff
@@ -113,7 +112,9 @@ func (s *server) selectUTXOs(spendAmount uint64, isSendAll bool, feePerInput uin
 		}
 
 		if broadcastTime, ok := s.usedOutpoints[*utxo.Outpoint]; ok {
-			if time.Since(broadcastTime) > time.Minute {
+			// We want to free an outpoint from the used outpoints set if we refresh the UTXOs since it was
+			// marked as used, and it least one minute has passed.
+			if time.Since(broadcastTime) > time.Minute && s.startTimeOfLastCompletedRefresh.After(broadcastTime) {
```
michaelsutton, line 117: I think it should be `if s.startTimeOfLastCompletedRefresh.After(broadcastTime + time.Minute)`
and no need for first condition bcs  `startTimeOfLastCompletedRefresh` <= now

also fix the comment accordingly

michaelsutton: the comment can be:

```
If the node returned a UTXO we previously attempted to spend and enough time has passed, we assume that the network rejected or lossed the previous transaction and allow a reuse. We set this time interval to a minute. 
We also verify that a full refresh UTXO operation started after this time point and has already completed, in order to make sure that indeed this state reflects a state obtained following the required wait time.
```

michaelsutton (approved):

PR #2275, opened 2024-05-05
Fix block verbosedata population…

TMKCodes:
Fix block verbosedata population from skipping transaction verbosedata population when BlockInfo errorenously reports block as Header only, but domainBlock is still fount with GetBlockEvenIfHeaderOnly. 

Basically this affects archival servers. Once archival server moves on pruning point it loses verboseData from transactions as such transaction id is lost unless repopulated from the block data.


bobpinella:
We are currently running an archive node and need this fix badly - will this PR be incorporated into the next version anytime soon?

app/rpc/rpccontext/verbosedata.go
```diff
@@ -93,6 +89,10 @@ func (ctx *Context) PopulateBlockWithVerboseData(block *appmessage.RPCBlock, dom
 		}
 	}
 
+	if domainBlock == nil && blockInfo.BlockStatus == externalapi.StatusHeaderOnly {
+		return nil
+	}
+
```
someone235, line 95: This condition is always true, since `GetBlockEvenIfHeaderOnly` returns an error if the block is not found. I guess this is the condition you wanted to check?
```suggestion
	if len(domainBlock.Transactions) == 0 {
		return nil
	}

```

someone235 (changes requested):

PR #2282, opened 2024-06-30
Add deprecated message

someone235:


aspect:
I suggest the following wording:

The full node reference implementation was rewritten in Rust, as a result, the Go implementation is now deprecated.

PLEASE NOTE: (or IMPORTANT:) Any pull requests or issues that will be opened in this repository will be closed without treatment, except for issues or pull requests related to the kaspawallet, which remains maintained. In any other case, please use the Rust implementation instead: https://github.com/kaspanet/rusty-kaspa

michaelsutton (approved):

PR #2291, opened 2024-09-04
Add fee estimation to wallet

someone235:


cmd/kaspawallet/daemon/server/bump_fee.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, errors.Errorf("new fee rate (%f) is not higher than the current fee rate (%f)", newFeeRate, feeRate)
+	}
+
+	outpointsSet := make(map[externalapi.DomainOutpoint]struct{})
+	for _, input := range domainTx.Inputs {
+		outpointsSet[input.PreviousOutpoint] = struct{}{}
+	}
+
+	var maxUTXO *walletUTXO
+	for _, utxo := range s.utxosSortedByAmount {
```
michaelsutton, line 45: Afaiu `s.utxosSortedByAmount` will not hold current mempool entries (filtered [here](https://github.com/someone235/kaspad/blob/6901e7d53863ac5b7b49e03f018d256bee9bfd64/cmd/kaspawallet/daemon/server/sync.go#L234)), so it seems this search will most likely fail. 

I think we need to keep the mempool entries (during the refresh; in `s.excludedMempoolUtxos`), and search through them here as well

cmd/kaspawallet/daemon/server/bump_fee.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	domainTx, err := appmessage.RPCTransactionToDomainTransaction(entry.Entry.Transaction)
+	if err != nil {
+		return nil, err
+	}
+
+	mass := s.txMassCalculator.CalculateTransactionMass(domainTx) // TODO: Does GetMempoolEntry already provide the mass?
```
michaelsutton, line 28: This is missing the storage mass component.

michaelsutton (approved):

PR #2301, opened 2025-01-09
 Adding support for mass.

veronica-bytes:
upgrade grpc to `v1.69.2`


someone235 (approved):

canximan:
I think you are a bit confused between Mass and MassCommitment.

In `/app/appmessage/domainconverters.go` line 223, you set `MassCommitment: rpcTransaction.Mass,`
And in `domain/consensus/utils/consensushashing/transaction.go`, line 130:

```
if includeMass {
    if tx.MassCommitment > 0 { // For backward compatibility, serialize MassCommitment only if it's not zero
	    err = binaryserializer.PutUint64(w, tx.Mass)
	    if err != nil {
		    return err
	    }
    }
}
```

You checked if MassCommitment > and then put Mass which is always ZERO. It is not consistent

someone235:
In kaspad's context it should always be the same, but you're correct that this is not consistent. Fixed by https://github.com/kaspanet/kaspad/pull/2302

Thanks @canximan

canximan:
Why did you add a MassCommitment field instead of using Mass to match rusty-kaspa?

I'm currently using kaspad, but this small change is confusing to me, and adding MassCommitment next to Mass would make the structure more expensive to store, which is confusing.

PR #2304, opened 2025-03-18
style: remove redundant else

x100111010:


someone235:
This repo is deprecated and https://github.com/kaspanet/rusty-kaspa is maintained instead

